PR #2, opened 2015-03-07
Possible warning

dimitrov-adrian:
There is some cases that this could be unset


gregpriday:
You're right, it is always safest to check array keys before accessing them. Thanks for the contribution Adrian.

PR #4, opened 2015-04-08
Fix wp-cli "wp_is_mobile()" fatal error when installing themes

vodou:
When using [wp-cli](http://wp-cli.org)'s `wp theme install` command with the so-widgets-bundle plugin activated, I would get a Fatal Error on the line that called `wp_is_mobile()`. This error is corrected if I check for the existence of `wp_is_mobile()` first.

I suspect that wp-cli isn't loading all of WP's functionality to do its job.


braamgenis:
Hi Patrick

Apologies for taking so long to respond and thanks very much for the fix. It has been implemented manually [here](https://github.com/siteorigin/so-widgets-bundle/commit/5548d99c056154dff3ed495e8a70cc49a6c0cb3c).

PR #6, opened 2015-04-09
apply filter to capability of admin page

drumba:
I want to filter the capability of the admin page. At the moment it is not possible to access the page if DISALLOW_FILE_MODS is set to true. Would be nice to see it in the next release.


gregpriday:
I agree - very useful filter. Thanks for the contribution! :+1:

PR #28, opened 2015-10-15
Fb874 wb features text size

braamgenis:
There are a couple of things which can be refactored. For example, the Google font import code is now duplicated in Features widget and Headline widget and perhaps how the measurement value is obtained when using the new measurement field.

### DIFF
--- a/base/inc.php
+++ b/base/inc.php
@@ -309,3 +309,25 @@ function siteorigin_widgets_tinymce_admin_print_styles() {
 	wp_enqueue_style( 'editor-buttons' );
 }
 add_action( 'admin_print_styles', 'siteorigin_widgets_tinymce_admin_print_styles' );
+
+/**
+ * Get list of supported measurements
+ *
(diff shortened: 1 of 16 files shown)


gregpriday:
Some feedback:
* To keep things neat, we should have all the text settings for the Features Widget in a single section. Collapsed by default.
* Especially for things like fonts, we should make default colors decided by the theme. Right now, they're all being set to `#000000`. Using an empty default should be enough.
* I know they're be overwritten, but to keep things neat, the default values of the font variables in features.less should be something that implies they'd come from the theme. I think just changing these to `@title_font: default;` might work.

Yeah, might not be a bad idea to move the Google Font import code into the base widget class. Can see this being reused quite a lot in the future. 

Not specific to this PR, but we should actually try apply that default color point to all our widgets. The only other widget I can think of is the headline widget.

gregpriday:
Epic

PR #31, opened 2015-10-23
Fixed incorrect LESS variable.

marcelweder:
The submit button in contact widget has an incorrect LESS variable.


gregpriday:
Very well spotted! Thanks @marcelweder.

PR #32, opened 2015-11-22
CHANGE - Allows for Adaptation based on User Info

lucanos:
- Does not insert Image Wrapper, etc. when no Image is set
- Does not set Anchor tag for Location, when no URL is set
- Sets the Anchor tag target based on New Window selection for that
testimonial
- Does not set Location element, when no Location is set

### DIFF
--- a/widgets/testimonial/tpl/default.php
+++ b/widgets/testimonial/tpl/default.php
@@ -17,12 +17,22 @@
 				<?php endif; ?>
 
 				<div class="sow-testimonial-user">
+					<?php if( $sow_user_image = $this->testimonial_user_image($testimonial, $design) ){ ?>
 					<div class="sow-image-wrapper">
-						<?php echo $this->testimonial_user_image($testimonial, $design) ?>
+						<?php echo $sow_user_image ?>
 					</div>
+					<?php } ?>
 					<div class="sow-text">
 						<strong><?php echo esc_html( $testimonial['name'] ) ?></strong>
-						<a><?php echo esc_html( $testimonial['location'] ) ?></a>
+						<?php if( !empty( $testimonial['location'] ) ){ ?>
+						<?php   if( empty( $testimonial['url'] ) ){ ?>
+						<?php echo esc_html( $testimonial['location'] ) ?>
+						<?php   }elseif( $testimonial['new_window'] ){ ?>
+						<a href="<?php echo $testimonial['url']; ?>" target="_blank"><?php echo esc_html( $testimonial['location'] ) ?></a>
+						<?php   }else{ ?>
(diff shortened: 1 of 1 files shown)


widgets/testimonial/tpl/default.php
```diff
@@ -17,12 +17,22 @@
 				<?php endif; ?>
 
 				<div class="sow-testimonial-user">
+					<?php if( $sow_user_image = $this->testimonial_user_image($testimonial, $design) ){ ?>
```
braamgenis: This would be a bit more readable if you assigned the variable before the `if` statement. E.g.
```
$sow_user_image = $this->testimonial_user_image($testimonial, $design);
if( ! empty( $sow_user_image ) )
//etc
```

widgets/testimonial/tpl/default.php
```diff
@@ -17,12 +17,22 @@
...
 					</div>
+					<?php } ?>
 					<div class="sow-text">
 						<strong><?php echo esc_html( $testimonial['name'] ) ?></strong>
-						<a><?php echo esc_html( $testimonial['location'] ) ?></a>
+						<?php if( !empty( $testimonial['location'] ) ){ ?>
+						<?php   if( empty( $testimonial['url'] ) ){ ?>
+						<?php echo esc_html( $testimonial['location'] ) ?>
+						<?php   }elseif( $testimonial['new_window'] ){ ?>
+						<a href="<?php echo $testimonial['url']; ?>" target="_blank"><?php echo esc_html( $testimonial['location'] ) ?></a>
```
braamgenis: The `url` field should be sanitized before being output to the front end, as described here: https://codex.wordpress.org/Data_Validation#URLs

We have a wrapper function for it called `sow_esc_url` which you can use here and in the line below.

braamgenis:
Hi @lucanos 

Thanks for the pull request, it's much appreciated! There are just two small changes you can make to improve it slightly. I've added inline comments, but will mention them here as well.

1. Separate the variable assignment out of the `if` statement.
2. Sanitize the `url` fields before they're `echo`ed.

Thanks again for your contribution. :+1:

braamgenis:
Hi @lucanos 

Unfortunately, we'd need you to make the suggested changes in order for us to merge., but we'll see these as raised issues and feature requests and make sure they're implemented. Thanks for the contributions.

lucanos:
I appreciate the feedback. I will try and take the suggested changes onboard and resubmit a pull request. (Just haven't been able to revisit this project with other time demands over the last week.)

braamgenis:
Hey, thanks for coming back. We really appreciate the input and are looking into these in issues #33 and #36. Please don't hesitate to submit a PR again in future.

PR #43, opened 2015-11-28
Add support for srcset to Image widget

bobbymarko:
Wordpress 4.4 adds support for responsive images via srcset to images added through the visual editor. However, these changes don't get applied to the Site Origin Image Widget. This change uses the new Wordpress wp_get_attachment_image_srcset function to add the required attributes to the widget.

### DIFF
--- a/widgets/image/tpl/base.php
+++ b/widgets/image/tpl/base.php
@@ -13,6 +13,9 @@ if( !empty($src) ) {
 
 	if(!empty($src[1])) $attr['width'] = $src[1];
 	if(!empty($src[2])) $attr['height'] = $src[2];
+	if (function_exists('wp_get_attachment_image_srcset')) {
+		$attr['srcset'] = esc_attr(wp_get_attachment_image_srcset($instance['image'], $instance['size']));
+ 	}
 }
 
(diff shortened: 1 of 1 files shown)


widgets/image/tpl/base.php
```diff
@@ -13,6 +13,9 @@
 
 	if(!empty($src[1])) $attr['width'] = $src[1];
 	if(!empty($src[2])) $attr['height'] = $src[2];
+	if (function_exists('wp_get_attachment_image_srcset')) {
+		$attr['srcset'] = esc_attr(wp_get_attachment_image_srcset($instance['image'], $instance['size']));
```
braamgenis: `esc_attr` isn't necessary here as it's called on all the values of the `$attr` array just before being output below.

braamgenis:
Hi @bobbymarko 

Thanks for the PR! It all looks good but I just have one comment which I've made inline about the unnecessary `esc_attr` call.

bobbymarko:
Good catch @braamgenis! It's fixed now. Thanks!

braamgenis:
Awesome, good to go! :+1:

Issue #34, opened 2015-11-28
Contact Form: Prevent Duplicate Contact Emails


gregpriday:
It seems like it's possible for a form submission to send multiple emails. We should have some sort of duplicate checking for this.

Ideally, to keep in the spirit of widgets, we should try do this without storing too much data. One possibility is to use transients to store the nonce keys and assure that they aren't used more than once.

This was reported here: https://siteorigin.com/thread/contact-form-sends-multi-emails/

Issue #35, opened 2015-11-28
Contact form: Add more recaptcha customization options


gregpriday:
There are a few customization options (e.g. dark or light theme) we could allow as a nice premium feature.

https://developers.google.com/recaptcha/docs/display

PR #48, opened 2015-12-02
Contact form fixes

braamgenis:
Resolves issues #34 and #35 as well as an issue I picked up with having multiple recaptchas on a page.

### DIFF
--- a/widgets/contact/js/contact.js
+++ b/widgets/contact/js/contact.js
@@ -1,3 +1,40 @@
+var SiteOriginContactForm = {
+  init: function ($, useRecaptcha) {
+    var $contactForms = $('form.sow-contact-form');
+    $contactForms.each(function () {
+      var $el = $(this);
+      if(useRecaptcha) {
+        // Render recaptcha
+        var $recaptchaDiv = $el.find('.sow-recaptcha');
+        if ($recaptchaDiv.length) {
+          var sitekey = $recaptchaDiv.data('sitekey');
+          grecaptcha.render($recaptchaDiv.get(0),
+            {
+              'sitekey': sitekey,
+              'callback': function(response) {
+                // Enable the submit button once we have a response from recaptcha.
+                $(this).find('.sow-submit-wrapper > input.sow-submit').prop('disabled', false);
+              }.bind(this),
+            }
+          );
+        }
+      }
+
+      // Disable the submit button on click to avoid multiple submits.
+      var $submitButton = $el.find('.sow-submit-wrapper > input.sow-submit');
+      $submitButton.click(function () {
+        $submitButton.prop('disabled', true);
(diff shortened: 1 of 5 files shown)


widgets/contact/js/contact.js
```diff
@@ -0,0 +1,61 @@
+var SiteOriginContactForm = {
```
gregpriday: @braamgenis For the most part, we've been doing indenting using 4 spaces in our Javascript files. This file seems to be using 2 space indenting. Although we really should be using the WordPress [JS best practices](https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/) and using tab indents.

At some point we should do a more thorough run through all our code to make sure we're following these guidelines.

widgets/contact/tpl/default.php
```diff
@@ -35,25 +42,13 @@
...
-			</script>
-			<?php endif; ?>
+			<div class="sow-recaptcha"
+				 data-config="<?php echo esc_attr( json_encode( $settings ) ) ?>"></div>
 		<?php endif; ?>
 
 		<div class="sow-submit-wrapper <?php if( $instance['design']['submit']['styled'] ) echo 'sow-submit-styled' ?>">
 			<input type="submit" value="<?php echo esc_attr( $instance['settings']['submit_text'] ) ?>"
-				   class="sow-submit <?php echo 'js-sow-submit-' . $instance['_sow_form_id'] ?>" <?php if( $use_recaptcha && !is_admin() ) echo 'disabled="true"'; ?>>
+				   class="sow-submit" <?php if( $use_recaptcha && !is_admin() ) echo 'disabled="true"'; ?>>
```
gregpriday: Are we making considerations for browsers with Javascript disabled? Wont this button always be disabled for these users?

Come to think of it, how does Recaptcha handle non JS browsers? Does it handle them at all? Or has the world decided disabling JS just isn't cool?

widgets/contact/js/contact.js
braamgenis: Ah, yeah, I'm going to add an .editorconfig file. Should help prevent this in future.

widgets/contact/tpl/default.php
braamgenis: Nope. I don't think many sites will work with JS disabled these days. Besides the fact that the Recaptcha API is loaded as a JS file. :grin:

gregpriday: Something that might come up is a user wanting to disable Recaptcha so they can offer their non JS users access. Ideally we want to offer users a non JS mode of operation, even if it becomes slightly limited.

Worth reading the arguments here to decide how far we want to go with JS support - http://www.punkchip.com/why-support-javascript-disabled/

Issue #44, opened 2015-11-29
Allow LatLong input for Google Maps widget center and markers.


braamgenis:
User requested we support LatLng inputs [here](https://siteorigin.com/thread/google-maps-widget-custom-latlong-markers/)

We need to detect or let the user specify when they want to use LatLng as input for the map center or map markers and then use [reverse geocoding](https://developers.google.com/maps/documentation/geocoding/intro#ReverseGeocoding)

PR #57, opened 2015-12-09
Allow LatLong input for Google Maps widget center and markers.

braamgenis:
To resolve issue #44 
Some refactoring to send map configuration options as single data attribute and separate markers and directions functions.
Fix for bug when using TinyMCE field in directions waypoints.


gregpriday:
Everything looks fine and dandy to me.

Issue #47, opened 2015-12-01
SO Editor seems to remove newlines in code formatting.


braamgenis:
When editing code in the visual editor it seems to remove newlines. For example
```
<script language="JavaScript">
<!--
function JumpToIt(frm) {
    var newPage = frm.url.options[frm.url.selectedIndex].value
    if (newPage != "None") {
        location.href=newPage
    }
}
//-->
</script>
```
is output as
```
<script language="JavaScript"><!--function JumpToIt(frm) {    var newPage = frm.url.options[frm.url.selectedIndex].value    if (newPage != "None") {        location.href=newPage    }}//--></script>
```

gregpriday:
My guess would be that some standard filters somewhere are doing this. Does it make much of a difference though? Removing newlines shouldn't ever cause syntax errors.

braamgenis:
So this is code for display purposes. So if a user wants to display some code in a post or something. The standard WP visual editor doesn't seem to remove newlines. And I'm guessing it will be a fairly straight forward fix.

gregpriday:
Aaah OK, got you. Definitely worth fixing then.

azghanvi:
+1 for this issue.

azghanvi:
vim ./wp-content/plugins/so-widgets-bundle/widgets/so-editor-widget/so-editor-widget.php

Comment the line, $string = str_replace("\n", "", $string); in function 'unwpautop'

        function unwpautop($string) {
                //$string = str_replace("\n", "", $string);
                $string = str_replace("<p>", "", $string);
                $string = str_replace(array("<br />", "<br>", "<br/>"), "\n", $string);
                $string = str_replace("</p>", "\n\n", $string);

                return $string;
        }

This seems to resolve the issue, but the actual authors will comment if it has some side effects or not.

PR #61, opened 2015-12-13
Fix of #47 - SO Editor seems to remove newlines in code formatting.

azghanvi:
Fix of #47 - SO Editor seems to remove newlines in code formatting.


gregpriday:
Thanks for the PR @azghanvi. Everything looks good. I think @braamgenis wanted to check something before merging, but we should get this in soon.

PR #64, opened 2015-12-18
Alter the default caps

BronsonQuick:
Hey folks!

Firstly, thanks for your great plugin!

I was creating a new widget this afternoon on a site and I hit a bug where activating a widget wouldn't actually save the activation. I noticed in the console that I was getting anything returned when I clicked activate so I started hunting down the issue and it turned out to be the default caps.

For our sites we have `define( 'DISALLOW_FILE_MODS', true );` set in `wp-config.php` so that clients can't update plugins however they can still alter settings. In that scenario they can get to the Plugins->SiteOrigin Widgets can they can click Activate and the plugin looks like it activated but when you hit refresh it's not.

To get around this `manage_options` is a more sensible capability to check for. It's an edge case but I think it's worth handling.

I have hooked into the filter to change the caps in my plugin in case you're not a fan of this PR.

Thanks again!

### DIFF
--- a/so-widgets-bundle.php
+++ b/so-widgets-bundle.php
@@ -294,7 +294,7 @@ class SiteOrigin_Widgets_Bundle {
 	 */
 	function admin_ajax_manage_handler(){
 		if( !wp_verify_nonce($_GET['_wpnonce'], 'manage_so_widget') ) exit();
-		if( !current_user_can( apply_filters('siteorigin_widgets_admin_menu_capability', 'install_plugins') ) ) exit();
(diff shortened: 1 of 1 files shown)


gregpriday:
Very well spotted @BronsonQuick, this is exactly what should be happening. We recently changed that management page to work with `manage_options`, but overlooked setting it for the ajax action.

https://github.com/siteorigin/so-widgets-bundle/commit/8c52764f4ac3e421bcd787341074d3847322913a

Very much appreciated :+1:

BronsonQuick:
Haha we're always impressed by how fast you are with PR's @gregpriday. Keep it up!

noeltock:
Thanks! :+1:

Issue #68, opened 2015-12-22
Contact Form: Sanitize from header


gregpriday:
This header should be sanitized to remove newlines and properly sanitize the email address.

https://github.com/siteorigin/so-widgets-bundle/blob/develop/widgets/contact/contact.php#L942

Issue #66, opened 2015-12-22
Contact Form: Escape variables in the email body


gregpriday:
The HTML content of the contact form looks like it was originally meant to be plain text. So certain things were overlooked like escaping user inputted content.

https://github.com/siteorigin/so-widgets-bundle/blob/develop/widgets/contact/contact.php#L931-L938

This should be escaped.

Issue #65, opened 2015-12-22
Contact Form: From in HTML body generates invalid code


gregpriday:
This from attribute in the email body looks like it was originally intended for standard plain text emails.

https://github.com/siteorigin/so-widgets-bundle/blob/develop/widgets/contact/contact.php#L931

This should have the brackets HTML encoded and the email should ideally be wrapped in `<a href="mailto:...">...</a>`.

PR #70, opened 2015-12-23
Fix sanitization in the contact form

gregpriday:
This resolves GH-68, GH-66 and GH-65.

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -946,7 +946,9 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 	}
 
 	/**
-	 * Sanitize a value for an email header
+	 * Sanitize a value for an email header.
+	 *
+	 * From Pear Mail https://pear.php.net/package/Mail (BSD Style license - https://pear.php.net/copyright.php).
 	 *
 	 * @param $value
 	 *
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -928,21 +928,32 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 	}
 
 	function send_mail( $email_fields, $instance ){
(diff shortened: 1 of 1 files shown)


widgets/contact/contact.php
```diff
@@ -928,22 +928,33 @@ function spam_check( $post_vars, $email_fields, $instance ){
 	}
 
 	function send_mail( $email_fields, $instance ){
-		$body = '<strong>From:</strong> ' . $email_fields['name'] . ' <' . sanitize_email( $email_fields['email'] ) . ">\n\n";
+		$body = '<strong>From:</strong> <a href="mailto:' . sanitize_email( $email_fields['email'] ) . '">' . esc_html( $email_fields['name'] ) . "</a>>\n\n";
```
braamgenis: Looks like an extra '>' after the closing anchor tag.

widgets/contact/contact.php
```diff
@@ -928,22 +928,33 @@ function spam_check( $post_vars, $email_fields, $instance ){
...
 
+	/**
+	 * Sanitize a value for an email header
+	 *
+	 * @param $value
+	 *
+	 * @return mixed
+	 */
+	static function sanitize_header( $value ){
+		return preg_replace('=((<CR>|<LF>|0x0A/%0A|0x0D/%0D|\\n|\\r)\S).*=i', null, $value);
```
braamgenis: Is this replacing any newline type character with a `null`? Why not just an empty string?

braamgenis:
Couple of minor things, but otherwise good to merge when you're ready.

widgets/contact/contact.php
gregpriday: Aaah, nicely spotted. Will fix that up.

widgets/contact/contact.php
gregpriday: I think that's directly out of another (GPL compatible) library. I should probably add some form of attribution header to that function though.

PR #71, opened 2015-12-26
Update attachments.php

olidac:
add > to close img tag


gregpriday:
Very well spotted @olidac. Thanks for the contribution!

PR #86, opened 2016-01-21
Added support for lat-lng markers.

miili:
Marker latlng check now w regex.
setTimeout only when `google.maps.Geocoder` was used.


gregpriday:
Thanks for the contribution! @braamgenis will take a look at this - the maps widget isn't my area.

gregpriday:
Sorry for the delay @miili - it looks like @braamgenis already implemented this in https://github.com/siteorigin/so-widgets-bundle/commit/738b73c6095286dfdb1b34694ee816129cbeb479.

Issue #82, opened 2016-01-21
Widget ID migration not working on update from 1.5.4


miili:
Look like `SiteOrigin_Widgets_ImageGrid_Widget` is missing in 1.5.3, or was it removed on purpose?

gregpriday:
It shouldn't have been removed. I'll look into this quickly. It's possible it's as a result of a fairly substantial folder renaming change we did.

gregpriday:
The widget still seems to be there for me. Can you check if you have it active in `Plugins > SiteOrigin Widgets`?

miili:
Hi Greg,

indeed the update deactivated the widget. Yet after re-activation the old content is broken.
Any ideas how to solve it quickly?

thanks for your support,

Marius

gregpriday:
For now I'd just suggest rolling back to version 1.5.4 while I try recreate the issue.

You can download it from the developer page - https://wordpress.org/plugins/so-widgets-bundle/developers/

miili:
:+1:

gregpriday:
@braamgenis this seems to be an issue with the migration of short names to the long names convention.  I was able to recreate the issue. Perhaps we need to just do a filter of the settings every time `get_active_widgets` is called instead of just doing it once on update.

You can test this either with fresh builds or downloading a 1.5.4 build off the Widgets Bundle developer page.

gregpriday:
Just moving the migration code to `get_active_widgets` actually fixes the problem. Not sure why it's not working with the version update function though.

I'll submit a PR for this soon.

PR #87, opened 2016-01-22
Moved widget ID migration

gregpriday:
This resolves GH-82. Basically just moved the migration function into the get_active_widgets function and added some caching to ensure there isn't much of a performance hit.


braamgenis:
Cool, looks good. Merging.

PR #91, opened 2016-01-27
OR relation for multiple taxonomy terms

live-mesh:
When multiple taxonomy terms are specified for a query, wordpress defaults to relation AND but my experience says the users mostly mean OR when they include multiple taxonomy terms. The use case of someone wanting to see all posts belonging to either category/term X or Y is much more common than someone wanting to create a query for a post which belongs to both category/terms X and Y, IMO. Pls review and merge if this seems right.

### DIFF
--- a/base/inc/post-selector.php
+++ b/base/inc/post-selector.php
@@ -53,7 +53,7 @@ function siteorigin_widget_post_selector_process_query($query){
 		$tax_queries = explode(',', $query['tax_query']);
 
 		$query['tax_query'] = array();
-        $query['tax_query']['relation'] = 'OR';
+		$query['tax_query']['relation'] = 'OR';
 		foreach($tax_queries as $tq) {
 			list($tax, $term) = explode(':', $tq);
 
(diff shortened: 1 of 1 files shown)


gregpriday:
You're definitely right here. Or makes a lot more sense. The only minor thing before I merge this would be to make sure your addition follows the indenting style of Widgets Bundle. So changing the space indenting to tab indenting.

Thanks for the contribution, and for creating your awesome addon to Widgets Bundle. I'd love to cover it in a video tutorial one day.

live-mesh:
Have made the indentation change. Have modified our portfolio addon to use build posts query now. This addresses the most requested feature for our portfolio pages in our themes - ability to choose specific portfolio categories when displaying portfolio items.

Glad you like the addon. Hope to put it on GitHub soon so that everyone benefits. Again thanks for building such great plugins!

gregpriday:
Awesome! Thanks a million. And be sure to let me know when you add your plugin to Github.

PR #92, opened 2016-01-28
Fix issue where margin units were doubling causing broken margin rule in Headline Widget.

bgallagh3r:
Adding the margin unit here causes the units to double up and thereby
breaking the CSS rule for the margin around the divider for the Headline
widget.


braamgenis:
Great, thanks for the fix @bgallagh3r!

PR #93, opened 2016-01-29
Default to no background color for slider frame instead of arbitrary gray

jambox:
I was trying to use PNGs with transparent rounded corners and kept seeing a gray color behind them...this removes that color as default and instead doesn't pass a `background-color` attr to the frame element.


deviodigital:
:+1:

gregpriday:
Agreed! Thanks for the contribution @jambox.

jambox:
Glad it helped!

PR #98, opened 2016-02-11
Clear widgets cache

braamgenis:


so-widgets-bundle.php
```diff
@@ -98,37 +102,56 @@ function plugin_version_check(){
...
-						$wp_filesystem->delete( $upload_dir['basedir'] . '/siteorigin-widgets/' . $file['name'] );
-					}
-				}
-			}
+		if( $is_new ) {
 
-			// An action to let widgets handle the updates.
+			update_option( 'siteorigin_widget_bundle_version', SOW_BUNDLE_VERSION );
+			// If this is a new version, then trigger an action to let widgets handle the updates.
 			do_action( 'siteorigin_widgets_version_update', SOW_BUNDLE_VERSION, $active_version );
```
gregpriday: A `clear_widget_cache` here wouldn't hurt. Just as an extra super backup for the other actions not firing for some reason. Or we could just hook another clear widget cache to the `siteorigin_widgets_version_update` action.

gregpriday:
A PR just isn't a PR without a small suggestion, besides that, this all looks ready to merge to me.

so-widgets-bundle.php
braamgenis: I'll add it if you're sure you want the extra super backup, but just be aware that it'll be called twice whenever the Widgets Bundle plugin is updated. I checked that all the actions work as advertised.

gregpriday: Maybe a good compromise would be to have something like this when we detect a version number change.

`if( !has_action( 'upgrader_process_complete' ) && !has_action( 'activated_plugin' ) ) { /* clear all the caches */ }`

Would rather err on the side of caution for something like this incase there's an edge case where those 2 actions aren't called on a plugin update. Some examples of possible edge cases

* A user just deleting the `so-widgets-bundle` folder, then reuploading it from a zip file through FTP. They just leave the plugin active so a `activated_plugin` isn't triggered and they don't go through the updater so `upgrader_process_complete` isn't triggered.
* Possible that something like WP-CLI doesn't trigger these actions when doing a command line plugin update.
* There might be some 3rd party ManageWP-style WordPress management plugins/services that don't call these actions when a plugin is updated. I assume ManageWP is, but there might be other similar services that aren't.

We need to make sure the core widgets are handled properly and offer as much cache clearing as possible to 3rd party addons.

PR #106, opened 2016-02-17
Adding fault tolerance to swipe method invocation

andypotanin:
JS throws a `Uncaught TypeError: $$.swipe is not a function` error in some cases.

Wrapping method into a `typeof` check to avoid issues.


gregpriday:
Looks good to me, also like the addition of the `SOW_BUNDLE_JS_SUFFIX` check. Thanks for the contribution!

Issue #62, opened 2015-12-17
Deal with WP_Filesystem returning false


gregpriday:
Right now, the save_css function doesn't have any fallback when there's a problem with the `WP_Filesystem()` call (this function can return `false`).

https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/siteorigin-widget.class.php#L490-L516

Ideally, instead of returning false, this should cache the CSS in the database and display it as an inline stylesheet in the header.

PR #128, opened 2016-03-02
WP_Filesystem fallback.

braamgenis:
Fall back to using `wp_cache` functions when `WP_Filesystem` not available.
To resolve GH-62.

### DIFF
--- a/base/siteorigin-widget.class.php
+++ b/base/siteorigin-widget.class.php
@@ -203,7 +203,12 @@ abstract class SiteOrigin_Widget extends WP_Widget {
 					}
 					else {
 						// Fall back to using inline CSS if we can't find the cached CSS file.
-						siteorigin_widget_add_inline_css( $this->get_instance_css( $instance ) );
+						// Try get the cached value.
+						$css = wp_cache_get( $hash, 'siteorigin_widgets' );
+						if ( empty( $css ) ) {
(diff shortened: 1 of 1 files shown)


base/siteorigin-widget.class.php
```diff
@@ -203,7 +203,12 @@ function generate_and_enqueue_instance_styles( $instance ) {
 					}
 					else {
 						// Fall back to using inline CSS if we can't find the cached CSS file.
-						siteorigin_widget_add_inline_css( $this->get_instance_css( $instance ) );
+						// Try get the cached value.
+						$css = wp_cache_get( $hash, 'siteorigin_widgets' );
```
gregpriday: Although the hash here is pretty much always going to be unique, it might be a good idea to include `$this->id_base` of the widget as the cache key. It just feels a little neater, because technically all widgets share the same hash space, so could clash.

We're using the hash this way in a few places:

https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/siteorigin-widget.class.php#L901
https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/siteorigin-widget.class.php#L182

braamgenis: Ok, cool. :ok_hand:

braamgenis:
@gregpriday Better?

gregpriday:
Looks good!

PR #136, opened 2016-03-10
Update action order of widget field buttons

KevinTaron:
Everywhere the action order is: "Edit, Copy, Delete" except here.
Here is "Delete, Copy, Edit".

Don't know if this has a special reason and I'm the only one who always clicks on the wrong button but I think this should be the same everywhere.
Fixed that.

### DIFF
--- a/base/css/admin.less
+++ b/base/css/admin.less
@@ -159,7 +159,7 @@
 		}
 
 		.siteorigin-widget-field-expand {
-			right: 12px;
+			right: 48px;
 
 			&:before {
 				content: "\f140";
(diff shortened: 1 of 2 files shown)


KevinTaron:
order-fix

KevinTaron:
Hi Guys,

don't know if that is a bug or a feature but in the Headline Widget some of my customers found a problem for them. 
If the headline is h3 and they don't use the subheadline the subheadline style is still applied.

For example: Headline is H3 with text align left. They don't use Subheadline therefore the style is not changed. 
On the website the Headline is centered because default style of subheadline is applied mit text-align center.

I think if the subheadline is empty the style should not be added.

KevinTaron:
@braamgenis here is the pull request from the issue with the headline / subheadline style

gregpriday:
Thanks for the contribution @KevinTaron! This will be in the next update - due out fairly soon.

Issue #137, opened 2016-03-12
Features: Incorrectly sized image icons


AlexGStapleton:
[issue was reported in this thread](https://siteorigin.com/thread/widgets-bundle-feature-widget-image-issues/), and a few others.

This issue is caused due how we're not specifying the size of the image for the thumbnail. [Issue is here in the code](https://github.com/siteorigin/so-widgets-bundle/blame/develop/widgets/features/tpl/base.php#L21).

wp_get_attachment_image_src() has a size parameter that when not set, it'll default to thumbnail - which by default is 150x150. Thus, this is causing distortion issues and potential undersized icons.

Issue #138, opened 2016-03-12
Features: reference to previous background on image icons after initial icon


AlexGStapleton:
[issue was reported in this thread](https://siteorigin.com/thread/widgets-bundle-feature-widget-image-issues/).

The output of the feature icon image will include the previous icons styling. Example:

`<div class="sow-icon-image" style="background-image: url(http://domain.com/wptemp/wp-content/uploads/2016/03/example-example-icon-250px-150x150.png); font-size: 125px; background-image: url(http://kelken.com/wptemp/wp-content/uploads/2016/03/example-example-101p-icon-250px-150x150.png); font-size: 125px"></div>`

This is due to not [resetting the $icon_styles variable](https://github.com/siteorigin/so-widgets-bundle/blame/develop/widgets/features/tpl/base.php#L23) before outputting the next feature icon. The icon (the non-image version) [does clear it](https://github.com/siteorigin/so-widgets-bundle/blame/develop/widgets/features/tpl/base.php#L30) and thus doesn't face this problem.

PR #143, opened 2016-03-15
Allow user to select size for uploaded icon image.

braamgenis:
Added a select field so users can choose from registered image sizes. 

To resolve GH-137, GH-138.

### DIFF
--- a/widgets/features/features.php
+++ b/widgets/features/features.php
@@ -8,11 +8,6 @@ Author URI: https://siteorigin.com
 
 class SiteOrigin_Widget_Features_Widget extends SiteOrigin_Widget {
 	function __construct() {
-		$image_size_configs = $this->get_image_sizes();
-		$sizes = array();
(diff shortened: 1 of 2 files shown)


widgets/features/features.php
```diff
@@ -8,6 +8,11 @@
 
 class SiteOrigin_Widget_Features_Widget extends SiteOrigin_Widget {
 	function __construct() {
+		$image_size_configs = $this->get_image_sizes();
```
gregpriday: Generally for this heavy code stuff, we should be implementing form options in the `modify_form` function. Huge advantage there is that this function is only called when the form is generated, while the constructor is called on every page load.

One day I'd actually like to move all form options out of the constructors. To make things a little more efficient.

braamgenis: Ah, good point. Cool, will move it. :+1:

Issue #153, opened 2016-03-25
Button: Allow the user to manually specify the buttons width


AlexGStapleton:
Sometimes users don't really care about having fluid widths for buttons and right now it's impossible to manually set a width without CSS.

Use Case: I have a three column row with three button widgets in each column. Each has different amounts of text but I want them all to be 150px in width so it looks consistent. Without a width setting they will look mismatched (width wise) with one being tiny and another being extremely long.

PR #159, opened 2016-03-30
Button widget width

braamgenis:
To resolve GH-153. Can manually set width or leave width field blank to resize to content.


gregpriday:
@braamgenis: you feel like merging develop. Conflicts probably due to moving form arrays.

braamgenis:
@gregpriday Cool fixed conflicts and an issue I found. The width setting was acting as a max-width, instead of width. So I removed the `div` wrapper and just applied the width directly to the link "button" with `box-sizing: border-box;` set.

Issue #154, opened 2016-03-27
Widget Request: Icon


AlexGStapleton:
There should be a way to add icons more easily. A dedicated icon widget wouldn't be a bad idea. It could just be a standard (but expanded) [Icon field](https://siteorigin.com/docs/widgets-bundle/form-building/form-fields/#heading-icon) with basic settings such as color, size and potentially even a link option.

PR #161, opened 2016-03-31
Icon widget.

braamgenis:
To resolve GH-154.

### DIFF
--- a/widgets/icon/icon.php
+++ b/widgets/icon/icon.php
@@ -1,0 +1,99 @@
+<?php
+
+/*
+Widget Name: Icon
+Description: An iconic icon.
+Author: SiteOrigin
+Author URI: https://siteorigin.com
+*/
+
+class SiteOrigin_Widget_Icon_Widget extends SiteOrigin_Widget {
+
+	function __construct() {
+
+		parent::__construct(
+			'sow-icon',
+			__( 'SiteOrigin Icon', 'so-widgets-bundle' ),
+			array(
+				'description' => __( 'An icon widget.', 'so-widgets-bundle' )
+			),
+			array(),
+			false,
+			plugin_dir_path( __FILE__ )
+		);
+	}
+
+	function initialize_form() {
+		return array(
+			'icon' => array(
+				'type'  => 'icon',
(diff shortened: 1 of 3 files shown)


widgets/icon/icon.php
```diff
@@ -0,0 +1,99 @@
...
+			),
+		);
+	}
+
+	function get_style_name( $instance ) {
+		return 'sow-icon';
+	}
+
+	function get_less_variables( $instance ) {
+		$les = ';asda';
```
gregpriday: What's this do?

widgets/icon/tpl/icon.php
```diff
@@ -0,0 +1,19 @@
+<?php
+/**
+ * @var $icon
```
gregpriday: Aaah nice, I need to start doing this.

widgets/icon/styles/sow-icon.less
```diff
@@ -0,0 +1,14 @@
+@color: h1;
```
gregpriday: Ideally, these defaults should make sense. So the actual default color as `@color`. Same for `@size`. Can't remember if this has any benefit, but it just feels neater.

widgets/icon/tpl/icon.php
```diff
@@ -0,0 +1,19 @@
...
+<div id="sow-icon-container" class="sow-icon">
+	<?php if ( ! empty( $url ) ) { ?>
+	<a href="<?php esc_attr_e( $url ) ?>" <?php if ( ! empty( $new_window ) ) echo 'target="_blank"'; ?>>
+	<?php } ?>
+		<?php
+		echo siteorigin_widget_get_icon( $icon );
+		?>
+	<?php if ( ! empty( $url ) ) { ?>
+	</a>
+<?php } ?>
```
gregpriday: One level of indenting on this line would be a little neater.

widgets/icon/icon.php
braamgenis: Let's me put breakpoints in the `get_less_variables` function. I guess I should clean up after myself. :sweat_smile:

Issue #162, opened 2016-03-31
Icon Widget: Icon selector closes when you switch icon library and then switch back to the one you were initially using


AlexGStapleton:
When you open the icon selector, switch icon library and then switch back to the initial font library it'll close the selector. This happens with all icon libraries and can be tested by saving with a different library and then repeating the test.

Console empty so it's not a JS issue.

PR #164, opened 2016-04-01
Fix icon selector closing unexpectedly

braamgenis:
To resolve GH-162.
It was simulating a click on the selected icon when choosing the icon family from the dropdown, which caused the icon grid to close. As far as I could see, the only purpose of this was to set the selected icon on the icon selector button.


gregpriday:
Yeah, that seems about right.

Issue #69, opened 2015-12-22
Contact Form: Add “descriptions” to form fields


bogdanb:
It would be nice to have the option of adding a “description” below fields in contact form. 

These are useful for, e.g., reminding people what sort of information is useful in a support request, or linking to a privacy policy when asking for an email address. (So HTML is better than text-only.)

Might be nice to have a “description” setting for the entire form, i.e. to add a message below the submit button. (You can do that separately in the WordPress editor, but if it’s part of the form it could help with styling.)

gregpriday:
Another great suggestion @bogdanb - please keep them coming.

PR #166, opened 2016-04-04
Contact Form: field descriptions

braamgenis:
To resolve GH-69. Field for adding descriptions containing HTML below contact form fields.

### DIFF
--- a/widgets/contact/styles/default.less
+++ b/widgets/contact/styles/default.less
@@ -85,11 +85,13 @@
 		}
 	}
 
+	@description_font_color: #999999;
+
 	.sow-form-field-description {
 		font-size: 0.9em;
 		margin-top: 0.2em;
 		margin-left: 0.2em;
-		color: #999;
+		color: @description_font_color;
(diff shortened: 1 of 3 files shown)


widgets/contact/styles/default.less
```diff
@@ -84,6 +84,14 @@
 			}
 		}
 	}
+
+	.sow-form-field-description {
+		font-size: 0.9em;
+		margin-top: 0.2em;
+		margin-left: 0.2em;
+		color: #999;
```
gregpriday: We should probably also have a field in the design section to change this default colour.

braamgenis: Okily dokily. :+1:

braamgenis:
@gregpriday Added field to set description color. You think we should allow more customization? (Like font family, size, etc)

gregpriday:
Font size would be a good one, but I wouldn't be keen on font family, or much more than that. Maybe an italics option. Italic vs non-italic would be a fairly common change.

PR #181, opened 2016-04-20
Contact Form field refactor

braamgenis:
Refactored contact form fields  to use autoloaded OO classes with render functions. Works with the same autoloader as used by admin form fields.

### DIFF
--- a/base/inc/fields/siteorigin-widget-field-class-loader.class.php
+++ b/base/inc/fields/siteorigin-widget-field-class-loader.class.php
@@ -33,7 +33,11 @@ class SiteOrigin_Widget_Field_Class_Loader {
 	 */
 	public function add_class_prefixes( $class_prefixes ) {
 		if( !isset( $this->class_prefixes ) ) $this->class_prefixes = array();
-		$this->class_prefixes = array_merge( $this->class_prefixes, $class_prefixes );
+		foreach ( $class_prefixes as $prefix ) {
+			if ( ! in_array( $prefix, $this->class_prefixes ) ) {
+				$this->class_prefixes[] = $prefix;
+			}
+		}
 	}
 
 	/**
@@ -43,7 +47,11 @@ class SiteOrigin_Widget_Field_Class_Loader {
 	 */
 	public function add_class_paths( $class_paths ) {
 		if( !isset( $this->class_paths ) ) $this->class_paths = array();
-		$this->class_paths = array_merge( $this->class_paths, $class_paths );
+		foreach ( $class_paths as $path ) {
+			if ( ! in_array( $path, $this->class_paths ) ) {
+				$this->class_paths[] = $path;
+			}
+		}
 	}
 
 	/**
@@ -64,6 +72,8 @@ class SiteOrigin_Widget_Field_Class_Loader {
 
 		foreach( $this->class_paths as $class_path ) {
 			$filepath = $class_path . $filename . '.class.php';
+			// TODO: Fix. Currently, if there is more than one file with the same name in the different class paths,
+			//they will all be loaded. Shouldn't cause any problems, just not as efficient as it could be.
 			if ( file_exists( $filepath ) ) {
 				require_once $filepath;
 			}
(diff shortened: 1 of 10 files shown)


gregpriday:
Is it not messy having a shared class loader when the field classes have very different uses?

base/inc/fields/siteorigin-widget-field-class-loader.class.php
```diff
@@ -64,6 +72,8 @@ public function load_field_class( $field_classname ) {
 
 		foreach( $this->class_paths as $class_path ) {
 			$filepath = $class_path . $filename . '.class.php';
+			// TODO: Fix. Currently, if there is more than one file with the same name in the different class paths,
+			//they will all be loaded. Shouldn't cause any problems, just not as efficient as it could be.
```
gregpriday: Wouldn't a `continue;` do the trick?

```
if ( file_exists( $filepath ) ) {
   require_once $filepath;
   continue;
}
```

widgets/contact/contact.php
```diff
@@ -1060,3 +1033,13 @@ static function sanitize_header( $value ){
 
 }
 siteorigin_widget_register( 'sow-contact-form', __FILE__, 'SiteOrigin_Widgets_ContactForm_Widget' );
+
+// Tell the autoloader where to look for contactform field classes.
+function contactform_fields_class_paths( $class_paths ) {
+	$class_paths[] = plugin_dir_path( __FILE__ ) . 'fields/';
+	return $class_paths;
+}
+add_filter( 'siteorigin_widgets_field_class_paths', 'contactform_fields_class_paths' );
```
gregpriday: Re: my previous comment, this is the main issue with sharing a class loader between the base form fields and the contact form fields.They're sharing this same filter which implies they're the same thing.

base/inc/fields/siteorigin-widget-field-class-loader.class.php
braamgenis: Yeah, that should do it. My brain was focussed on other things, so I just made a note and hit a `continue`.

braamgenis:
So I think I made a mistake here in assuming we wouldn't want to autoload other classes in future, or at least if we did, we'd create more autoloaders. My thinking is that there should really only be one autoloader per project (or plugin) which should be able to handle all autoloading requirements generically. Otherwise we're going to end up with a bunch of autoloaders all over the place, each basically doing the same thing as the other, just with a different name.

gregpriday:
Can definitely go with having a single autoloader, as long as it's handling independant class groups better.

Are you sure we're not opening ourselves up to conflicts by sharing the same filters? So maybe another plugin adding a contact form field with a fairly generic name like `location`, and then us adding that as a field to the core Widgets Bundle forms? Is there no chance that there'd be a conflict?

Something else to think about is that [file_exists](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/inc/fields/siteorigin-widget-field-class-loader.class.php#L67) is a [semi](http://stackoverflow.com/questions/316629/is-file-exist-in-php-a-very-expensive-operation) expensive operation. This means adding a fields path for Contact Form fields, will add overhead for the rest of the Widgets Bundle. It'll need to start checking additional folders.

Probably worse, on the frontend of a user's site, the Contact Form field autoloader will potentially check through each of the Widget Bundle field folders before autoloading its class.

All minor things though, if you're confident a shared class loader wont cause any issues, then we can go ahead with it.

gregpriday:
So here's my idea for a grouped class loader: https://github.com/siteorigin/so-widgets-bundle/tree/feature/contact-form-fields-refactor-alt - easier to code it than explain it.

* It's slightly more efficient. Allows us to safely use [continue](https://github.com/siteorigin/so-widgets-bundle/blob/feature/contact-form-fields-refactor-alt/base/inc/fields/siteorigin-widget-field-class-loader.class.php#L91) (this might have caused problems in the other implementation).
* It gives other plugins a clear way to add Contact Form fields. There are [separate filters](https://github.com/siteorigin/so-widgets-bundle/commit/a08660257c1c4908076c263a9653672d9e2c4146#diff-37339cb063cee02ce4218349c2872499R1042). So plugins can add their class path and contact form field prefixes without worrying about conflicting with the core form fields.
* The Contact Form fields share a prefix that better implies that they come from a different base to the core fields.

So this version uses the same class loader, but it's a little easier to for other plugins to extend and exposes us to fewer potential conflicts across the different field groups. It's also backwards compatible with any other plugins that have already added core fields.

PR #194, opened 2016-04-27
Alternative Contact Form fields refactor.

gregpriday:
Explained this alternative implementation here: https://github.com/siteorigin/so-widgets-bundle/pull/181#issuecomment-215109911

### DIFF
--- a/base/inc/fields/siteorigin-widget-field-class-loader.class.php
+++ b/base/inc/fields/siteorigin-widget-field-class-loader.class.php
@@ -12,8 +12,9 @@ class SiteOrigin_Widget_Field_Class_Loader {
 
 	function __construct(){
 		// Setup the loader with default prefixes and paths
-		$this->add_class_prefixes( array( 'SiteOrigin_Widget_Field_' ) );
-		$this->add_class_paths( array( plugin_dir_path( __FILE__ ) ) );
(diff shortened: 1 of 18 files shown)


base/inc/fields/siteorigin-widget-field-class-loader.class.php
```diff
@@ -29,29 +30,35 @@ static function single(){
 	/**
 	 * Regsiter class prefixes to watch for in this loader
 	 *
-	 * @param $class_prefixes
+	 * @param string|array $class_prefixes
+	 * @param string $group
 	 */
-	public function add_class_prefixes( $class_prefixes ) {
+	public function add_class_prefixes( $class_prefixes, $group = 'core' ) {
```
braamgenis: I suspect having a default value will lead to 3rd party devs using the default too which defeats the whole purpose of having the groups.

braamgenis:
@gregpriday Cool, it looks good. Just have the one comment about default value for group.

base/inc/fields/siteorigin-widget-field-class-loader.class.php
gregpriday: The main reason I'd want this default is for backwards compatibility. I don't know of any plugins that are using this function, but it's possible some devs have. Either way, this is in the spirit of backwards compatibility - if there is such a thing.

PR #214, opened 2016-05-19
Image Subtitles

atomicnation:
By adding few lines of code we can accomplish to have titles + subtitles
texts for your images. The modified files create the "subtitle text"
field and the "subtitle position" field in the SiteOrigin Image widget
backend.

I tested for previous version localy and now I've made the changes in
the updated plugin (Version 1.5.11) and working fine.

Maybe the way of echoing way for titles and subtitles in base.php have to be updated, but does the job.

This is a minor change that could be easily integrated for the new
version.

### DIFF
--- a/widgets/image/image.php
+++ b/widgets/image/image.php
@@ -79,11 +79,29 @@ class SiteOrigin_Widget_Image_Widget extends SiteOrigin_Widget {
 					'below' => __( 'Below', 'so-widgets-bundle' ),
 				),
 			),
+			
+			/****BY SGC****/
+			'subtitle' => array(
+				'type' => 'text',
+				'label' => __('Subtitle text', 'so-widgets-bundle'),
+			),
+			'subtitle_position' => array(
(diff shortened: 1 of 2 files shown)


gregpriday:
Thanks for the contribution @atomicnation - unfortunately, a subtitle isn't something we'd really want to the image widget. The standard for widgets is to have a single title, that's echoed between the `before_title` and `after_title` tags. Even having the `<h2>` tag in the image widget was a bug, which I've corrected.

https://github.com/siteorigin/so-widgets-bundle/commit/a637b000e6090c956e983ead7a4c583c719cac30

atomicnation:
Such a pitty! I see the point in having only titles in any other kind of widgets, but having the possibility to add title plus other text field (called subtitle or whatever) allows to have a little description when using the images as a link banner, like in this example: http://postimg.org/image/thggfj5uj/. About the <h2>, ok this is something that can be avoided.

PR #241, opened 2016-06-20
Image size field

### LINKED ISSUE #238: UX Image: Add image dimensions in image size list
Right now the image sizes in the SiteOrigin Image widget aren't intuitive as the average person won't know how large an image size actually is. Awkwardly, the SiteOrgin Features widget lists the image dimensions.

braamgenis:
To resolve #238


gregpriday:
Looks good to me

PR #258, opened 2016-07-05
Fix XSS vulnerability

chakzefir:
Vulnerable file **/so-widgets-bundle/widgets/contact/tpl/default.php** do not properly escape REQUEST_URI that makes possible **XSS** attack. Fix this filtering the final url via **esc_url()**

### DIFF
--- a/widgets/contact/tpl/default.php
+++ b/widgets/contact/tpl/default.php
@@ -28,7 +28,7 @@ else {
 		);
 	}
 	?>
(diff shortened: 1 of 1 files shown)


gregpriday:
Thanks for the contribution! Yeah, this should be escaped by `esc_url` - I'll merge your change now. I investigated and can't see any major ways for an attacker to exploit this. That said, I'll definitely push an update ASAP.

chakzefir:
@gregpriday thanks for so fast response.

chakzefir:
@gregpriday have you include this fix in 1.6.1 version ?

chakzefir:
@gregpriday could include current fix in nearest release ?

PR #259, opened 2016-07-05
Basic accordion

KevinTaron:
Hi,

here is a basic accordion widget. 
Don't know maybe we should create 2-3 themes or something like that.

BR Kevin

### DIFF
--- a/widgets/accordion/accordion.php
+++ b/widgets/accordion/accordion.php
@@ -1,0 +1,130 @@
+<?php
+/*
+Widget Name: Accordion
+Description: A powerful yet simple accordion widget for your sidebars or Page Builder pages.
+Author: SiteOrigin
+Author URI: https://siteorigin.com
+*/
+
+class SiteOrigin_Widget_Accordion_Widget extends SiteOrigin_Widget {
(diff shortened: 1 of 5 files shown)


braamgenis:
Hi @KevinTaron  

Thanks so much for your contribution. Going forward we'll be using a designer to give widgets a consistent look and feel. This means the development work will most likely depend on the design direction taken. When we do that we'll look at and incorporate parts of your code where we can and definitely attribute it accordingly. Thanks again for taking some time to contribute!

KevinTaron:
no problem. 
Would love to see a "widget designer". 
I've created a few smaller widgets for our customers. The designer could improve my workflow!

gregpriday:
@KevinTaron - I created the start of a Widget Builder plugin. Something I really need to finish off - https://github.com/siteorigin/so-widgets-builder

Is this what you mean? I think the main thing I needed to finish was exporting widgets created in the builder.

KevinTaron:
@gregpriday - Yeah that could be exactly what I mean.
I'll look further into this. 
Export function would be necessary to use the widgets on other sites.

Issue #248, opened 2016-06-23
Contact: Default Subject doesn't work unless a subject field is present


AlexGStapleton:
Introduced as a result of #130 

[This check](https://github.com/siteorigin/so-widgets-bundle/blame/develop/widgets/contact/contact.php#L795) is the only possible time the default subject will be used. The issue here is that It relies on having a subject field that isn't required and has to be empty for the default subject to be used.

The default subject, if applicable, should be used instead of a generated one.

PR #280, opened 2016-07-29
contact widget: add email default email subject if no subject defined in form

dl4ner:

- do not define a subject field
- set a default subject field

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -832,6 +832,11 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 			}
 		}
 
+		// Add in the default subject if no subject field is defined in the form at all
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
Is this basically the solution for #248?

dl4ner:
sorry, I did not have a look at any issues - but as you mention it, yes it is. 

we just had the final finishes on a new site going live on aug/01 and we discovered the above problem and I digged into the code and fixed "our" problem. 
And as it does not seem to have side effects I sended the commit upstream to avoid having a local fork :)

btw. thanks for responding to the PR, thanks for the awesome plugin.

AlexGStapleton:
@gregpriday This looks to be a valid PR. Thoughts?

dl4ner:
sorry, messed up my source repository branch. 
this PR comes again as pr #293 
I will close this.

PR #292, opened 2016-08-10
changing the html markup on the title of features widget

juliotak:
I'm proposing this because a simple accessibility issue. Using headings here, we may have problems if you don't have other headings before H5. Changing to STRONG it will be ok when you do an accessibility check (at http://achecker.ca/ for instance).

### DIFF
--- a/widgets/features/tpl/default.php
+++ b/widgets/features/tpl/default.php
@@ -40,11 +40,11 @@ $last_row = floor( ( count($instance['features']) - 1 ) / $instance['per_row'] )
 
 			<div class="textwidget">
 				<?php if(!empty($feature['title'])) : ?>
-					<h5>
+					<strong>
 						<?php if( !empty( $feature['more_url'] ) && $instance['title_link'] ) echo '<a href="' . sow_esc_url( $feature['more_url'] ) . '" ' . ( $instance['new_window'] ? 'target="_blank"' : '' ) . '>'; ?>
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
I'm not sure of an instance where you wouldn't have a single header before using this widget. Could you provide a use case?

juliotak:
Let me provide you an example: We may have H1, H2, H3 before the widget, but not an H4 heading. The lack of any heading level will result an accesibility issue when we make the validation according the W3C standards. If you give me an <strong> markup at this level, we dont have this issue anymore. Another solution could be to provide a way that allows me to choose the heading level I want. This could be pretty good. 
I can't provide you right now an online example because the plugin is under homologation process in my client. They are the Pan American Health Organization, and accesibility is really important for them, in some cases mandatory. 
I like very much the plugin but I can't use it if I can't solve the acessibility issues.

braamgenis:
Hi @juliotak 

Thanks very much for your contribution. I see your point and understand, but due to the large number of users using this plugin, we need to be careful about changes like this. It is almost certain to cause breakage on sites where the user might, for example, have CSS selectors using the `h5` tag. So to do something like this, you'd need to make it an option by providing a field in the Features widget form. It will also have to default to `h5` so that existing users' sites aren't affected by this change.

Otherwise, feel free to open an issue for us [here](https://github.com/siteorigin/so-widgets-bundle/issues/new) and we'll take a look at making this change as soon as we have a chance.

PR #293, opened 2016-08-15
added default_sender in settings, use case: if no email field is defined.

### LINKED ISSUE #247: Contact: Email always set as required even when no email field is present
I've noticed that certain people use the contact form as a method of gathering phone numbers with no desire of emailing the user. So they remove the field and it doesn't work as desired as the contact form will reject it and state an email is required.

dl4ner:
sorry, I messed up with branches in my local repo, so I had to send this PR again.

this basically should fix #247 
- added a setting for default_sender. 
- check it default sender is set and use it if no email is definined in form.

now you can remove the email addresse from the form 
and add a "from address" that will be used in every case.


dl4ner:
sorry, I messed up with branches in my local repo, so I had to send this PR again.

it is the same as pull #280 which I messed up due to more commits and so I will close them.

again the info:

* do not define a subject field
* set a default subject field

this should basically fix #248

dl4ner:
@AlexGStapleton @gregpriday just updated this PR to 
1) resolv conflicts with current develop
2) conform to wp coding standards (spacing etc.) I hope I fetched all :)

cheers.

widgets/contact/contact.php
```diff
@@ -63,6 +63,12 @@ function get_widget_form(){
 						'description' => __('Where contact emails will be delivered to.', 'so-widgets-bundle'),
 						'sanitize' => 'multiple_emails',
 					),
+					'default_sender' => array(
+						'type' => 'text',
+						'label' => __('Sender email address', 'so-widgets-bundle'),
+						'description' => __('Sender address to be used when there is no email field defined.', 'so-widgets-bundle'),
+						'sanitize' => 'sanitize_email',
```
braamgenis: It's possible that a user might still miss that this is a required field if they don't want an 'email' field in their contact form, so we still need some default value in code to bypass validation requirements in this case.

widgets/contact/contact.php
```diff
@@ -846,13 +852,18 @@ function contact_form_action( $instance, $storage_hash ){
 			}
 		}
 
+		// Add in the default sender if no email field is defined 
+		if ( ! isset( $email_fields['email'] ) && ! empty( $instance['settings']['default_sender'] ) ) {
+			$email_fields['email'] = $instance['settings']['default_sender'];
```
braamgenis: Perhaps here, if `email` and `default_sender` is empty, populate the `email` field with the current user's email address. Similar to what we're doing on [L557-558](https://github.com/dl4ner/so-widgets-bundle/blob/a8240aed0a11b9e0b4745917158f4c609a4fbae4/widgets/contact/contact.php#L557-L558)

dl4ner: hm... 
the behavior before has been, that the email sender has been just the webserver instance. 

I don't like the idea to put the users email here - as a mail from user@domain.com to user@domain.com sent from a webserver could typically create more spam points than www-data@domain.com to user@domain.com.

So somewhere there is already a catchall which puts a "valid" sender adress in the smtp traffic; 
I just wanted to have control to overwrite this if I wanted, so my patch would not change the default behaviour the contact form did have before.

PR #294, opened 2016-08-15
contact widget: add email default email subject if no subject defined in form

### LINKED ISSUE #248: Contact: Default Subject doesn't work unless a subject field is present

dl4ner:
sorry, I messed up with branches in my local repo, so I had to send this PR again.

it is the same as pull #280 which I messed up due to more commits and so I will close them.

again the info:

- do not define a subject field
- set a default subject field

this should basically fix #248

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -832,6 +832,11 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 			}
 		}
 
+		// Add in the default subject if no subject field is defined in the form at all
+		if ( !isset( $email_fields['subject'] ) && !empty($instance['settings']['default_subject']) ) {
+			$email_fields['subject'] = $instance['settings']['default_subject'];
+		}
+
 		// Add in the default subject prefix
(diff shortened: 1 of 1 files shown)


widgets/contact/contact.php
```diff
@@ -832,6 +832,11 @@ function contact_form_action( $instance, $storage_hash ){
 			}
 		}
 
+		// Add in the default subject if no subject field is defined in the form at all
+		if ( !isset( $email_fields['subject'] ) && !empty($instance['settings']['default_subject']) ) {
```
AlexGStapleton: Would it be possible to space the second brackets to conform with [WordPress PHP coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#space-usage)?

gregpriday:
Thanks for the addition @dl4ner!

Issue #287, opened 2016-08-09
Add Dedicated Widget Settings


AlexGStapleton:
With the likes of the [SiteOrigin Widgets Builder](https://github.com/siteorigin/so-widgets-builder) being created it's quite likely a lot of new and exciting third party widgets will start to be created soon. Said widgets may may require specific settings and to prevent cluttering of the main SiteOrigin Page Builder settings we should add a settings interface for widgets on the widgets page.

![](https://i.imgur.com/fiXxxJA.png)

When clicked you would be able to adjust whatever settings the user has specified. For example, for the SiteOrigin Google Maps widget we could allow the user to specificity their API key, thus removing the need to set a key for each individual Google Maps widget (this will replace #249). It doesn't need to be anything fancy as a simple modal would suffice.

Said button wouldn't show if settings where empty.

gregpriday:
Great idea.

gregpriday:
Implemented in https://github.com/siteorigin/so-widgets-bundle/tree/feature/widget-settings - will merge after some tweaks. No widgets use the feature yet, but this is going to be hugely valuable.

AlexGStapleton:
Awesome! :smile:

PR #302, opened 2016-08-22
Widget Settings

gregpriday:
Although no widgets are using this yet, this PR allows widgets to specify settings fields using `initialize_settings_form` function. This just returns a normal form array.

Widgets can get the global settings using `$this->get_global_settings()`. This returns a fairly standard instance array with defaults added.

A good example of the use will be for the Google Maps API key.

This PR resolves GH-287


gregpriday:
Busy testing this and it seems like it doesn't handle repeaters too well. Will try fix that.

gregpriday:
Rad, fixed. This is ready for merging once you've had a chance to take a look @braamgenis.

admin/tpl/admin.php
```diff
@@ -67,6 +67,12 @@
 							<button class="button-secondary so-widget-deactivate" data-status="0"><?php esc_html_e( 'Deactivate', 'so-widgets-bundle' ) ?></button>
 						</div>
 
+						<?php
+						$widget_object = !empty( $widget_objects[$file] ) ? $widget_objects[$file] : false;
+						if( !empty( $widget_object ) && $widget_object->initialize_settings_form() ) {
```
braamgenis: I would have a method `has_settings()` or maybe `has_global_settings()` to try make it a bit clearer what this is checking for.

so-widgets-bundle.php
```diff
@@ -534,6 +534,32 @@ function get_widgets_list(){
 	}
 
 	/**
+	 * Get instances of all the widgets. Even ones that are not active.
+	 */
+	function get_widget_objects(){
```
braamgenis: This could be quite a heavy call, couldn't it? Should the function perhaps be scoped private to try avoid abuse?

admin/admin.less
```diff
@@ -307,10 +307,211 @@
 		font-style: italic;
 	}
 
+	#sow-settings-dialog {
```
braamgenis: Isn't all of this duplication of CSS we already have somewhere else?

admin/admin.js
```diff
@@ -122,6 +122,41 @@ jQuery(function($){
...
     });
 
-    // Finally enable css3 animations on the widgets list
+    // Enable css3 animations on the widgets list
     $('#widgets-list').addClass('so-animated');
-});
\ No newline at end of file
+
+    // Handle the dialog
+    var dialog = $('#sow-settings-dialog');
```
braamgenis: Same with this code. Haven't we already done this somewhere?

so-widgets-bundle.php
gregpriday: Yeah, very good idea.

admin/admin.less
gregpriday: This is the problem with having all our plugins that work together but need to function independently. Not sure on a good alternative, but at least these won't both be enqueued on the same page. This `admin.less` is only used on the Widgets Bundle admin page _Plugins > SiteOrigin Widgets_

admin/admin.js
gregpriday: Not really. Most of this code is fairly specific to the Widgets Bundle settings dialog. It submits to an iframe while the Page Builder dialog stores to a JS object. Similar, but it would be very difficult to reuse.

so-widgets-bundle.php
```diff
@@ -351,6 +356,59 @@ function admin_ajax_manage_handler(){
...
 	/**
+	 *
+	 */
+	function admin_ajax_settings_form(){
+		$widget_objects = $this->get_widget_objects();
+		$widget_object = !empty( $widget_objects[ $_GET['id'] ] ) ? $widget_objects[ $_GET['id'] ] : false;
+
+		if( empty( $widget_object ) ) exit();
+
+		$form_options = $widget_object->initialize_settings_form();
```
braamgenis: Seems a bit weird to get the `$form_options` from the `$widget_object` and then pass it back in further down the line, without actually doing anything to `$form_options`. Here a method for `has_gloabl_settings()` would be useful again.

gregpriday: Good point. On line 388 we'd still need to pass in the form options, but we could just use a line line

```
$widget_object->form( $value, $widget_object->initialize_settings_form() )
```

Alternatively, we could have a second function called `settings_form( $value )`, which would just be a wrapper to that call. That actually feels the neatest to me.

base/js/admin.js
```diff
@@ -615,19 +616,22 @@
                     var $in = $(input);
 
                     if(typeof pos !== 'undefined') {
-                        var newName = $in.data('original-name');
+                        var newName = $in.attr('data-original-name');
```
braamgenis: I'm confused. I thought the jQuery `data()` method effectively created an attribute with the given name and prefixed with `data-`. I'll have to go through the code to see what the difference is there.

gregpriday: The difference is that the attribute survives across duplication. There was a problem when duplicating a repeater item, for the widget settings form. Oddly, I wasn't getting the same issue when using the form in a standard widget form though.

braamgenis:
Rad. Just commented on a few **very** minor things. No big issues.

We should just test repeaters very thoroughly as they tend to be quite prone to breakage with any minor changes.

ehrig:
Can you provide a code snippet how I would go about adding a global widget setting for the Google Maps API key? Any new Google Maps widget I add to a page will be pre-populated with that global API key or how do the existing Google Maps widget API keys that I have copy+pasted to all individual Google Maps widgets play together with a globally set Google Maps API key?

PR #320, opened 2016-09-14
SiteOrigin_Widget: Fix use of wrong variable

leeps:
Actually use `$control_options` instead of `$widget_options`
for passing control options to parent constructor.

Original code has been in place since Nov 2014, so this fix could possibly break workarounds in other places. A compatibility fix is possible if deemed necessary.


gregpriday:
Very nice fix @leeps. There shouldn't be any workaround code for this anywhere, so I don't imagine this fix will cause any problems.

leeps:
Thnx :)

PR #321, opened 2016-09-19
Add margin to recaptcha

### LINKED ISSUE #199: Contact Form: Add bottom margin to reCAPTCHA
For all intents and purposes, the reCAPTCHA is a form field but as the field isn't given the sow-form-field class it doesn't have a bottom margin applied (1em) and thus, you end up with very small spacing between the reCAPTCHA and the submit button.

For reference, here is the form without:

And with:

The possible fix for this is to either add sow-form-field to the container div or apply a bottom margin of 1em to sow-recaptcha.

AlexGStapleton:
Resolves #199, it might be a better add sow-form-field to the recaptcha.

### DIFF
--- a/widgets/contact/styles/default.less
+++ b/widgets/contact/styles/default.less
@@ -98,6 +98,10 @@
 	}
 }
 
+.sow-recaptcha {
+	margin-bottom: 1em;
+}
+
 .sow-contact-form {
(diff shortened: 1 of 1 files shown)


gregpriday:
Thank you, good sir.

PR #325, opened 2016-09-22
Removed leading slash in paths after plugin_dir_url()

brandomeniconi:
Google AppEngine does not auto-correct double slashes in a URL, the following one result in a 404 error:

`so-widgets-bundle/base/inc/fields//js/media-field.min.js?ver=1.6.5`

plugin_dir_url() already outputs the trailing slash so it's not needed in the string path.

### DIFF
--- a/base/inc/fields/code.class.php
+++ b/base/inc/fields/code.class.php
@@ -31,6 +31,6 @@ class SiteOrigin_Widget_Field_Code extends SiteOrigin_Widget_Field_Text_Input_Ba
 	}
 
 	function enqueue_scripts(){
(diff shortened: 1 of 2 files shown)


AlexGStapleton:
Good find.

gregpriday:
Thanks @brandomeniconi! I always make that trailing slash mistake. It's [get_template_directory_uri](https://developer.wordpress.org/reference/functions/get_template_directory_uri/) that doesn't return a trailing slash.

brandomeniconi:
Same problem here @gregpriday, need to check on Codex every time 😄.  Thanks guys

PR #326, opened 2016-09-22
Replaced glob() function with PHP5 SPL Iterators

brandomeniconi:
Some hosting disable `glob` function or in restricted environments, like Google AppEngine, it doesn't work like expected. I propose to replace `glob` with the SPL Directory Iterators.

Tested in Google AppEngine.

### DIFF
--- a/so-widgets-bundle.php
+++ b/so-widgets-bundle.php
@@ -569,8 +569,18 @@ class SiteOrigin_Widgets_Bundle {
 		$widgets = array();
 		foreach( $folders as $folder ) {
 
-			$files = glob( $folder . '*/*.php' );
-			foreach($files as $file) {
+			$Directory = new RecursiveDirectoryIterator($folder);
+			$Iterator = new RecursiveIteratorIterator($Directory);
+			$files = new RegexIterator($Iterator, '/^.+\.php$/i', RecursiveRegexIterator::GET_MATCH);
+			
+			foreach($files as $file_r) {
+				
+				if(is_array( $file_r)){
+					$file = $file_r[0];
+				} else {
+					$file = $file_r;
+				}
+				
 				$widget = get_file_data( $file, $default_headers, 'siteorigin-widget' );
 				//skip the file if it's missing a name
 				if ( empty( $widget['Name'] ) ) {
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
Interesting find. Would it not be a good idea to maintain the usage of glob() in instances where it's applicable rather than replacing it entirely?

brandomeniconi:
That's something I was thinking about, but in my case I'm running Wordpress in Google AppEngine where the glob function is available but in this case doesn't return anything, breaking plugin functionality. I run also a little raw benchmark and Iterators seems faster:

> Found 623 files in 0.0002 seconds with iterators
> Found 621 files in 0.0009 seconds with glob
> Found 1214 files in 0.0002 seconds with recursive iterators
> Found 1214 files in 0.0011 seconds with recursive glob

Iterators also are safer, easier to understand and provide more file information (for future use). Maybe we can use Iterators as main method and fallback to glob for environments that don't have PHP SPL installed (PHP <4).

AlexGStapleton:
Thank you for clarification and running a benchmark. I don't care about PHP 4 (I would be shocked if it would run anyway) so I wouldn't worry about adding a fallback at this time. I'll send Greg a message about this to see what he thinks. 👍

AlexGStapleton:
Just thought of something. Would it be possible to update your code with the [WordPress PHP coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/) in mind? We're really trying to push for that with all new commits so this would really help us out - even more, you've been a huge help! 😄

brandomeniconi:
Ok, I've updated the code with a static function that contains all the file listing logic. Should be easier to add fallbacks in the future. Added an SPL extension check to avoid fatal errors. Tried to follow all the WP coding standards :bowtie: . If you have any other suggestion, feel free to ask.

so-widgets-bundle.php
```diff
@@ -232,6 +232,25 @@ function get_widget_folders(){
 	}
 
 	/**
+	* Get all files in a folder
+	*/
+	static function get_folder_files( $folder, $regex = '^.+\.php$', $max_depth = -1 ){
+
+		if( ! extension_loaded("SPL") ){
+			return array();
```
AlexGStapleton: So if SPL isn't loaded, it'll return nothing? Maybe use blog anyway?

brandomeniconi:
Updated code with glob fallback.

AlexGStapleton:
Thanks for that. Last thing, could you lower the amount empty lines in the glob fallback? A couple of them I can understand for spacing, but a few of them, well, don't really make sense.

brandomeniconi:
sorry, I missed that comment while ago. Empty lines removed.

AlexGStapleton:
Thanks mate! 👍 

Just forwarded this to Greg.

gregpriday:
Huge apologies for letting this fall through the cracks @brandomeniconi. I'm going through older PRs and just came across this again.

I'm going to close this for now because there are merge conflicts and because I'd rather not change away from `glob` at this stage. There is the possibility that this would introduce unintended bugs.

As for App Engine, I know from personal experience that it's insanely difficult getting WordPress running properly there. I think this is why App Engine introduced their flexible environment, which gives you actual instances where you can install what you want. This is fairly similar to how Elastic Beanstalk works.

The older fixed and limited environments seem to cause issues with most open source systems, so they're not widely used for that. Good evidence is the [App Engine WordPress plugin](https://wordpress.org/plugins/google-app-engine/), which Google recommends to get WordPress working on App Engine. It only has ~300 active installs.

So I don't really think it's worth solving this edge case, especially in 2019.

PR #327, opened 2016-09-26
Headline: Remove !important from styling

### LINKED ISSUE #275: Headline: Remove !important from styling
We're already using [highly specific selectors](https://i.imgur.com/Xlz9CdI.png) there's little reason to !important it. This also prevents other designers from making adjustments to the margins.

[Reference in default.less](https://github.com/siteorigin/so-widgets-bundle/blame/develop/widgets/headline/styles/default.less#L70)

AlexGStapleton:
Resolves #275

We're already using highly specific selectors there's little reason to !important it. As doing so just prevents other designers from making adjustments to the margins with CSS without writing a page long selector.

### DIFF
--- a/widgets/headline/styles/default.less
+++ b/widgets/headline/styles/default.less
@@ -68,10 +68,10 @@
     }
 
     *:first-child {
-        margin-top: 0 !important;
+        margin-top: 0;
     }
 
     *:last-child {
-        margin-bottom: 0 !important;
+        margin-bottom: 0;
     }
 }


gregpriday:
I just need to double check the reasons I had for using `!important`. I'm not 100% sure, but I think a selector like `@{headline_tag}.sow-headline` would land up being more specific than `*:first-child`. Is this not the case?

AlexGStapleton:
It will yes. (_anything_ is more specific than *) I'm guessing that's undesired? Is there a specific thing that CSS is meant to be applied too?

gregpriday:
It's meant to be applied to whichever element happens to be on the top. It was an important change since we introduced ordering to the headline widget. Have any issues come up with this selector?

gregpriday:
@AlexGStapleton - I'm going to close this for now. We can always reopen if it causes any problems.

PR #328, opened 2016-09-26
Image Widget: use alt / title set via WP if not set

### LINKED ISSUE #210: Image: if alt text field isn't set, detect if image has alt text set and output that instead of nothing
Right now there's no way to output the alt text set specifically for an image in WordPress unless you set it for that widget. This can mislead people into thinking that their alt text will be output due to previously setting when in reality this isn't the case.

References:

"Attachment Details"

AlexGStapleton:
Resolves #210

This PR will allow for the image widget image to use the title / alt set in WordPress if the widget values aren't set.

### DIFF
--- a/widgets/image/image.php
+++ b/widgets/image/image.php
@@ -102,16 +102,13 @@ class SiteOrigin_Widget_Image_Widget extends SiteOrigin_Widget {
 	}
 
 	public function get_template_variables( $instance, $args ) {
-
-		$existing_alt = get_post_meta( $instance['image'], '_wp_attachment_image_alt', true );
-
 		return array(
 			'title' => !empty( $instance['title'] ) ? $instance['title'] : get_the_title( $instance['image'] ),
 			'title_position' => $instance['title_position'],
 			'image' => $instance['image'],
 			'size' => $instance['size'],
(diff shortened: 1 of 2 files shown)


gregpriday:
This is a very nice addition. Although, we should have as much of the logic as possible in `get_template_variables` function - https://github.com/siteorigin/so-widgets-bundle/blob/develop/widgets/image/image.php#L104-L115

So in this case, something along the lines of: `'title' => !empty( $instance['title'] ) ? $instance['title'] : get_the_title( $image ),`. This would be a lot neater.

widgets/image/image.php
```diff
@@ -102,18 +102,22 @@ function get_style_hash($instance) {
 	}
 
 	public function get_template_variables( $instance, $args ) {
+
+		$existing_alt = get_post_meta( $instance['image'], '_wp_attachment_image_alt', true );
```
gregpriday: This variable isn't strictly necessary. You could just include the `get_post_meta` call in line 114.

AlexGStapleton: Oh. Just saw this. Don't ask me why I did this.

PR #336, opened 2016-10-14
Do not compile empty less stylesheet

pravdomil:
Widget less styles are filtered:

```
$less = apply_filters( 'siteorigin_widgets_styles', $less, $this->widget_class, $instance );
$less = apply_filters( 'siteorigin_widgets_less_' . $this->id_base, $less, $instance, $this );
```

If some filter returns empty string there is no reason to start compiler.

### DIFF
--- a/base/siteorigin-widget.class.php
+++ b/base/siteorigin-widget.class.php
@@ -818,7 +818,7 @@ abstract class SiteOrigin_Widget extends WP_Widget {
 				}
 			}
 			catch ( Exception $e ) {
-				$css = '';
+				
 			}
 
 			// Remove any attributes with default as the value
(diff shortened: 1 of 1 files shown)


base/siteorigin-widget.class.php
```diff
@@ -792,38 +792,42 @@ public function get_instance_css( $instance ){
 		$less = apply_filters( 'siteorigin_widgets_styles', $less, $this->widget_class, $instance );
 		$less = apply_filters( 'siteorigin_widgets_less_' . $this->id_base, $less, $instance, $this );
 
-		$style = $this->get_style_name( $instance );
-		$hash = $this->get_style_hash( $instance );
-		$css_name = $this->id_base . '-' . $style . '-' . $hash;
+		$css = '';
```
AlexGStapleton: Isn't this a tad redundant due to line 821?

AlexGStapleton:
Outside of that, I see no issues with this and think it would be a good change. 👍

base/siteorigin-widget.class.php
pravdomil: you are right

pravdomil:
fixed, it's ready to merge

PR #344, opened 2016-10-25
Removed double slash from icon form field

AlexGStapleton:
Same issue as #325

"Google AppEngine does not auto-correct double slashes in a URL, [...] result[ing] in a 404 error"


gregpriday:
Double slashes will be the death of me.

PR #348, opened 2016-11-02
Throw less compiler errors in debug mode

pravdomil:
very useful

### DIFF
--- a/base/siteorigin-widget.class.php
+++ b/base/siteorigin-widget.class.php
@@ -819,7 +819,9 @@ abstract class SiteOrigin_Widget extends WP_Widget {
 				}
 			}
 			catch ( Exception $e ) {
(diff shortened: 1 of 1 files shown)


pravdomil:
well I just copy paste from https://github.com/pravdomil/so-widgets-bundle/blob/184c66d6b955fa2c7755f4c5bb2fa6ff022fe655/base/siteorigin-widget.class.php#L260

AlexGStapleton:
Ah. Yeah, there's quite a lot of older code that still doesn't comply with the WordPress PHP coding standards but we're trying to ensure that all new code does.

Thanks mate - I'll let Greg know about this.

gregpriday:
Awesome addition, thanks @pravdomil

PR #351, opened 2016-11-09
Contact: Prevent empty title markup from being output

AlexGStapleton:
If Display Title is ticked but the contact title is empty the contact title will still be output. There's a reason to have the title not display, but there's no real reason to have the markup present if it's not being used.


braamgenis:
All good. :+1:

PR #354, opened 2016-11-14
Slider: Account for 0 speed

### LINKED ISSUE #108: Slider widget. Other slides won't be displayed if animation speed is set to 0.
I had a support thread where a user didn't want the "dimming effect". To remove it I thought the most obvious answer was to simply set the animation speed to 0 (translation; none). This, however, broke the animation effect and only made the first frame visible (out of two possible frames). An empty value works with the desired effect.

This feels like a PHP consistency issue more than a bug report.

AlexGStapleton:
Resolves #108.

Setting it to 0 prevents the slider from working correctly.

### DIFF
--- a/base/inc/widgets/base-slider.class.php
+++ b/base/inc/widgets/base-slider.class.php
@@ -137,7 +137,7 @@ abstract class SiteOrigin_Widget_Base_Slider extends SiteOrigin_Widget {
 	function slider_settings( $controls ){
 		return array(
 			'pagination' => true,
-			'speed' => $controls['speed'] != 0 ? $controls['speed'] : 1,
+			'speed' => $controls['speed'] !== 0 ? $controls['speed'] : 1,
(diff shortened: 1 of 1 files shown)


base/inc/widgets/base-slider.class.php
```diff
@@ -137,7 +137,7 @@ function video_form_fields(){
 	function slider_settings( $controls ){
 		return array(
 			'pagination' => true,
-			'speed' => $controls['speed'],
+			'speed' => $controls['speed'] !== 0 ? $controls['speed'] : 1,
```
gregpriday: It would be neater using an `empty` here. That would cover `0` and a few other possible edge cases. Not essential, but worth changing unless it'll cause any problems you can think of.

AlexGStapleton:
How's that? I couldn't think of any problems - there was little reason not too to be honest.

PR #357, opened 2016-11-21
Contact: Improved type Validation and added empty field check

AlexGStapleton:
Added empty field check to prevent empty fields from overwriting valid fields (for example, two name fields). This will also prevent empty fields from being added to the email.

Changed the type validation to a single `switch()`.

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -804,19 +804,18 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 			$field_name = $this->name_from_label( !empty($field['label']) ? $field['label'] : $i, $field_ids ) . '-' . $instance['_sow_form_id'];
 			$value = !empty( $post_vars[$field_name] ) ? $post_vars[$field_name] : '';
 
-			if( $field['required']['required'] && empty($value) ) {
-				// Add in the default subject
-				if( $field['type'] == 'subject' && !empty($instance['settings']['default_subject'] ) ) {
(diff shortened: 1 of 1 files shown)


widgets/contact/contact.php
```diff
@@ -813,36 +813,38 @@ function contact_form_action( $instance, $storage_hash ){
...
 
+			// Type Validation
 			switch( $field['type'] ) {
 				case 'email':
-					if( $value != sanitize_email($value) ) {
-						$errors[$field_name] = __('Invalid email address.', 'so-widgets-bundle');
+					if( $value != sanitize_email( $value ) ) {
+						$errors[ $field_name ] = __( 'Invalid email address.', 'so-widgets-bundle' );
 					}
+
```
braamgenis: Great idea to tidy up this `switch`, but I'm not really a fan of this style of allowing a `case` to fall through after the `case` is partially handled. It's not clear whether the `break` was left out accidentally or intentionally and it could have unintended effects. I would either repeat the line `$email_fields[ $field['type'] ] = $value;` in the `'email'` case or group the three cases together and have an additional check for the `'email'` type. I.e.
```
	case 'email':
	case 'name':
	case 'subject':
		if( $field['type'] === 'email' && $value != sanitize_email($value) ) {
			$errors[$field_name] = __('Invalid email address.', 'so-widgets-bundle');
		}
		$email_fields[$field['type']] = $value;
		break;
```

braamgenis:
Change approved!

gregpriday:
🎉

PR #362, opened 2016-11-29
CTA: Add design setting to change title and subtitle colors

### LINKED ISSUE #254: CTA: Add design setting to change text color
Oddly you're able to change the button text color but you're unable to change the text color.

You should only be able to change the overall text color. As subtitle has opacity set it would be confusing to some (why is the color slightly different?) if you could specifically set the color for the subtitle.

AlexGStapleton:
Resolves #254 

This PR will allow users to change the title and subtitle font. It doesn't have a default defined to allow the theme styling to be taken in. It also lays the groundwork for adding support for different header levels (if desired).

### DIFF
--- a/widgets/cta/styles/default.less
+++ b/widgets/cta/styles/default.less
@@ -1,8 +1,8 @@
 
 @border_color: #E3E3E3;
 @background_color: #F8F8F8;
-@title_color: '';
-@subtitle_color: '';
+@title_color: default;
+@subtitle_color: default;
 
 .sow-cta-base {
 	border-color: @border_color;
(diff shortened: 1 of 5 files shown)


widgets/cta/styles/default.less
```diff
@@ -1,12 +1,25 @@
 
 @border_color: #E3E3E3;
 @background_color: #F8F8F8;
+@title_color: '';
```
gregpriday: The widgets bundle supports variables of the form `@title_color: default`. This removes any attributes that are still at default. So they'll just use whatever the theme specifies.

You can get a better idea of how we use this in the Headline widget https://github.com/siteorigin/so-widgets-bundle/tree/develop/widgets/headline

PR #364, opened 2016-11-30
Update posts-selector.js to fix #360

### LINKED ISSUE #360: Posts field shows previews data on empty post__in and tax_query
Using some widget with posts field clicking on **build posts query** button shows right data in all fields.
If you close query builder view and click on **build posts query** button on another widget with empty post__in or tax_query, it shows previews values of widget post__in and tax_query. 
The problem is only for this vars probably because they are array in the backbone model.
As workaround I make changes in posts-selector.js at row [57](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/js/posts-selector.js#L57), [58](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/js/posts-selector.js#L58), [83](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/js/posts-selector.js#L83), [84](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/js/posts-selector.js#L84),[124](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/js/posts-selector.js#L124) and [125](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/js/posts-selector.js#L125) as follow.
This doesn't work with existing widgets.

```
post__in: [],
tax_query: [],
...
if( typeof this.get('post__in') !== 'undefined' && !_.isEmpty( this.get('post__in') ) ) query.push( 'post__in=' + this.get('post__in').join(',') ); else query.push('post__in=')
if( typeof this.get('tax_query') !== 'undefined' && !_.isEmpty( this.get('tax_query') ) ) query.push( 'tax_query=' + this.get('tax_query').join(',') );else query.push('tax_query=')
...
if( params.hasOwnProperty('post__in') ) theQuery.post__in = params.post__in.split(',');else theQuery.post__in=[]
if( params.hasOwnProperty('tax_query') ) theQuery.tax_query = params.tax_query.split(',');else theQuery.tax_query =[]
            
```

giuseppebonanno:
Fix posts field shows previews data on empty post__in and tax_query #360

### DIFF
--- a/base/js/posts-selector.js
+++ b/base/js/posts-selector.js
@@ -54,8 +54,8 @@ var soWidgetPostSelector = ( function ($, _) {
         terms: null,
         post_status: null,
         posts_per_page: null,
-        post__in: null,
-        tax_query: null,
+        post__in: [],
+		tax_query: [],
         date_range: null,
 
         // The order fields for get_posts.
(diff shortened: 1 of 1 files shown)


braamgenis:
Hi @giuseppebonanno

Thanks very much for your contribution! We should really just ensure that a new `Query` model is set when the `QueryBuilder` view is reused.

I've updated the priority on #360 for this bug, so we'll be sure to fix it as soon as possible. Thanks again for reporting the issue! :+1:

PR #365, opened 2016-11-30
Google Font import: Replace is_ssl() with a HTTPS check of site_url()

### LINKED ISSUE #277: Hero Widget: Google Fonts https problem
The Hero widget imports a Google font as follows:

`@import url(http://fonts.googleapis.com/css?family=Dosis);`

This isn't ideal when https is being used. Let's update the import method to allow for http and https.

Thanks :)

AlexGStapleton:
Resolves #277

Using is_ssl() is fine for most occurrences but when the website allows both http and https it falls flat if the version stored in the cache wasn't secure at the time of generation.

### DIFF
--- a/base/base.php
+++ b/base/base.php
@@ -105,7 +105,7 @@ function siteorigin_widget_get_font($font_value) {
 			$font['weight'] = $font_parts[1];
 			$font_url_param .= ':' . $font_parts[1];
 		}
-		$font['css_import'] = '@import url(http' . ( is_ssl() ? 's' : '' ) . '://fonts.googleapis.com/css?family=' . $font_url_param . ');';
+		$font['css_import'] = '@import url(http' . ( strpos( site_url(), 'https' ) !== false ? 's' : '' ) . '://fonts.googleapis.com/css?family=' . $font_url_param . ');';
 	}
 	else {
 		$font['family'] = $font_value;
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
@Misplon Can you test this one out to see if it resolves #277 for you? 😃

base/base.php
```diff
@@ -105,7 +105,7 @@ function siteorigin_widget_get_font($font_value) {
 			$font['weight'] = $font_parts[1];
 			$font_url_param .= ':' . $font_parts[1];
 		}
-		$font['css_import'] = '@import url(http' . ( is_ssl() ? 's' : '' ) . '://fonts.googleapis.com/css?family=' . $font_url_param . ');';
+		$font['css_import'] = '@import url(http' . ( strpos( site_url(), 'https' ) !== false ? 's' : '' ) . '://fonts.googleapis.com/css?family=' . $font_url_param . ');';
```
braamgenis: We should just always use `https` as Google will "always" serve their assets over `https`. It removes the complexity and any potential bugs. And it won't cause any issues to use an `https` request on a site without SSL enabled.

AlexGStapleton: Actually. That does make more sense in hindsight. :S

Misplon:
Haven't tested but looks good. Braam's comment spot on. Rad, let's do it.

PR #366, opened 2016-12-03
Post Selector: Exclude current post id

AlexGStapleton:
Exclude the current post (if applicable) to avoid any issues associated with showing the same post again.

Use case: A third party plugin doesn't handle things correctly and lets SiteOrigin Page Builder handle the post. This, in turn, will leed to an infinite loop.

I can't think of a valid reason as to why you would want the current post to appear so I feel it's safe to disregard it completely.

### DIFF
--- a/base/inc/post-selector.php
+++ b/base/inc/post-selector.php
@@ -93,6 +93,11 @@ function siteorigin_widget_post_selector_process_query($query){
 		}
 		unset( $query['sticky'] );
 	}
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
This is actually redundant as [it's done in the base widget](https://github.com/siteorigin/siteorigin-panels/blob/develop/widgets/basic.php#L251).

PR #369, opened 2016-12-06
Slider Background Video: Try embedding the video if oEmbed fails

### LINKED ISSUE #246: Video: Self Hosted video plays when video widget modal closes after adding first video
- Add SiteOrigin Video Player Widget and open it
- Set Video Location to self-hosted and add a video (only one required)
- Tick Autoplay in video playback settings
- Click done and video will start playing in background.

This only seems to happen once

### LINKED ISSUE #359: Slider Video: Add fallback embed option if oembed fails
Currently, the slider video background external video functionality only supports [oEmbed supported sites](https://codex.wordpress.org/Embeds#Okay.2C_So_What_Sites_Can_I_Embed_From.3F). As per [this line](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/inc/widgets/base-slider.class.php#L323).

This is an odd limitation and we should try and remove it to support both oEmbed and non-oEmbed videos (such as AWS).

[Slack Reference](https://siteorigin.slack.com/archives/plugins/p1479524934000002).

AlexGStapleton:
Resolves #359.

@gregpriday Would it be a good idea to check the video URL for the selected file type? It could be an in for adding third-party JavaScript.

### DIFF
--- a/base/inc/widgets/base-slider.class.php
+++ b/base/inc/widgets/base-slider.class.php
@@ -309,19 +309,27 @@ abstract class SiteOrigin_Widget_Base_Slider extends SiteOrigin_Widget {
 
 		foreach($videos as $video) {
 			if( empty( $video['file'] ) && empty ( $video['url'] ) ) continue;
-
-			if( empty( $video['url'] ) ) {
-				$video_file = wp_get_attachment_url($video['file']);
-				$video_element .= '<source src="' . sow_esc_url( $video_file ) . '" type="' . esc_attr( $video['format'] ) . '">';
-			}
-			else {
-				$args = '';
+			// If video is an external file, try and display it using oEmbed
+			if( !empty( $video['url'] ) ) {
 				if ( ! empty( $video['height'] ) ) {
 					$args['height'] = $video['height'];
 				}
+				$embeded_video = wp_oembed_get( $video['url'], $args );
 
-				echo wp_oembed_get( $video['url'], $args );
+				// Check if we can oEmbed the video or not
+				if( !$embeded_video ) {
+					$video_file = sow_esc_url( $video['url'] );
+				}else{
+					echo $embeded_video;
+					continue;
+				}
 			}
+
+			// If $video_file isn't set video is a local file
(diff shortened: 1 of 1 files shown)


braamgenis:
@AlexGStapleton Do you have an example video hosted on AWS or any other non-oEmbed provider? Did a quick search, but couldn't find any.

AlexGStapleton:
Not off by chance, but this Google Drive video should surface: https://drive.google.com/file/d/0BymU-6VMHHPfUTV6REJSaWtKZmM/preview

base/inc/widgets/base-slider.class.php
```diff
@@ -309,19 +309,27 @@ function video_code( $videos, $classes = array() ){
 
 		foreach($videos as $video) {
 			if( empty( $video['file'] ) && empty ( $video['url'] ) ) continue;
-
-			if( empty( $video['url'] ) ) {
-				$video_file = wp_get_attachment_url($video['file']);
-				$video_element .= '<source src="' . sow_esc_url( $video_file ) . '" type="' . esc_attr( $video['format'] ) . '">';
-			}
-			else {
-				$args = '';
```
braamgenis: I think we should keep this around for the case where we don't have `$video['height']` defined. It should perhaps rather be initialized to an empty `array()`.

base/inc/widgets/base-slider.class.php
```diff
@@ -309,19 +309,27 @@ function video_code( $videos, $classes = array() ){
...
+			// If video is an external file, try and display it using oEmbed
+			if( !empty( $video['url'] ) ) {
 				if ( ! empty( $video['height'] ) ) {
 					$args['height'] = $video['height'];
 				}
+				$embeded_video = wp_oembed_get( $video['url'], $args );
```
braamgenis: Small thing, but spelling should be `$embedded_video`

AlexGStapleton: 😢

AlexGStapleton:
How's that? @braamgenis

PR #370, opened 2016-12-06
Feature responsive global setting

### LINKED ISSUE #298: Features: Change media query to reflect site settings

AlexGStapleton:
Resolves #298.

Basically, the responsive CSS is set to 520px - which causes problems as the breakpoint is too low. I feel it should instead be set to the site mobile breakpoint and this PR does that.


AlexGStapleton:
Ah, crap. I used the forks PR. 😞 I'll resubmit this.

PR #371, opened 2016-12-07
Google Maps: Add ability to set link for Static Map

AlexGStapleton:
This PR adds the ability to set a link for the static map. Includes option to open in new window.

### DIFF
--- a/widgets/google-map/tpl/static-map.php
+++ b/widgets/google-map/tpl/static-map.php
@@ -1,9 +1,10 @@
-<?php
-if( !empty( $destination_url ) ) echo '<a href="' . $destination_url . '" ' . ( $new_window ? 'target="_blank"' : '' ) . '>'; ?>
+<?php if( !empty( $destination_url ) ): ?>
+<a href="<?php echo sow_esc_url( $destination_url ) ?>" <?php echo $new_window ? 'target="_blank"' : '' ?>>
+<?php endif; ?> 
 
 <img
 	border="0"
-	src="<?php echo $src_url ?>">
+	src="<?php echo sow_esc_url( $src_url ) ?>">
 
 <?php
 if( !empty( $destination_url ) ) echo '</a>';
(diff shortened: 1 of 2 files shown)


widgets/google-map/tpl/static-map.php
```diff
@@ -1,3 +1,9 @@
+<?php
+if( !empty( $destination_url ) ) echo '<a href="' . $destination_url . '" ' . ( $new_window ? 'target="_blank"' : '' ) . '>'; ?>
+
 <img
 	border="0"
-	src="<?php echo sow_esc_url( $src_url ) ?>">
```
gregpriday: In general, we want to [escape as late as possible](http://www.wptherightway.org/en/ssl/). Let's remove escaping from `get_template_variables ` and add it to the template files.

widgets/google-map/tpl/static-map.php
```diff
@@ -1,3 +1,9 @@
+<?php
+if( !empty( $destination_url ) ) echo '<a href="' . $destination_url . '" ' . ( $new_window ? 'target="_blank"' : '' ) . '>'; ?>
```
gregpriday: For this sort of template stuff, keeping HTML as HTML and using alternative control structure makes for neater (and more consistent) template files. Echoing HTML tags shouldn't happen often, especially in template files.

AlexGStapleton: Hm. So would something more like:

```
<?php if( !empty( $destination_url ) ){ ?>
<a href="<?php echo $destination_url ?>" <?php ( $new_window ? $new_window : '' ); ?>>
<?php } ?> "

```

Or do you want me to try and offload it so it comes "pre-setup" when passed from get_template_variables?

gregpriday: Something more like :

```
<?php if( !empty( $destination_url ) ) : ?>
<a href="<?php echo sow_esc_url( $destination_url ) ?>" <?php echo $new_window ? 'target="_blank"' : '' ) ?>>
<?php endif; ?>
```

Where `$new_window` should just be a boolean.

AlexGStapleton:
Ah, crap. Labeled the commit incorrectly. :/

PR #372, opened 2016-12-07
Features: Change media query to reflect site settings

### LINKED ISSUE #298: Features: Change media query to reflect site settings

AlexGStapleton:
Resolves #298.

Basically, the responsive CSS is set to 520px - which causes problems as the breakpoint is too low. I feel it should instead be set to the site mobile breakpoint and this PR does that.

--
This is the corrected version of #370

### DIFF
--- a/widgets/features/features.php
+++ b/widgets/features/features.php
@@ -253,7 +253,7 @@ class SiteOrigin_Widget_Features_Widget extends SiteOrigin_Widget {
 			}
 		}
 
-		$less_vars['responsive_breakpoint'] = siteorigin_panels_setting('mobile-width') . 'px';
+		$less_vars['responsive_breakpoint'] = ( function_exists( 'siteorigin_panels_setting' ) ?  siteorigin_panels_setting( 'mobile-width' ) : 520 ) . 'px';
 		return $less_vars;
 	}
 
--- a/widgets/features/features.php
+++ b/widgets/features/features.php
@@ -253,6 +253,7 @@ class SiteOrigin_Widget_Features_Widget extends SiteOrigin_Widget {
 			}
 		}
 
+		$less_vars['responsive_breakpoint'] = siteorigin_panels_setting('mobile-width') . 'px';
 		return $less_vars;
 	}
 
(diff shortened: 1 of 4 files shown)


widgets/features/features.php
```diff
@@ -253,6 +253,7 @@ function get_less_variables( $instance ) {
 			}
 		}
 
+		$less_vars['responsive_breakpoint'] = siteorigin_panels_setting('mobile-width') . 'px';
```
gregpriday: We just need to deal with the condition where Page Builder isn't being used. This might be a rare case now, but it'll become more common with the work we've done on getting Widgets Bundle to play-nice with more page builder plugins.

Right now, this will cause a fatal error if Page Builder isn't installed.

AlexGStapleton: So, uh, how do I handle that? Do I just default to what it was previously or is there a setting I can pull from?

gregpriday: Something along the lines of 

`$less_vars['responsive_breakpoint'] = ( function_exists( 'siteorigin_panels_setting' ) ? siteorigin_panels_setting('mobile-width') : 520 ) . 'px';`

AlexGStapleton: Okay, so just default to 520 - gotcha. I'll keep in this mind for future edits and I'll submit a commit now.

AlexGStapleton:
Maybe we should introduce a filter in a future version so people can override siteorigin_panels_setting rather than "locking" or "defaulting" the setting itself.

braamgenis:
I realize I'm a bit late to the party on this conversation :sweat_smile: , but perhaps we can have responsive layout settings for the widget and leave it up to the user. For an example see the Masonry widget's form, where we allow the user to set desktop, tablet and mobile layouts giving them complete control, specific to the widget's requirements. It also eliminates any dependency on Page Builder settings, whether it's ours or a third party page builder.

AlexGStapleton:
Maybe we could make use global widget settings? That will allow user's to set it once and forget about it and... if need be, override it on a widget by widget basis.

braamgenis:
Yeah, I agree. I'd just add it as global widget settings for the features widget then and if we have a lot of requests for instance specific customization, or breakpoint settings for other widgets we can add those as required.

AlexGStapleton:
Resolved in favour of #637

PR #383, opened 2016-12-28
Contact: Add radio field Type

AlexGStapleton:


widgets/contact/fields/radio.class.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+class SiteOrigin_Widget_ContactForm_Field_Radio extends SiteOrigin_Widget_ContactForm_Field_Base {
+
+	public function render_field( $options ) {
+
+		if ( ! empty( $options['field']['options'] ) ) {
+			?><ul><?php
```
gregpriday: This just needs a small style improvement. It should be using the [alternative control structure](http://php.net/manual/en/control-structures.alternative-syntax.php) because it's a mix of HTML and PHP. This makes the code look a lot neater and you can keep the format more HTML friendly.

AlexGStapleton: Ah. Makes sense. :)

widgets/contact/fields/radio.class.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+class SiteOrigin_Widget_ContactForm_Field_Radio extends SiteOrigin_Widget_ContactForm_Field_Base {
+
+	public function render_field( $options ) {
+		if ( ! empty( $options['field']['options'] ) ): ?>
+		<ul>
+			<?php foreach ( $options['field']['options'] as $i => $option ): ?>
+			<li>
+				<label>
```
gregpriday: Missed this, there should be a closing `<label>` tag too. Also, double check the indenting for this section. The `foreach` and the `if` should indent the next lines. So make sure that the opening and closing `<ul>` and `<li>` are at the same depth.

AlexGStapleton: Funny story - I actually used the checkbox field as a base and it turns out that has a missing label. :S I've updated the checkboxes structure - how does that look?

gregpriday:
Code is looking nice and tidy. These little changes/fixes are awesome. Keep em coming.

PR #386, opened 2017-01-02
Price Table: Shortcode support for feature text

AlexGStapleton:
This will allow users to use shortcode tooltip plugins while we currently don't have said functionality.

@gregpriday I ran do_shortcode before escaping the text as I suspect that might be better overall. With that said, should it be escaped before? I worry that a plugin might not escape the text at all and will end up trusting "bad" data.


braamgenis:
Yeah, better not to trust that 3rd parties will handle escaping correctly, or at all. :+1:

PR #387, opened 2017-01-02
Testimonials: Add responsive image sizes

AlexGStapleton:
If the user decides to have a large image size it can cause rather large display issues on tablet and mobile devices. This PR aims to fix that by offering responsive settings for image sizes so they can specificly set their desired mobile and tablet image sizes.

### DIFF
--- a/widgets/testimonial/styles/default.less
+++ b/widgets/testimonial/styles/default.less
@@ -48,6 +48,18 @@
 
 	.sow-round-image-frame {
 		border-radius: @image_size;
+		width: @image_size;
+		height: @image_size;
+		
+		@media screen and (max-width: @tablet_width) {
+			width: @tablet_image_size;
+			height: @tablet_image_size;
+		}
+
+		@media screen and (max-width: @mobile_width) {
+			width: @mobile_image_size;
+			height: @mobile_image_size;
+		}
 	}
 
 	// All the specific layouts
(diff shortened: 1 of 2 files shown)


widgets/testimonial/styles/default.less
```diff
@@ -1,6 +1,5 @@
-@import "../../../base/less/mixins";
```
braamgenis: Why remove the relative path to the `mixins` file?

gregpriday: I think this is automatically handled by preprocessing now. Although the only problem is that it'll no longer autocomplete the mixins in something like PHP Storm.

gregpriday: This is where the preprocessing is happening - https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/siteorigin-widget.class.php#L867-L875

braamgenis: Yeah, I was just busy writing a larger comment about that. :smile: So we've basically hardcoded the relative paths in there, which is somewhat hacky, but doesn't cause any issues because the relative path in the `@import` is stripped. I think we should be able to, for example, use a command line LESS compiler on this file and if you do that without the relative path it will throw an error. And then, as you mentioned, the fact that IDEs like PHPStorm and IntelliJ complain.

PR #388, opened 2017-01-04
Compatibility with Elementor

braamgenis:
Some changes to Beaver Builder compat too, so that the active builder's compat code is loaded as early as possible and then each can decide for itself which hooks to use for initialization etc.

### DIFF
--- a/base/js/admin.js
+++ b/base/js/admin.js
@@ -209,6 +209,7 @@
                     value: parseInt( $input.val() ),
                     slide: function( event, ui ) {
                         $input.val( parseInt(ui.value) );
+                        $input.trigger( 'change' );
                         $$.find('.siteorigin-widget-slider-value').html( ui.value );
                     }
                 });
--- a/base/js/admin.js
+++ b/base/js/admin.js
@@ -175,7 +175,13 @@
             $el.find('.siteorigin-widget-field-repeater-item').sowSetupRepeaterItems();
 
             // Set up any color fields
-            $fields.find('> .siteorigin-widget-input-color').wpColorPicker();
+			$fields.find('> .siteorigin-widget-input-color').wpColorPicker( {
+				change: function(event, ui) {
+					setTimeout(function() {
+						$(event.target).trigger('change');
+					}, 100);
+				}
+			} );
 
             ///////////////////////////////////////
             // Handle the sections
--- a/compat/compat.php
+++ b/compat/compat.php
@@ -2,8 +2,8 @@
 
 class SiteOrigin_Widgets_Bundle_Compatibility {
 
-	const BEAVER_BUILDER = 'BEAVER_BUILDER';
-	const ELEMENTOR = 'ELEMENTOR';
+	const BEAVER_BUILDER = 'Beaver Builder';
+	const ELEMENTOR = 'Elementor';
 
 	/**
 	 * Get the singleton instance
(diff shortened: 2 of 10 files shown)


base/js/admin.js
```diff
@@ -175,7 +175,13 @@
             $el.find('.siteorigin-widget-field-repeater-item').sowSetupRepeaterItems();
 
             // Set up any color fields
-            $fields.find('> .siteorigin-widget-input-color').wpColorPicker();
+			$fields.find('> .siteorigin-widget-input-color').wpColorPicker( {
```
gregpriday: Indenting on this section seems to have changed to tab instead of space indentation.

compat/elementor/elementor.php
```diff
@@ -0,0 +1,103 @@
...
+	function __construct() {
+		add_action( 'template_redirect', [ $this, 'init' ] );
+
+		add_action( 'wp_ajax_elementor_render_widget', [ $this, 'ajax_render_widget_preview' ] );
+		add_action( 'wp_ajax_elementor_editor_get_wp_widget_form', [ $this, 'ajax_render_widget_form' ] );
+	}
+
+	function init() {
+
+		$this->plugin = Elementor\Plugin::instance();
```
gregpriday: What happens here, in the rare case that a user is using PHP 5.2, plus has Elementor and Widgets Bundle enabled. I know Elementor has some code to prevent itself from loading - does that also stop our compatibility code from loading?

braamgenis: No, but we do our own check [here](https://github.com/siteorigin/so-widgets-bundle/blob/3b240d6a98220d1b12decf3c11429900ce7ea9ed/compat/compat.php#L44) which prevents our Elementor compat code from loading.
We could probably just check for `class_exists( 'Elementor\\Plugin' )` because I don't think that class is loaded if PHP version is < 5.4, but I feel safer having the version guard too. :sweat_smile:

base/js/admin.js
braamgenis: Oh yeah, meant to have a chat about this. That was autoformatted due to the [`.editorconfig`](https://github.com/siteorigin/so-widgets-bundle/blob/82e93dbe224ad355e3483efe06fc5154c00eceee/.editorconfig) file. According to WordPress coding standards for [JavaScript](https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#spacing) we should be using tabs for indentation in JavaScript files.

compat/compat.php
```diff
@@ -15,19 +16,32 @@ public static function single() {
...
 
 	function is_active( $builder ) {
-		switch ( $builder ) {
+		switch ( $builder[ 'name' ] ) {
 			case self::BEAVER_BUILDER:
-				return class_exists( 'FLBuilderModel' ) && FLBuilderModel::is_builder_active();
+				return class_exists( 'FLBuilderModel' );
+			break;
+			case self::ELEMENTOR:
+				return version_compare( PHP_VERSION, '5.4', '>=' ) && class_exists( 'Elementor\\Plugin' );
```
KingYes: You don't need to compare PHP version. Just check if `\Elementor\Plugin` class exists. We don't include any files if the website is not in our [requirements](https://docs.elementor.com/article/38-requirements).

WPDevHQ:
Activating the develop branch causes Elementor to hang.

Tested with the latest version of Elementor.

WPDevHQ:
Traced the cause to compat.php - line 6 should be `const ELEMENTOR = 'Elementor\Plugin';`

gregpriday:
@WPDevHQ Hi Zulf! It's been a while, hope you're well. So that constant is actually an internal constant, so changing it is just stopping our Elementor compatibility code from loading entirely.

It seems like the issue is actually from `$this->plugin->db->get_builder( $post_id, Elementor\DB::REVISION_DRAFT );`, because `Elementor\DB::REVISION_DRAFT` no longer exists. @braamgenis - is there any way you can replace this line so everything works with the latest version of Elementor?

WPDevHQ:
Been a while indeed Greg. All's good on my side, how you been? Happy New Year :)

Ah OK. I just saw the omission or so I thought, changed it and Elementor was loading again. Didn't check further than that.

WPDevHQ:
Looks like switching that line to `$this->plugin->db->get_builder( $post_id, Elementor\DB::STATUS_DRAFT );` does the trick. I have access to the widgets in Elementor and content renders.

Not sure if there is or would be any side effects though.

compat/compat.php
braamgenis: @KingYes Thanks for reviewing! I removed the PHP version check. :+1:

braamgenis:
@WPDevHQ Thanks for helping us test! That fixed it. :+1:

KingYes:
Hey @braamgenis 

This is Yakir from Elementor. Today we want to release new version of Elementor, and I see your plugin is not supported anymore. You print styles & scripts for Elementor Editor by WP `wp_enqueue_scripts` action. We don't want to embed ALL scripts in the editor, just what is needed.

So please use this action for that: `elementor/editor/before_enqueue_scripts` and remove all your conditions to check if the script is in Elementor editor or not :)

Let me know if you get have problem with that.

BTW This action has been a part of Elementor for a while so you don't need to worry about compatibility.

braamgenis:
Hi @KingYes Thanks so much for letting us know! Can I use the `release/1.4.0` branch for testing? I couldn't find anything about your dev build process. Is it enough to run `npm install` and then run `grunt build`?

KingYes:
We released this version. So you can download from WordPress repo..

PR #393, opened 2017-01-10
Taxonomy Display Setting: Text

AlexGStapleton:
This PR adds a text display setting for the taxonomy widget. This is ideal for when the user doesn't actually want to link to a specific page but instead purely wants to list them.


gregpriday:
Looks like this just needs some line height. Most themes do some line height styling on paragraph tags, so maybe just wrapping the list in `<p>` tags will do the trick. Always best to get as much styling as possible from the theme.

Could do the same for the links list.

Issue #250, opened 2016-06-26
Contact: Ability to set field heigths (text input / textarea)


AlexGStapleton:
A design setting that will allow you to override theme styling for form text input or textarea height would allow users to do this directly via the widget rather than CSS.

braamgenis:
Do you maybe want to add this as part of #394? Shouldn't take much to add height there.

AlexGStapleton:
Sure.

AlexGStapleton:
Resolved in #394

PR #394, opened 2017-01-12
Contact: Add Field Design Settings

AlexGStapleton:
This PR adds the following field design settings.
- font 
- font size
- colors
- margin
- padding
- background
- Border
- Border width
- Border style
- Border Rounding

 These design settigns are applied to input, textarea and datetime fields.

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -343,6 +343,72 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 						),
 					),
 
+					'fields' => array(
+						'type'   => 'section',
+						'label'  => __( 'Fields', 'so-widgets-bundle' ),
+						'fields' => array(
+							'font' => array(
+								'type'    => 'font',
+								'label'   => __( 'Font', 'so-widgets-bundle' ),
+								'default' => 'default',
+							),
+							'font_size' =>  array(
+								'type'  => 'measurement',
+								'label' => __( 'Size', 'so-widgets-bundle' )
+							),
+							'color' =>  array(
+								'type'  => 'color',
+								'label' => __( 'Color', 'so-widgets-bundle' ),
+							),
+							'margin' => array(
+								'type'  => 'measurement',
+								'label' => __( 'Margin', 'so-widgets-bundle' )
+							),
+							'padding' => array(
+								'type'  => 'measurement',
+								'label' => __( 'Padding', 'so-widgets-bundle' )
+							),
+							'background' =>  array(
+								'type'  => 'color',
+								'label' => __( 'Background', 'so-widgets-bundle' ),
+							),
+							'border_color' => array(
+								'type'    => 'color',
+								'label'   => __( 'Border color', 'so-widgets-bundle' ),
+								'default' => '#c0c0c0',
+							),
+							'border_width' => array(
+								'type'    => 'measurement',
+								'label'   => __( 'Border width', 'so-widgets-bundle' ),
+								'default' => '1px',
+							),
+							'border_style' => array(
+								'type'    => 'select',
+								'label'   => __(' Border style', 'so-widgets-bundle' ),
+								'default' => 'solid',
+								'options' => array(
+									'none'   => __( 'None', 'so-widgets-bundle' ),
+									'hidden' => __( 'Hidden', 'so-widgets-bundle' ),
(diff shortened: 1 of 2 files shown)


widgets/contact/contact.php
```diff
@@ -343,6 +343,72 @@ function get_widget_form(){
 						),
 					),
 
+					'fields' => array(
+						'type'   => 'section',
+						'label'  => __( 'Fields', 'so-widgets-bundle' ),
+						'fields' => array(
+							'font' => array(
+								'type'    => 'font',
```
braamgenis: This needs to be added to the array returned by `get_google_font_fields()` to get the font imported properly. I.e.
```
	function get_google_font_fields( $instance ) {
		return array(
			$instance['design']['labels']['font'],
			$instance['design']['fields']['font'],
		);
	}
```
 At some point we still want to try simplify the process of adding and using a font field, so that this won't be necessary.

widgets/contact/styles/default.less
```diff
@@ -1,4 +1,4 @@
-@import "../../../base/less/mixins";
```
AlexGStapleton: Ah, wasn't sure if you wanted to leave that as is or not - I'll make sure to revert it in the follow up commit.

AlexGStapleton:
Note to self: Add #250 in follow up commit.

widgets/contact/styles/default.less
```diff
@@ -126,7 +156,7 @@
 		background: @error_background;
 		border: 1px solid @error_border;
 		color: @error_text;
-		border-radius: 2px;
+		.rounded(2px);
```
AlexGStapleton: There's very little reason to do this, but I was doing it for consistency reasons. It's fine though.

widgets/contact/styles/default.less
```diff
@@ -56,12 +56,42 @@
...
-		.box-sizing(border-box);
 		display: block;
 		width: 100%;
-		border-radius: 0;
+	}
+	
+	textarea,
+	input.sow-text-field,
+	.so-premium-datepicker,
+	.so-premium-timepicker {
```
AlexGStapleton: I've opted to just change the selector to textarea, input. Yes, it'll technically effect checkboxes but the styling able to be added to them is _very_ limited so it won't be noticeable as none of the applied CSS will be visible. Is that okay with you?

AlexGStapleton:
There we go - how does that look? Please note that I also did #250 in the latest commit.

widgets/contact/contact.php
AlexGStapleton: To be honest, I forgot to do this. :(
I do like the idea of automating this however as it feels a tad redundant to me.

widgets/contact/styles/default.less
braamgenis: I think we should be able to apply the `.sow-text-field` class to datepicker and timepicker fields too, without any problems. I prefer to avoid unintentionally applying styles to checkboxes.

widgets/contact/styles/default.less
braamgenis: Cool, yeah. I'm all for keeping things consistent. :+1:

widgets/contact/contact.php
braamgenis: Yeah, I'll maybe take a crack at simplifying this soon. I think there might still be an issue about this somewhere. :sweat_smile:

widgets/contact/styles/default.less
```diff
@@ -56,12 +56,42 @@
...
+
+	@field_background: default;
+
+	@field_margin: 0;
+	@field_padding: 0;
+	@field_border_radius: 0;
+	@field_border_color: default;
+	@field_border_width: default;
+	@field_border_style: default;
+	@field_height: 0;
```
braamgenis: These should have `default` as the default value. Using `0` could unintentionally overwrite existing users' styles.

braamgenis:
Cool, all looking good. This PR will make the contact form fields much more customizable. I'm going to apply the changes I mentioned in the last few comments to get this merged quickly.

widgets/contact/styles/default.less
AlexGStapleton: That... that's a good point. Updated.

AlexGStapleton:
Oddly, I can't seem to see the comment regarding `.sow-text-field` so I didn't know you wanted to do that before making my full up commit. 😕

PR #401, opened 2017-01-30
Simple Masonry: Add load event for resizeMasonry

AlexGStapleton:
If the user has an awful connection or the designers are using awfully large images, the timer is too slow - gotta rely on the .load to ensure the masonry effect. Hopefully, it doesn't look too bad during the wait (ie. the time between 100 and load). :/

### DIFF
--- a/widgets/simple-masonry/js/simple-masonry.js
+++ b/widgets/simple-masonry/js/simple-masonry.js
@@ -64,9 +64,9 @@ sowb.setupSimpleMasonry = function() {
         });
     };
 	
-	$(window).on('resize panelsStretchRows', resizeMasonry);
+	$(window).on('resize panelsStretchRows load', resizeMasonry);
 	
-	// Ensure that the masonry has resized correct on load.
+	// Ensure that the masonry has resized correctly on load.
 	setTimeout( function () {
 		resizeMasonry();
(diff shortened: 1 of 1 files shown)


braamgenis:
We're already using [`imagesLoaded`](http://packery.metafizzy.co/layout.html#imagesloaded) to make sure the images are loaded before the layout happens [here](https://github.com/AlexGStapleton/so-widgets-bundle/blob/a2912d5930a1db6a1aa9cd07ff440b2b9abf7b2b/widgets/simple-masonry/js/simple-masonry.js#L26). Is this not working?

AlexGStapleton:
I wasn't aware of that, but no, it's not happening reliably.

braamgenis:
:suspect: Hrm... OK. Perhaps we should figure out why that is and make sure that `imagesLoaded` isn't working as it should. If it's the culprit, then it should be removed. I can play around with large images and do some bandwidth throttling in DevTools to see if I can reproduce.

AlexGStapleton:
Yeah, that, overall is the better solution tbh. Gonna close this as this is a bandaid fix for a much larger issue.

braamgenis:
Ok, cool. I'll create a separate issue.

PR #391, opened 2017-01-10
Post Carousel: Add default thumbnail

AlexGStapleton:
When a post doesn't have a featured image set, it shows nothing. This PR will allow users to specify a default image to provide more of a site-specific touch to their carousel.

PR #412, opened 2017-02-15
Post carousel default image

braamgenis:
Recreated PR #391 
@gregpriday Mind just checking my usage of instance storage in this commit 02a13014deb9f4cbc436b1b65b18f42f67f51aac


widgets/post-carousel/post-carousel.php
```diff
@@ -16,17 +16,23 @@ function sow_carousel_register_image_sizes(){
...
-		array(
-			'post_status' => 'publish',
-			'posts_per_page' => 10,
-			'paged' => empty( $_GET['paged'] ) ? 1 : $_GET['paged']
-		)
-	);
-
-	$posts = new WP_Query($query);
+
+	$instance_hash = $_GET['instance_hash'];
```
gregpriday: Always slightly safer to check if `$_GET['instance_hash']` isn't `empty` before accessing it to make sure we don't trigger any warnings. Can assign to `false` if it is empty and make the check on line 23 a standard boolean check.

widgets/post-carousel/post-carousel.php
```diff
@@ -25,8 +25,17 @@ function sow_carousel_get_next_posts_page() {
 		)
 	);
 
+	$instance_hash = $_GET['instance_hash'];
+	$template_vars = array();
+	if ( ! empty( $instance_hash ) ) {
+		$widget = new SiteOrigin_Widget_PostCarousel_Widget();
```
braamgenis: Yeah, I think you mentioned that previously and I fixed it up [here](https://github.com/siteorigin/so-widgets-bundle/pull/412/commits/75c8c7b1f26624991c2f1fa477290a8f3d538a9e) :smile:

Issue #81, opened 2016-01-18
Features Widget: Option to left/right align icon.


gregpriday:
At the moment, the icon in the features widget always appears above the text. Having an option to place the icon to the left, right or below the text would be very useful.

Misplon:
A long running request ever since we launched the Features widget. Mainly because the Circle Icon widget has this ability. I'm sure users will be stoked to see this one go live. Thanks guys :)

PR #413, opened 2017-02-16
Allow specifying position of features widgets icons.

braamgenis:
This implements #81


widgets/features/styles/default.less
```diff
@@ -17,25 +17,162 @@
...
-           color: @text_color;
-       }
-
-       > p.sow-more-text {
-           .font(@more_text_font, @more_text_font_weight);
-           font-size: @more_text_size;
-           color: @more_text_color;
-       }
-   }
+@container_size: default;
```
gregpriday: Working from memory here, so I could be wrong, but the main intention with default was that if an attribute has the value `default`, then it's removed from the CSS. You can see that in action here - https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/siteorigin-widget.class.php#L853

We're probably pretty safe with `@container_size`, because it's a measurement field, but technically if this is ever empty and the value of `@container_size` stays as `default`, then we might have problems with things like `padding-right: @container_size + 10px;`, which would be the same as `padding-right: default + 10px;`.

Because default doesn't really serve much of a purpose here, we could enter a safer default, or just use the main widget's default.

braamgenis: Ah, yeah, cool. Will just make it the widget default of `84px`.

braamgenis:
Oh, I didn't see the comment about using flexbox anywhere, but yeah, I'll look into it. I have a slight concern that it might cause layout changes for existing users, but maybe not, then it will be ideal.

braamgenis:
@gregpriday Updated to use flexbox. Seems to have made it a bit simpler and shouldn't impact existing layouts.

gregpriday:
@braamgenis Flexbox is the future. We should actually look at using it for Page Builder too.

PR #416, opened 2017-02-17
Option to equalize row heights in price table widget.

### LINKED ISSUE #226: Price Table: Add functionality for consistent feature height across columns
[Suggested in this thread](https://siteorigin.com/thread/dynamic-and-consistent-row-height-in-price-table/)

A really nice feature would be an option to have consistent feature row height across columns. Basically, what would happen is we would write a script to compare a feature row and the row with the highest height would be set as the height of all columns for that row. This will provide a more consistent look to the pricing widget.

braamgenis:
Resolves #226


widgets/price-table/js/pricetable.js
```diff
@@ -1,4 +1,28 @@
 jQuery( function($){
+    $('.ow-pt-columns-atom.sow-equalize-row-heights').each(function () {
```
gregpriday: So right now, we only have one visual style for the price table (atom), but in the future, we'll probably be adding more. Is there any way to make this style independent? Maybe target the generic widget wrapper, which should be one level up from `.sow-equalize-row-heights`.

widgets/price-table/price-table.php
```diff
@@ -205,6 +210,7 @@ function get_template_variables( $instance, $args ) {
 			'before_title' => $args['before_title'],
 			'after_title' => $args['after_title'],
 			'button_new_window' => $instance['button_new_window'],
+            'equalize_row_heights' => ! empty( $instance['equalize_row_heights'] ),
```
gregpriday: This line is using space, instead of tab spacing.

widgets/price-table/tpl/atom.php
```diff
@@ -24,15 +25,15 @@
 				<div class="ow-pt-per"><?php echo esc_html($column['per']) ?></div>
 			</div>
 
-			<?php if(!empty($column['image'])) : ?>
-				<div class="ow-pt-image">
-					<?php $this->column_image($column) ?>
-				</div>
-			<?php endif; ?>
+            <div class="ow-pt-image">
```
gregpriday: What's the main reason for switching this around? Wouldn't this result in empty `ow-pt-image` divs if there's no row image set? That feels a little messy.

There are also a few lines below this one that use space instead of tab indenting.

widgets/price-table/js/pricetable.js
braamgenis: Hmm, yeah that's a good point. However, the current row height calculation is quite specific to the `atom` template layout and its elements. Say we added another template that used a horizontal layout instead of vertical, I think the calculation would be, at least slightly, different. So IMHO it's not worth trying to generalize this before we have another template that would use this.

widgets/price-table/price-table.php
braamgenis: :sob:

widgets/price-table/tpl/atom.php
braamgenis: Main reason is that if the image row is present in one feature and we're doing the row height equalization, there needs to be an element on which to set the height that matches the height of the image. (I hope that made sense :sweat_smile: ) So perhaps I'll change this to output if at least one column has an image.

braamgenis:
@gregpriday Let me know if you're happy with my changes for the image `div` output and thoughts on leaving the more generalized layout for when we decide to add another template.

widgets/price-table/js/pricetable.js
gregpriday: In this case, if we targetted `sow-equalize-row-heights` (along with something else to make sure we only target the price table), we could make it up to the template to decide if it wants equalised heights. Would something like that work? It seems better than adding each individual class to this selector.

braamgenis:
@gregpriday I changed the row height equalization to use the widget class selector. Now any other templates can just output the `sow-equalize-row-heights` class.

gregpriday:
@braamgenis perfect

Issue #233, opened 2016-06-10
Image: Add Custom Size Option in size select


AlexGStapleton:
The image widget sizing is really quite restrictive as it is right now so it would be a good idea to add a custom size option in the size select. It should work, and act in a similar method to the custom size option in the image details modal when you edit an image in the wp editor.

Reference Images:
![](https://i.imgur.com/3G1ZBaI.png)
![](https://i.imgur.com/9nNDp87.png)

gregpriday:
Good call. I haven't checked yet, but I assume that if this is in the WordPress editor, there must be extra functionality in the WordPress core to handle generating arbitrary image sizes without registering them using `add_image_size`.

braamgenis:
@AlexGStapleton @gregpriday This can be done using the built in WP Image Editor functions, but I wonder if we're not better off just providing a description with a link to the WP image editor? Then users can choose whether they want to scale the image or just crop it to a certain size etc. Just trying to avoid reinventing the wheel. What do you guys think?

AlexGStapleton:
I'm kind of indifferent, although I'm leaning towards the not in favor side.

braamgenis:
@AlexGStapleton Not in favor of... linking to the existing WP Image Editor interface?

AlexGStapleton:
Whoops. Yeah, sorry - I should have been more specific. I thinking pointing someone in the direction of a tool rather than giving them access to it directly feel a tad... odd. It seems odd to me that we would want to direct them away from what they're currently doing (at least interface wise) and alter the image on a completely separate page.

braamgenis:
I agree, it's not ideal to interrupt the flow, but this way the user has a lot more options. E.g. for their custom sized image they can choose whether they want to resize by cropping or scaling. Perhaps we can do something like opening the WP Image Editor in a dialog to make the UX a bit smoother. That might have it's own problems, but will see if something like that is possible.

Misplon:
When selecting an image the **Edit Image** link is available. However, when using the Edit Image link in the SiteOrigin Editor widget, the user stays on the same page. Unfortunately, when using the Edit Image link in the SiteOrigin Image widget, the user is taken to a new page. If we could mimic the behaviour of the Editor widget/core process, that would be ideal.

<img width="1971" alt="edit-image" src="https://user-images.githubusercontent.com/789159/39858195-2105ce6e-5436-11e8-8df2-b1bdcf8c76cd.png">

[...]

PR #422, opened 2017-02-21
Allow different sizes for post carousel images

### LINKED ISSUE #269: Carousel: Add the ability to select a custom image size
A custom request is more control over the carousel image dimensions - currently locked to 272x182 via an image size. We should allow users to select another image size, similar to that of the image widget.

This change will require that the frontend CSS for this widget not be so locked to 242px width.

braamgenis:
Resolves #269 

@AlexGStapleton Please check whether this is what you were after and give any feedback. Thanks!


AlexGStapleton:
I never received notifciation of this. 😕 

I just tried it mate, looks good! :)
Although, it does really make me wish #233 was a thing.

PR #423, opened 2017-02-22
Allow multiple post types in posts selector field.

### LINKED ISSUE #271: Build Posts Query: Add the ability to change the post type field to a multiple select field
For certain widgets, the user may need to select multiple post types. Right now the post query builder isn't directly set up for this and as a result, the user is out of luck.

braamgenis:
Resolve #271.

This just adds the `multiple` attribute to the `select` element. We should consider using the generic form fields for the posts selector.

### DIFF
--- a/base/css/admin.css
+++ b/base/css/admin.css
@@ -418,6 +418,10 @@
 .siteorigin-widgets-query-builder .query-builder-content .query-builder-form-field input[type="number"] {
   width: 150px;
 }
+.siteorigin-widgets-query-builder .query-builder-content .query-builder-form-field select[multiple] {
+  min-width: 150px;
+  max-height: 83px;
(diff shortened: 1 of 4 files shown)


gregpriday:
Everything looking good to me. The only thing I'd suggest would be using checkboxes instead of a multiple select. I don't even know why those things exist, they're pretty bad from a usability perspective.

And yeah, I'd be so much happier moving the whole post selector over to our standard form fields. The post selector is a touch messy. We can save changing from a multi selec to checkboxes for the post types when we make this change.

braamgenis:
Yeah, I thought about using checkboxes, but if there are a lot of post types, the multiple select field is a bit friendlier in that it scrolls etc.

PR #443, opened 2017-03-23
Add support for Page Builder database render

gregpriday:
This is to support a new mode Page Builder has where it creates a special version of the content to store in post_content.

### DIFF
--- a/base/inc/shortcode.php
+++ b/base/inc/shortcode.php
@@ -18,13 +18,10 @@ function siteorigin_widget_shortcode( $attr, $content = '' ){
 	if( ! empty( $attr[ 'class' ] ) && isset( $wp_widget_factory->widgets[ $attr[ 'class' ] ] ) ) {
 		$the_widget = $wp_widget_factory->widgets[ $attr[ 'class' ] ];
 
-		list( $post_id, $widget_id ) = explode( ':', $attr['id'] );
-		if( empty( $post_id ) ) return '';
+		$meta = json_decode( htmlspecialchars_decode( $content ), true );
 
-		$meta = get_post_meta( $post_id, 'siteorigin_widget_instances', true );
-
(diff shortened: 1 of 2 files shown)


base/inc/shortcode.php
```diff
@@ -0,0 +1,97 @@
...
+	$attr = shortcode_atts( array(
+		'class' => false,
+		'id' => '',
+	), $attr, 'siteorigin_widget' );
+
+	global $wp_widget_factory;
+	if( ! empty( $attr[ 'class' ] ) && isset( $wp_widget_factory->widgets[ $attr[ 'class' ] ] ) ) {
+		$the_widget = $wp_widget_factory->widgets[ $attr[ 'class' ] ];
+
+		$meta = json_decode( htmlspecialchars_decode( $content ), true );
```
braamgenis: Cool, the only reason I can think of not to use `$content` is if we expect to ever want to actually use it for content. :stuck_out_tongue:

base/inc/shortcode.php
```diff
@@ -0,0 +1,97 @@
...
+		$meta[$i] = array(
+			'args' => $args,
+			'instance' => $instance,
+		);
+		update_post_meta( get_the_ID(), 'siteorigin_widget_instances', $meta );
+
+		$html .= $before_widget;
+		$html .= '[siteorigin_widget ';
+		$html .= 'class="' . get_class( $the_widget ) . '"]';
+		$html .= htmlspecialchars( wp_json_encode( $meta[ $i ] ) );
```
braamgenis: Just a note that `htmlspecialchars()` only encodes the characters `& " ' < >`. I can't think of an example, but it feels like there's potential for breakage if the content contains characters like square brackets `[ ]`. It might be safer to use `htmlentities()` which will encode all characters with HTML entity equivalents.

PR #451, opened 2017-04-08
Compatibility with visual composer

braamgenis:
Here it is at last. :smile:


base/js/admin.js
```diff
@@ -808,65 +745,138 @@
...
+			$r.data('item-id', itemId);
+
+			return itemId;
+		}
+		else {
+			return false;
+		}
+	};
+
+	sowbForms.getWidgetFieldVariable = function ( widgetClass, elementName, key ) {
```
gregpriday: Do you want to add comments for all these functions? I can't actually remember what `getWidgetFieldVariable` and some of these other functions do.

compat/visual-composer/siteorigin_widget_vc_template.php
```diff
@@ -0,0 +1,5 @@
+<?php
+$widget_settings = $this->get_widget_settings( $atts );
+$instance = $this->update_widget( $widget_settings['widget_class'], $widget_settings['widget_data'] );
```
gregpriday: Does this mean we're going to be running the widget through the update function on every page load?

compat/visual-composer/siteorigin_widget_vc_template.php
```diff
@@ -0,0 +1,5 @@
...
+$this->render_widget( $widget_settings['widget_class'], $widget_settings['widget_data'] );
+?>
```
gregpriday: Do we need this closing PHP tag in here? Generally, it's best to remove this to prevent any accidental white space after the file.

compat/visual-composer/sowb-vc-widget.js
```diff
@@ -0,0 +1,86 @@
...
+			data,
+			function(result) {
+				$formContainer.html(result);
+				// To ensure data is updated.
+				$formContainer.trigger('change');
+			},
+			'html'
+		);
+	});
+    vc.atts.sowb_json_escaped = {
```
gregpriday: Space indenting used here.

compat/visual-composer/sowb-vc-widget.js
```diff
@@ -0,0 +1,86 @@
...
+        }
+    };
+
+	vc.events.on("shortcodeView:updated:siteorigin_widget_vc", function() {
+
+		if( typeof vc.frame_window !== 'undefined' && typeof vc.frame_window.sowb !== 'undefined') {
+			var sowb = vc.frame_window.sowb;
+
+			// Trigger Widgets Bundle widgets to setup
+			// This isn't working for some reason, so keep calling the functions directly for now.
```
gregpriday: Any ideas on why this might not be working? We're more than likely going to run into issues if we keep relying on a widget by widget setup. Maybe it just needs a short delay before triggering `setup_widgets`?

compat/visual-composer/visual-composer.php
```diff
@@ -0,0 +1,282 @@
...
+			/* @var $widget SiteOrigin_Widget */
+			$widget->form( array() );
+		}
+
+		wp_die();
+	}
+
+	function update_widget_data( $content ) {
+
+		$content = preg_replace_callback( '/\[siteorigin_widget_vc [^\]]*\]/', array(
```
gregpriday: This is totally a style preference, but I'd prefer to see this as one argument per line, where `array( $this, 'update_shortcode'  )` is a single argument. So:

```
$content = preg_replace_callback(
    '/\[siteorigin_widget_vc [^\]]*\]/', 
    array( $this, 'update_shortcode'),
    $content
);
```

To me, that's a lot more clear.

compat/visual-composer/visual-composer.php
```diff
@@ -0,0 +1,282 @@
...
+		if ( ! empty( $widget ) && is_object( $widget ) && is_subclass_of( $widget, 'SiteOrigin_Widget' ) ) {
+			/* @var $widget SiteOrigin_Widget */
+			$widget_atts['widget_data'] = $widget->update( $widget_atts['widget_data'], $widget_atts['widget_data'] );
+		}
+
+		$widget_json = json_encode( $widget_atts );
+
+		$widget_json = htmlentities( htmlentities( $widget_json ) );
+
+		$widget_json = str_replace( array(
```
gregpriday: Similar style comment to earlier. This would be much clearer with arguments on their own lines. Unless there's some sort of style guideline that says we should be doing it the current way.

```
$widget_json = str_replace(
    array( '[', ']', ),
    array( '&#91;', '&#93;',
    $widget_json
);
```

base/js/admin.js
```diff
@@ -407,72 +409,7 @@
 
             // TODO: This very closely resembles the data extraction code in Page Builder. Try find a way to avoid having
```
braamgenis: Yeah, this comment is pretty old. Will just remove it.

gregpriday:
Found a warning.

1. Create a new visual composer page.
2 Add a SiteOrigin Widget, select a button widget from the dropdown.
3. Close the widget without editing anything, then save the page

You'll get this warning:

`Notice: Undefined index: so_widget_data in /Users/gpriday/Dropbox/Sites/dev/wp-content/plugins/so-widgets-bundle/compat/visual-composer/visual-composer.php on line 216`

AlexGStapleton:
Suggestions: Is it possible to give it a custom icon? The VC icon is okay and all but it would be better if the Widgets Bundle icon was used.

AlexGStapleton:
Suggestions: Would it be possible to add a warning when trying to switch widget? Stating something like "switching widget will reset your widget settings. Are you sure you wish to change widget?". Think of it like the revert to editor warning

AlexGStapleton:
Bug: The SiteOrigin Editor widget doesn't seem to clear when changing widget. I suspect this may be related to how wp_editor() works but if you can prevent this, consider clearing it.

AlexGStapleton:
Bug (Frontend): If FitText is used, in a headline widget for example, it'll only kick in if the user reloads the page or FitText was already present on load. In other words, it's not loading on the fly.

AlexGStapleton:
Bug (Frontend): If you use the SiteOrigin video player and setup a local video to Autoplay, the video will play (based on the audio), but the picture won't change. I suspect the cover is still visible for some reason. This may not be on us.

AlexGStapleton:
Bug (Frontend): If you add a post carousel, save and then make any changes to the query, the navigation (the arrows) won't work - they're treated like typical links. I suspect the post carousel JavaScript is only being loaded when it's initially being added and not when the interface reloads when other widgets are saved.

compat/visual-composer/siteorigin_widget_vc_template.php
braamgenis: Yeah, but this isn't necessary. Just something I forgot to remove while developing and testing. :sweat:

compat/visual-composer/siteorigin_widget_vc_template.php
braamgenis: :+1:

compat/visual-composer/sowb-vc-widget.js
braamgenis: :sob:

compat/visual-composer/sowb-vc-widget.js
braamgenis: Yeah, don't really know. I'd guess it has something to do with the preview being in an iframe window. Will look into it further.

compat/visual-composer/visual-composer.php
braamgenis: :+1:

compat/visual-composer/visual-composer.php
braamgenis: :+1:

compat/visual-composer/sowb-vc-widget.js
braamgenis: Yeah, got it fixed. Just needed to use the iframe window's jQuery object instead of this window.

gregpriday:
@AlexGStapleton can you check if all these issues have been fixed by the last round of commits? I'm going to do a code review so long.

AlexGStapleton:
Some problems:

- The editor didn't seem to clear for me when I switched widget

- The icon wasn't appearing for me. The former is likely on me though.
- The widget change confirmation dialog only happens for me if I've made any changes to the widget this save. If I reload the page and then try changing the widget it'll change without issue.

braamgenis:
@AlexGStapleton 

- The editor is clearing on my side. Could you give some more specific steps to reproduce what you're seeing?
- Where is the icon not appearing for you? You likely just need to build the CSS again. (Run `gulp build:dev`)
- Yeah, that's intentional. Don't want to annoy the user with too many alerts. If they change to a different widget, their settings on the old widget will only be lost if they click "Save changes" in which case, that's what they wanted. Otherwise to revert to their old settings they can just close and reopen the dialog.

AlexGStapleton:
@braamgenis 

- I didn't really do anything special, I edited an existing page (although it wasn't VC enabled) and then adding a SiteOrigin... widget (I can't remember what they call them) and I tried to replicate this issue and was able to without much issue.
- I just re-ran gulp build:dev and no dice. This is likely on me so I wouldn't worry too much about it (I'm sure I'm doing something wrong in this instance).
- Ah. Yeah, that's fine in that case imo.

gregpriday:
@AlexGStapleton - editor seems to be clearing my side too. Going to merge this so we can start working towards the 1.9 release.

PR #461, opened 2017-04-25
Posts selector refactor

braamgenis:
@gregpriday Please take a look when you can.

Still need to do a bit of cleanup to remove the no longer used code for the old post selector.


gregpriday:
Epic, all looks good to me. One shotted! Going to do some minor indenting cleanup after the merge.

PR #465, opened 2017-05-02
New PHP LESS compiler

braamgenis:
Still going to update our old usage of guards to the new better way.


base/inc/lessc.inc.php
```diff
@@ -1,3675 +1,276 @@
...
- * Licensed under MIT or GPLv3, see LICENSE
+ * This file provides the part of lessphp API (https://github.com/leafo/lessphp)
+ * to be a drop-in replacement for following products:
+ *  - Drupal 7, by the less module v3.0+ (https://drupal.org/project/less)
+ *  - Symfony 2
  */
 
+// Register autoloader for non-composer installations
+if ( !class_exists( 'Less_Parser' ) ) {
+	require_once __DIR__ . '/lib/Less/Autoloader.php';
```
gregpriday: I think `__DIR__ ` isn't PHP 5.2 compatible, which we're still supporting for the time being.

gregpriday:
Everything looks good. I just need to double check PHP 5.2 compat. Is everything in the new LESS library compatible with this?

base/inc/lessc.inc.php
braamgenis: Ok, cool. Is that the only change for 5.2 compat?

gregpriday:
Did some basic tests with PHP 5.2. Everything seems to be working.

PR #470, opened 2017-05-05
Google Maps Fallback Image

braamgenis:
Sends the fallback image markup as a data attribute to the front end. 

For interactive maps using the JS API the front end detects an API error by hooking into the `window.console.error` function and checking the error string for the message format specified [here](https://developers.google.com/maps/documentation/javascript/error-messages#checking-errors) and displays the fallback image if one occurs.

For static maps using the Static Map API the front end sets a flag on the element if the `onerror` handler has been called, or adds a listener for the `error` event if loading isn't complete yet.


widgets/google-map/js/js-map.js
```diff
@@ -404,6 +404,30 @@ jQuery(function ($) {
 				apiUrl += '&key=' + apiKey;
 			}
 
+			// This allows us to "catch" Google Maps JavaScript API errors and do a bit of custom handling. In this case,
+			// we display a user-specified fallback image if there is one.
+			if ( window.console && window.console.error ) {
+				var errLog = window.console.error;
+
+				sowb.onLoadMapsApiError = function ( error ) {
+					var matchError = error.match( /^Google Maps API (error|warning): ([^\s]*)\s([^\s]*)(?:\s(.*))?/ );
```
gregpriday: Will this also use the fallback image in the case of warnings? What does the maps API class as a warning?

braamgenis: Yeah. The warnings we're likely to see are to do with API keys being invalid, in which case the map might not display, so I think we should rather display the fallback in these cases.

braamgenis: You can see a list of the errors and warnings (about halfway down the list) [here](https://developers.google.com/maps/documentation/javascript/error-messages#deverrorcodes)

gregpriday: Aaah, yeah, warnings mostly look like "breakers".

PR #473, opened 2017-05-08
Posts Selector Relative Dates

### LINKED ISSUE #466: Post Selector: Ability to specify current day, month, year.
A semi-common request for the additional field is the ability to set relevant current dates, for example, day, month, year, etc. An example use case would be if I wanted to include all posts I posted last week in a post loop I would currently have to manually update the date range every week.

[Slack reference](https://siteorigin.slack.com/archives/C02SYBBH2/p1493732957835081)

braamgenis:
To resolve #466. Allow selection of a date range relative to today.

### DIFF
--- a/base/inc/fields/css/date-range-field.less
+++ b/base/inc/fields/css/date-range-field.less
@@ -6,10 +6,24 @@
 		span {
 			display: inline-block;
 			width: 35px;
+
+			&:last-child {
+				margin-left: 10px;
+			}
 		}
 
 		input[type="text"].datepicker {
 			width: 258px;
+		}
+	}
+
+	.sowb-relative-date {
+		> input {
+			max-width: 100px;
+		}
+
+		> select {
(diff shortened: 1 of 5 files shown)


AlexGStapleton:
0 as the to date seems to be ignored. For example, if I set the dates from 1 days ago to 0 days ago, nothing will be returned. 

This may be related but, it also seems like if it's a single day different, nothing will be returned. For example, from 3 to 4 days.

braamgenis:
@AlexGStapleton The default for date queries excludes the selected dates. I.e. if you select from 1 May to 3 May it will only return the posts for 2 May. I've set it to always include the selected dates as that feels more intuitive to me. Please take a look and let me know what you think.

AlexGStapleton:
Much better! 👍

gregpriday:
Everything looks good on the code side, so going to merge.

PR #479, opened 2017-05-17
Widgets browser storage backups WIP

### LINKED ISSUE #464: Editor Widget: Add some sort of draft saving in the background
If possible, look at adding some sort of background draft saving for Editor widget editing only. Something similar to Gmail. We had a case where a user didn't update the widget while editing and lost work.

braamgenis:
To resolve #464. Just to get some initial feedback. Still needs a bit tweaking to get all fields restoring properly.


base/siteorigin-widget.class.php
```diff
@@ -568,7 +569,15 @@ function enqueue_scripts( $form_type = false ) {
 
 			wp_localize_script( 'siteorigin-widget-admin', 'soWidgets', array(
 				'ajaxurl' => wp_nonce_url( admin_url('admin-ajax.php'), 'widgets_action', '_widgets_nonce' ),
-				'sure' => __('Are you sure?', 'so-widgets-bundle')
+				'sure' => __('Are you sure?', 'so-widgets-bundle'),
+				'backup' => array(
+					'newerVersion' => __(
+						'There is a newer version of this widget\'s content available.',
```
gregpriday: For translation strings, it's better to avoid slashes inside the string. So in this case, we'd have `"There is a newer version of this widget's content available."`.

Also, I'm not sure if it's important, but I'd always prefer having translation functions all on one line just to make sure it's compatible with anything that parses for translatable strings.

braamgenis: Ah, yeah, good point about the slashes.

Are you saying there are apps that search PHP files and replace `__()`? :confused:

AlexGStapleton: I suspect certain software may not be able to extract the string if it's broken up over multiple lines like that.

AlexGStapleton:
Is the notice meant to fade after the data has been restored? If so, it doesn't seem to be doing that for me. It does disappear when dismissed though.

We should add a disclaimer stating that the data will be removed. I know that it will and you know that it will but I'm sure someone will click it thinking something else. Better to be safe. :/ For reference, here's the WordPress disclaimer:


Also, I find it odd that the notification buttons are so lighter compared to the regular buttons (such as the preview button in the editor widget). Is that intentional? For comparison, here's the difference:

braamgenis:
@AlexGStapleton The notice will now disappear when the backup is restored, but reappear if the page isn't updated. I've also added a warning about the contents being replaced. I'm not sure what you mean about the buttons being lighter though? They should have the normal WP core styling applied.

AlexGStapleton:
Compare the buttons in my screenshot. The one on the left has been modified to use the same styling as the preview button at the bottom of the editor widget. The one on the right is the default button.

braamgenis:
Hmmm... They look pretty close to the preview button to me. I've also modified the button on the left to use the preview button styles. The only way I get the preview button to look like yours is on hover.
screenshot_20170519_093354
screenshot_20170519_093612

AlexGStapleton:
Hm. Maybe I accidentally copied the hover styling. 😕

base/inc/fields/js/media-field.js
```diff
@@ -1,8 +1,9 @@
 ( function( $ ) {
 
     $(document).on( 'sowsetupformfield', '.siteorigin-widget-field-type-media', function(e) {
-        var $media = $(this).find('> .media-field-wrapper');
-        var $field = $media.closest('.siteorigin-widget-field');
+		var $field = $( this );
+        var $media = $field.find('> .media-field-wrapper');
```
gregpriday: Indenting is usually the only thing I can find ;)

gregpriday: Mmm, looks like there's a mix of space and tab indents in a few of these field JS files. So maybe we should just leave this for now, and reformat all of them in a separate commit after the merge.

base/js/admin.js
```diff
@@ -542,7 +586,7 @@ var sowbForms = window.sowbForms || {};
 			});
 			var repeaterHtml = repeaterObject.html().replace(/_id_/g, $nextIndex);
 
-			var readonly = typeof $el.attr('readonly') != 'undefined';
+			var readonly = typeof $el.attr('readonly') !== 'undefined';
```
gregpriday: I wonder if we still need to use `typeof` anymore. Maybe `_.isEmpty` or just `!== undefined`. I think `typeof` was a hack for older browsers.

AlexGStapleton: I'm about 99% isEmpty doesn't check if it's undefined, only that it doesn't equal to 0. [isUndefined](http://underscorejs.org/#isUndefined) could very well be used though.

gregpriday: @AlexGStapleton aah, yeah `isUndefined` is probably what we should be using then. We can save that for a refactor later on.

PR #480, opened 2017-05-17
Google Map Widget optimization - Custom Marker Icon for each marker

KevinTaron:
Now you can define a custom icon per map marker.


gregpriday:
This looks great @KevinTaron - thanks for the contribution! We'll just get @braamgenis to review the PR because it's his widget.

PR #483, opened 2017-05-26
Some reformatting to fields JS

gregpriday:


gregpriday:
Was sure I merged develop. Going to try this again.

PR #527, opened 2017-07-31
Codacy fixes

braamgenis:
Not complete, but creating the PR to trigger re-analysis by codacy.

### DIFF
--- a/base/inc/actions.php
+++ b/base/inc/actions.php
@@ -118,7 +118,7 @@ function siteorigin_widget_action_search_terms() {
 		WHERE
 			terms.name LIKE '%s'
 		LIMIT 20
-	", '%'.$term.'%');
+	", '%' . esc_sql( $term ) . '%');
 
 	$results = array();
 
--- a/base/inc/actions.php
+++ b/base/inc/actions.php
@@ -4,14 +4,21 @@
  * Action for displaying the widget preview.
  */
 function siteorigin_widget_preview_widget_action() {
-	if( empty( $_POST['class'] ) ) exit();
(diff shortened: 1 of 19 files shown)


widgets/contact/contact.php
```diff
@@ -913,8 +913,8 @@ function render_form_label( $field_id, $label, $position ) {
 	 * Ajax action handler to send the form
 	 */
 	function contact_form_action( $instance, $storage_hash ) {
-		if ( ! wp_verify_nonce( $_POST['_wpnonce'] ) ) {
-			exit();
+		if ( ! wp_verify_nonce( $_POST['_wpnonce'], '_contact_form_submit' ) ) {
+			return false;
```
AlexGStapleton: Is there a reason why you `return false` here rather than using `wp_die()` like everywhere else?

braamgenis: Yeah, this is where the issue occurred with Elementor. The `contact_form_action` function is called from the widget template file, which is sometimes rendered as a means to enqueue scripts required by the front end, e.g. for live previews. In those cases `$_POST['_wpnonce']` doesn't exist and then `wp_die` halts all script execution and breaks things. I'll add a comment there.

base/inc/actions.php
```diff
@@ -4,14 +4,21 @@
  * Action for displaying the widget preview.
  */
 function siteorigin_widget_preview_widget_action() {
-	if( empty( $_POST['class'] ) ) exit();
-	if ( empty( $_REQUEST['_widgets_nonce'] ) || !wp_verify_nonce( $_REQUEST['_widgets_nonce'], 'widgets_action' ) ) return;
+	if ( empty( $_REQUEST['_widgets_nonce'] ) ||
+	     ! wp_verify_nonce( $_REQUEST['_widgets_nonce'], 'widgets_action' ) ) {
```
gregpriday: Some space indenting here.

PR #561, opened 2017-09-26
Accordion widget

braamgenis:


Misplon:
@braamgenis If possible, please, could we change the default plus and minus icons to the ones in Ionicons.

plus

minus

braamgenis:
@Misplon There are a different pluses and minuses in Ionicons. Could you check the `data-value` attribute of those to see which ones they are exactly?
`data-value="ionicons-android-add"`, `data-value="ionicons-android-remove"`
`data-value="ionicons-ios-plus-empty"`, `data-value="ionicons-ios-minus-empty"`
`data-value="ionicons-plus"`, `data-value="ionicons-minus"`

Misplon:
@braamgenis Please, try:

`ionicons-plus`
`ionicons-minus`

base/inc/array-utils.php
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * From this SO answer: https://stackoverflow.com/a/18781630/3710600
+ *
+ * Works with both integer and string positions.
+ *
+ * @param array      $array
+ * @param int|string $position
+ * @param mixed      $insert
+ */
+function array_insert( &$array, $position, $insert ) {
```
gregpriday: We'd need to prefix this with something SiteOrigin related so we don't get any fatal error conflicts with other plugins.

base/inc/array-utils.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+/**
+ * From this SO answer: https://stackoverflow.com/a/18781630/3710600
```
gregpriday: I haven't checked this, but is code posted on Stack Overflow GPL compatible? Isn't their license creative commons or something? If it's not compatible, I assume this function would just need a rewrite.

base/inc/array-utils.php
braamgenis: :+1:

PR #577, opened 2017-10-18
Presets field

braamgenis:


base/inc/fields/js/presets-field.js
```diff
@@ -0,0 +1,66 @@
...
+				
+				var presetValues = presets[ selectedPreset ].values;
+				
+				var $formContainer = $presetSelect.closest( '.siteorigin-widget-form-main' );
+				var previousValues = $presetSelect.data( 'previousValues' );
+				if ( ! previousValues ) {
+					var presetClone = JSON.parse( JSON.stringify( presetValues ) );
+					var widgetData = sowbForms.getWidgetFormValues( $formContainer );
+					var recurseDepth = 0;
+					var copyValues = function( from, to ) {
```
gregpriday: This could very well be doing something extra, but doesn't `jQuery.extend` offer this functionality already? Can we use that?

https://api.jquery.com/jquery.extend/

braamgenis: Yeah, I did look at using `jQuery.extend` and it's very similar but I didn't see an option to do exactly what we need. We _only_ want the values for properties that are present on the `presets` object to be copied. `jQuery.extend` will copy all properties from source objects, even if they don't exist on the target, which could result in overwriting a user's content changes.

PR #579, opened 2017-10-19
Tabs widget

braamgenis:


Misplon:
@braamgenis Ric has helped us out by tidying up spacing and showing the selected tab so we can see exactly how much margin is between each tab.

tabs-selected

widgets/tabs/tpl/default.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+$last_index = count( $tabs ) - 1;
+
+echo $args['before_title'] . $instance['title'] . $args['after_title'];
+?>
+<div class="sow-tabs">
+	<div class="sow-tabs-tab-container">
+	<?php foreach ( $tabs as $i => $tab ) : ?>
+		<?php
```
gregpriday: Similar to my feedback for the accordion, I'd really like to see this in `get_template_vars` and possibly with a filter.

PR #586, opened 2017-10-25
add ability to select an image size on hero widget

rlamarche:
This PR add a form field to select the format of image in the background of each frame.
It is backward compatible (default value is "Full").

### DIFF
--- a/widgets/hero/hero.php
+++ b/widgets/hero/hero.php
@@ -266,7 +266,7 @@ class SiteOrigin_Widget_Hero_Widget extends SiteOrigin_Widget_Base_Slider {
 	function get_frame_background( $i, $frame ){
 		$background_image = siteorigin_widgets_get_attachment_image_src(
 			$frame['background']['image'],
-			$frame['background']['size'],
+			!empty( $frame['background']['size'] ) ? $frame['background']['size'] : 'full',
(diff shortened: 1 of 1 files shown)


widgets/hero/hero.php
```diff
@@ -261,7 +266,7 @@ function get_widget_form(){
 	function get_frame_background( $i, $frame ){
 		$background_image = siteorigin_widgets_get_attachment_image_src(
 			$frame['background']['image'],
-			'full',
+			$frame['background']['size'],
```
braamgenis: This needs an `empty` check and should fall back to using `'full'` for existing widget instances which won't yet have the `$frame['background']['size']` property.

rlamarche:
You're welcome @braamgenis  :) and thank you for this wonderful wordpress extension, very easy to improve :+1: 
I just made the change you asked. I thought it was an automatic behaviour because I tested the change on an existing widget and I had no problems, but anyway it's checked now ;)

braamgenis:
@rlamarche Yeah, it does "work" with existing instances, but the `size` field is an empty string which may have unintended consequences for some, so better to make completely sure the default is what we expect it to be. It's better to be careful when you have such a large number of active installations. :wink:

rlamarche:
Ok I fully agree with you, and thank you for the merge :+1:

PR #588, opened 2017-10-31
Add noopener for all instances of _blank

AlexGStapleton:
[Reported here](https://siteorigin.com/thread/widget-image-security-problem-no-relnoopener-tag/)

Reference: https://mathiasbynens.github.io/rel-noopener
[Slack Reference](https://siteorigin.slack.com/archives/C02SXRL8N/p1509362961000191)

### DIFF
--- a/base/tpl/posts-selector/post.html
+++ b/base/tpl/posts-selector/post.html
@@ -1,6 +1,6 @@
 <div class="sow-post-selector-summary ui-draggable" data-id="<%= id %>">
 	<div class="sow-thumbnail">
-		<a href="<%= editUrl %>" target="_blank" rel="noopener">
+		<a href="<%= editUrl %>" target="_blank" rel="noopener noreferrer">
 			<img src="<%= thumbnail %>" />
 		</a>
 	</div>
--- a/base/tpl/posts-selector/post.html
+++ b/base/tpl/posts-selector/post.html
@@ -1,6 +1,6 @@
 <div class="sow-post-selector-summary ui-draggable" data-id="<%= id %>">
 	<div class="sow-thumbnail">
-		<a href="<%= editUrl %>" target="_blank">
+		<a href="<%= editUrl %>" target="_blank" rel="noopener">
 			<img src="<%= thumbnail %>" />
 		</a>
 	</div>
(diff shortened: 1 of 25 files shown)


widgets/button/button.php
```diff
@@ -242,6 +242,7 @@ function get_template_variables( $instance, $args ) {
 
 		if ( ! empty( $instance['new_window'] ) ) {
 			$button_attributes['target'] = '_blank';
+			$button_attributes['rel'] = 'noopener';
```
braamgenis: This will be overidden by the `$attributes['rel']` property a few lines below.

base/tpl/posts-selector/post.html
```diff
@@ -1,6 +1,6 @@
 <div class="sow-post-selector-summary ui-draggable" data-id="<%= id %>">
 	<div class="sow-thumbnail">
-		<a href="<%= editUrl %>" target="_blank">
+		<a href="<%= editUrl %>" target="_blank" rel="noopener">
```
braamgenis: I left this one out because I don't think we use these templates anymore. I'll log as an issue to check later.

AlexGStapleton:
Oh whoops  - I didn't sync the follow up commit where I had applied noopener to the rest of the links.


I had a look at js/jquery.slider.js but it didn't appear to be the source - there's something else adding the link directly through HTML. I could never get the JavaScript to affect the resulting window open. Maybe I'm missing something. :S

braamgenis:
I was referring to this manual handling of links: https://github.com/siteorigin/so-widgets-bundle/blob/7b1c7b5b91a27ece3a4efcbc6fc57657ea2ae438/js/slider/jquery.slider.js#L87
But yeah, I couldn't get that to fire now either for some reason. I'll add another issue to see what's happening there, but I think we should still add the rel="noopener" there while we're at it.

braamgenis:
And then you still need to prevent the button's rel field from overriding the `noopener` value and add it to the teaser link in `widgets/accordion/accordion.php`

braamgenis:
And now it's time for you to hate Microsoft: http://caniuse.com/#feat=rel-noopener
So it seems we also need to add `noreferrer` to all the above places to at least cover MS Edge too. ( http://caniuse.com/#feat=rel-noreferrer )

AlexGStapleton:
Regarding the slider - sorry, that's what I was actually referring too - I should have been more specific.
According to this, you simply need to set noopenner as the third 

Ah crap... didn't push the commit again. Sorry. :(

Regarding Microsoft. :/ Just a sec.

braamgenis:
Finally all done! :+1:

PR #593, opened 2017-11-01
Accordion deep linking

braamgenis:
Base functionality required to enable deep linking.

### DIFF
--- a/widgets/accordion/tpl/default.php
+++ b/widgets/accordion/tpl/default.php
@@ -9,16 +9,9 @@
 ?>
 <div>
 	<div class="sow-accordion">
-	<?php foreach ( $panels as $i => $panel ) : ?>
-		<?php
-		if ( empty( $panel['title'] ) ) {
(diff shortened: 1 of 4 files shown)


widgets/accordion/tpl/default.php
```diff
@@ -9,8 +9,16 @@
 ?>
 <div>
 	<div class="sow-accordion">
-	<?php foreach ( $panels as $panel ) : ?>
-		<div class="sow-accordion-panel<?php if ( $panel['initial_state'] == 'open' ) echo ' sow-accordion-panel-open'; ?>">
+	<?php foreach ( $panels as $i => $panel ) : ?>
+		<?php
```
gregpriday: I feel like this code block is big enough that it should go in the widget's `get_template_vars`. Would just need to add `anchor` to the `$panels[$i]` array.

gregpriday: Another thing to consider is adding filters more often. The anchor is a good example of something that a developer might want to filter.

Issue #617, opened 2017-12-05
Accordion: Large titles or limited space will result in icon moving below title


AlexGStapleton:
If you set an long enough accordion title to be two lines, the expand icon will be on the same line as the last line:

![](https://i.imgur.com/yxrNpsz.png)

The same applies if there's not enough available space:

![](https://i.imgur.com/wVDPleA.gif)

Ideally, we should account for this and vertically center the expand icon relative to the title size.

acollis:
Here's how you fix this:

/* Fixes Accordion Title width */
.sow-accordion-title{
  width: calc(100% - 20px);
}

PR #623, opened 2017-12-15
Accordion overflow

acollis:
This fixes long title overflow and the toggle expand / collapse button location. 

/* Fixes Accordion Title width */
.sow-accordion-title{
width: calc(100% - 20px);
}


AlexGStapleton:
I'm not too sure how, but there's an issue with your PR. It's a duplicate of #622 which is a fix for a different issue and unlike the CSS you provided in #617.

acollis:
Yeah I realized afterwards that I accidentally did it against 622 instead.   Sorry about that but glad you caught it.

Adam

PR #630, opened 2017-12-25
Social buttons mobile alignment

### LINKED ISSUE #485: Social Media Buttons: Mobile Alignment Options
A semi-request for the SiteOrigin Social Media Buttons widget is the ability to set mobile alignment as it's quite common that users will set the alignment to left or right but want the icons centered on mobile.

AlexGStapleton:
Resolve #485
Has support for other builders by a dedicated breakpoint field. Default collapse value is 780 (default Page Builder width) or whatever the responsive Page Builder breakpoint is set to.


widgets/social-media-buttons/social-media-buttons.php
```diff
@@ -143,6 +143,17 @@ function get_widget_form(){
 							'justify' => __( 'Justify', 'so-widgets-bundle' ),
 						),
 					),
+					'mobile_align'      => array(
+						'type'    => 'select',
+						'label'   => __( 'Mobile Alignment', 'so-widgets-bundle' ),
```
braamgenis: The field above this one is labelled 'Align', so for consistency this should be 'Mobile align'.

widgets/social-media-buttons/social-media-buttons.php
```diff
@@ -154,6 +165,13 @@ function get_widget_form(){
 							'0.4' => __( 'Very high', 'so-widgets-bundle' ),
 						),
 					),
+					'mobile_width' => array(
+						'type' => 'text',
+						'label' => __( 'Mobile Collapse Width', 'so-widgets-bundle' ),
+						'description' => __( 'The resolution to treat as a mobile resolution.', 'so-widgets-bundle' ),
+						'default' => ( function_exists( 'siteorigin_panels_render' ) ) ? siteorigin_panels_setting( 'mobile-width' ) : 780,
```
braamgenis: Same here again. I don't think it's a good idea to have dependencies on other plugins (even optional). It feels like we have a need for a way to set break points globally, not just on a per widget basis, but for the entire bundle. 🤔

AlexGStapleton: I'm going to move this setting to a widget setting. I still think we should do a SiteOrigin Page Builder settings check for that setting as it's a minor thing (it's purely for a default) and it's not like it's a third party plugin - it's ours.

braamgenis: Creating a dependency means that a future change in PB could break this code. So if you _really_ want to use it, you'll need to do more checking to avoid that. E.g. you're checking for `siteorigin_panels_render`, but using `siteorigin_panels_setting`. Rather check that `siteorigin_panels_setting` exists. Also make sure that `siteorigin_panels_setting('mobile-width')` returns a valid value.

widgets/social-media-buttons/social-media-buttons.php
```diff
@@ -216,12 +238,15 @@ function get_less_variables( $instance ) {
...
-			'rounding'  => $design['rounding'] . 'em',
-			'padding'   => $design['padding'] . 'em',
-			'align'     => $design['align'],
-			'margin'    => $margin
+			'icon_size'             => $design['icon_size'] . 'em',
+			'rounding'              => $design['rounding'] . 'em',
+			'padding'               => $design['padding'] . 'em',
+			'align'                 => $design['align'],
+			'mobile_align'          => $design['mobile_align'],
+			'responsive_breakpoint' => $global_settings['responsive_breakpoint'],
```
braamgenis: Another `empty` check here.

widgets/social-media-buttons/social-media-buttons.php
AlexGStapleton: I was checking siteorigin_panels_render based on the layout slider but in hindsight, that makes more sense.

AlexGStapleton: Regardless, I don't agree with not using the mobile width but to make things easier for you I'm going to exclude this check.

Issue #587, opened 2017-10-30
Contact: Add ability to set onclick for submit button


AlexGStapleton:
We should add onclick support for the contact submit button to easily allow for Adwords (conversion) and analytics (event) tracking.

([Slack Reference](https://siteorigin.slack.com/archives/C02SYBBH2/p1509339845000083))

AlexGStapleton:
Resolved in #631

Issue #562, opened 2017-09-27
Contact: Add Placeholder for field type


AlexGStapleton:
We should add a placeholder field type to the contact form will default too rather than the name field. This will prevent confusion when users add new fields and wonder why certain fields aren't appearing in the contact form.

![](https://vgy.me/Z8vAer.png)

If possible, we should also hide the label, description and required fields repeater until a field type is selected.

Issue #629, opened 2017-12-25
Contact: Add alignment options for submit button


AlexGStapleton:
Certain designs require the contact form submit button to have its placement altered and currently the only way to do this is via CSS.

Issue #198, opened 2016-04-30
Contact Form: Ability to control the width of the submit button


AlexGStapleton:
You should be able to adjust the width of the submit button. Be it a checkbox to make the button full width or a measurement field.

PR #631, opened 2017-12-25
Contact adjustments

AlexGStapleton:
Resolves the following issues: #198, #629, #197 and #139

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -577,6 +577,16 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 								'type'    => 'measurement',
 								'label'   => __( 'Width', 'so-widgets-bundle' ),
 							),
+							'align'    => array(
+								'type'    => 'select',
+								'label'   => __( 'Align', 'so-widgets-bundle' ),
+								'default' => 'left',
+								'options' => array(
+									'left'    => __( 'Left', 'so-widgets-bundle' ),
+									'right'   => __( 'Right', 'so-widgets-bundle' ),
+									'center'  => __( 'Center', 'so-widgets-bundle' ),
+								)
+							),
 							'inset_highlight'     => array(
 								'type'        => 'slider',
 								'label'       => __( 'Inset highlight', 'so-widgets-bundle' ),
@@ -783,6 +793,7 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 			'submit_weight'              => $instance['design']['submit']['weight'],
 			'submit_padding'             => $instance['design']['submit']['padding'],			
 			'submit_width'               => $instance['design']['submit']['width'],
+			'submit_align'               => $instance['design']['submit']['align'],
 			'submit_inset_highlight'     => $instance['design']['submit']['inset_highlight'] . '%',
 
 			// Input focus styles
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -573,6 +573,10 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 								'label'   => __( 'Padding', 'so-widgets-bundle' ),
 								'default' => '10px',
 							),
+							'width' => array(
+								'type'    => 'measurement',
+								'label'   => __( 'Width', 'so-widgets-bundle' ),
+							),
 							'inset_highlight'     => array(
 								'type'        => 'slider',
 								'label'       => __( 'Inset highlight', 'so-widgets-bundle' ),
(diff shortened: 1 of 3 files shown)


AlexGStapleton:
Additional commit resolves #562

AlexGStapleton:
New commit resolves #587

widgets/contact/contact.php
```diff
@@ -1066,6 +1066,9 @@ function contact_form_action( $instance, $storage_hash ) {
 				$errors['_general']['send'] = $success->get_error_message();
 			} else if ( ! $success ) {
 				$errors['_general']['send'] = __( 'Error sending email, please try again later.', 'so-widgets-bundle' );
+			} else {
+				// This action to allow other plugins to run code when contact form has successfully been sent 
+				do_action( 'siteorigin_widgets_contact_sent', $instance, $email_fields );
```
braamgenis: Nice addition. Perhaps we should also add an action when an error occurs?

widgets/contact/contact.php
```diff
@@ -712,19 +721,24 @@ function modify_instance( $instance ) {
...
 		unset( $instance['panels_info'] );
 
 		// Include '_sow_form_id' in generation of 'instance_hash' to allow multiple instances of the same form on a page.
-		$vars['instance_hash'] = md5( serialize( $instance ) );
-
+		$instance_hash = md5( serialize( $instance ) );
 		unset( $instance['_sow_form_id'] );
 
-		return $vars;
+		if ( ! empty( $instance['settings']['submit_attributes']['submit_id'] ) ) {
```
braamgenis: I don't see `submit_attributes` or `submit_id` being defined anywhere? Is there a commit you forgot to push?

AlexGStapleton: Last second change. 😞

widgets/contact/contact.php
AlexGStapleton: 👍

AlexGStapleton:
Incoming Codacy bad code result. There's no reason not to use `} else {` in this situation imo.

widgets/contact/contact.php
```diff
@@ -730,7 +730,7 @@ function get_template_variables( $instance, $args ) {
 		$instance_hash = md5( serialize( $instance ) );
 		unset( $instance['_sow_form_id'] );
 
-		if ( ! empty( $instance['settings']['submit_attributes']['submit_id'] ) ) {
+		if ( ! empty( $instance['settings']['submit_id'] ) ) {
```
braamgenis: This was defined as `id` in the `settings` form field, so either need to change this to `$instance['settings']['id']` or change the form field key to `submit_id`

AlexGStapleton: Whoops - sorry.

widgets/contact/contact.php
```diff
@@ -697,19 +721,24 @@ function modify_instance( $instance ) {
...
 
 		// Include '_sow_form_id' in generation of 'instance_hash' to allow multiple instances of the same form on a page.
-		$vars['instance_hash'] = md5( serialize( $instance ) );
-
+		$instance_hash = md5( serialize( $instance ) );
 		unset( $instance['_sow_form_id'] );
 
-		return $vars;
+		if ( ! empty( $instance['settings']['submit_id'] ) ) {
+			$submit_attributes['id'] = $instance['settings']['submit_id'];
```
braamgenis: I don't see `$submit_attributes` being declared or initialized anywhere? Is it necessary to make this an `array`? It seems like it only ever contains `'id'`,

widgets/contact/tpl/default.php
```diff
@@ -48,8 +48,9 @@
 		<?php endif; ?>
 
 		<div class="sow-submit-wrapper <?php if( $instance['design']['submit']['styled'] ) echo 'sow-submit-styled' ?>">
-			<input type="submit" value="<?php echo esc_attr( $instance['settings']['submit_text'] ) ?>"
-				   class="sow-submit">
+			<input type="submit" value="<?php echo esc_attr( $instance['settings']['submit_text'] ) ?>" class="sow-submit"
+				   <?php foreach( $submit_attributes as $name => $val ) echo $name . '="' . esc_attr( $val ) . '" ' ?>
```
braamgenis: Seeing as `$submit_attributes` only contains `'id'` this can just rather just use a variable called `$id` and `echo` without the loop.

widgets/contact/contact.php
AlexGStapleton: That's correct. I based this purely on how we handle other buttons in SiteOrigin Widgets Bundle - they all have an option of outputting additional values based on the contents of `$submit_attributes`. This allows other developers to easily add additional attributes to the button without having to override the template.

If you would rather not do this for this button, that's fine - just say the word and I'll remove it.

braamgenis: Ok that's cool. Then we'd still need `$submit_attributes` to be declared and initialized before using it. And to allow third party devs to alter it's contents you'd need to add a filter.

Issue #612, opened 2017-11-30
Hero Widget: Unable to remove text shadow from paragraph text


Misplon:
In the Hero widget design section we have: Heading shadow intensity. However, shadow is also applied to paragraph text. So right now we don't have a way to change the paragraph text shadow or to remove it. I suggest we consider re-labelling the shadow intensity field and apply it to both headings and text.

@braamgenis What do you reckon?

braamgenis:
So it's not the heading text shadow that's being applied to the `<p>` text. We have this in the stylesheet for the hero widget:
```
p {
    color: @text_color;
    text-shadow: 0 1px 2px rgba(0,0,0,0.25);
    margin: 1em 0;
    font-size: @text_size;
}
```
So we could perhaps add another field to make the `<p>` shadow configurable?

Misplon:
Perfect. That'll be great.

Issue #504, opened 2017-07-05
Hero: Add link color picker


AlexGStapleton:
We have a header and text color but not a link color picker. We really should add a color picker for links as neither of the other two colors are likely to affect links.

PR #636, opened 2017-12-26
Hero adjustments

### LINKED ISSUE #100: Hero Widget: Add font family setting for paragraphs
There is only a setting for headings at the moment, let's also add a font-family setting for paragraphs in the Hero widget.

AlexGStapleton:
Resolves #100, #504 and #612

### DIFF
--- a/widgets/hero/hero.php
+++ b/widgets/hero/hero.php
@@ -385,17 +385,18 @@ class SiteOrigin_Widget_Hero_Widget extends SiteOrigin_Widget_Base_Slider {
 		$less['link_color'] = ! empty( $instance['design']['link_color'] ) ? $instance['design']['link_color'] : '';
 		$less['link_color_hover'] = ! empty( $instance['design']['link_color_hover'] ) ? $instance['design']['link_color_hover'] : '';
 
-		$text_font = siteorigin_widget_get_font( $instance['design']['text_font'] );
-		$less['text_font'] = $text_font['family'];
-		if ( ! empty( $text_font['weight'] ) ) {
-			$less['text_font_weight'] = $text_font['weight'];
-		}
 
-		if ( ! empty( $instance['design']['heading_font'] ) ) {
-			$heading_font = siteorigin_widget_get_font( $instance['design']['heading_font'] );
-			$less['heading_font'] = $heading_font['family'];
-			if ( ! empty( $heading_font['weight'] ) ) {
-				$less['heading_font_weight'] = $heading_font['weight'];
+		$heading_font = siteorigin_widget_get_font( $instance['design']['heading_font'] );
+		$less['heading_font'] = $heading_font['family'];
+		if ( ! empty( $heading_font['weight'] ) ) {
+			$less['heading_font_weight'] = $heading_font['weight'];
+		}
+		
+		if ( ! empty( $instance['design']['text_font'] ) ) {
+			$text_font = siteorigin_widget_get_font( $instance['design']['text_font'] );
+			$less['text_font'] = $text_font['family'];
+			if ( ! empty( $text_font['weight'] ) ) {
+				$less['text_font_weight'] = $text_font['weight'];
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


widgets/hero/hero.php
```diff
@@ -368,10 +372,16 @@ function get_less_variables($instance) {
 		$less['text_color'] = $instance['design']['text_color'];
 		
 
-		$font = siteorigin_widget_get_font( $instance['design']['heading_font'] );
-		$less['heading_font'] = $font['family'];
-		if ( ! empty( $font['weight'] ) ) {
-			$less['heading_font_weight'] = $font['weight'];
+		$text_font = siteorigin_widget_get_font( $instance['design']['text_font'] );
```
braamgenis: `empty` check here

widgets/hero/hero.php
```diff
@@ -394,13 +404,11 @@ function add_default_measurement_unit($val) {
...
-		if( empty( $instance ) ) return;
-
-		$font_import = siteorigin_widget_get_font( $instance['design']['heading_font'] );
-		if( !empty( $font_import['css_import'] ) ) {
-			return  $font_import['css_import'];
-		}
+	function get_google_font_fields( $instance ) {
+		return array(
+			$instance['design']['heading_font'],
+			$instance['design']['text_font'],
```
braamgenis: `empty` check here

widgets/hero/hero.php
```diff
@@ -370,7 +379,9 @@ function get_less_variables($instance) {
 		$less['heading_color'] = $instance['design']['heading_color'];
 		$less['text_shadow'] = intval( $instance['design']['text_shadow'] );
 		$less['text_color'] = $instance['design']['text_color'];
-		
+
+		$less['link_color'] = $instance['design']['link_color'];
+		$less['link_color_hover'] = $instance['design']['link_color_hover'];
```
braamgenis: `empty` checks for the above two.

widgets/hero/hero.php
```diff
@@ -355,9 +363,10 @@ function get_less_variables($instance) {
 		}
 
 		$less['heading_shadow'] = intval( $instance['design']['heading_shadow'] );
-
 		$less['heading_color'] = $instance['design']['heading_color'];
+		$less['text_shadow'] = intval( $instance['design']['text_shadow'] );
```
braamgenis: `empty` check here.

widgets/hero/styles/default.less
```diff
@@ -32,7 +33,9 @@
 			h1,h2,h3,h4,h5,h6{
 				line-height: 1.375em;
 				color: @heading_color;
-				text-shadow: 0 2px 2px rgba( 0, 0, 0, @heading_shadow/100 );
+				& when not ( @heading_shadow = 0 ) {
+					text-shadow: 0 2px 2px rgba( 0, 0, 0, @heading_shadow/100 );
```
braamgenis: I recently made a change which allows float values in the slider. I.e. you can set min and max to 0 and 1 and step to 0.01

widgets/hero/hero.php
AlexGStapleton: Ah crap. Missed this. :/

widgets/hero/styles/default.less
AlexGStapleton: I've adjusted the text shadow to reflect this change. Is there any sort of built-in migration code in the SiteOrigin Widgets Bundle? I ask as that'll make it easier to handle the transition between the current heading_shadow and the new value.

braamgenis: Not purpose built for migrations, but we'd normally use the `modify_instance` function to make any required changes. Would just need to try test thoroughly to make sure it does the migration correctly in all cases. Perhaps less risky to leave `heading_shadow` as is... 🤔

AlexGStapleton: Ah. So that's what that does (or rather can do). I would rather leave this as is for the time being as there's no real harm in leaving it.

braamgenis: Ok, awesome. Merging.

PR #637, opened 2017-12-27
Features: Add ability to control responsive breakpoint

### LINKED ISSUE #298: Features: Change media query to reflect site settings

AlexGStapleton:
Replaces #372. Resolves #298

This PR adds a global widget setting that adds the ability to control the features breakpoint. If SiteOrigin Page Builder is installed, it'll default to the global mobile breakpoint but the global widget setting is still present if the user wishes to override the Page Builder setting for the features widgets.

### DIFF
--- a/widgets/features/features.php
+++ b/widgets/features/features.php
@@ -276,11 +276,38 @@ class SiteOrigin_Widget_Features_Widget extends SiteOrigin_Widget {
 			}
 		}
 
-        $less_vars['container_size'] = $instance['container_size'];
-        $less_vars['icon_size'] = $instance['icon_size'];
-        $less_vars['use_icon_size'] = empty( $instance['icon_size_custom'] ) ? 'false' : 'true';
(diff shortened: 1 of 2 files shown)


widgets/features/features.php
```diff
@@ -276,13 +276,40 @@ function get_less_variables( $instance ) {
...
 
 		return $less_vars;
 	}
 
+	function get_settings_form() {
+		return array(
+			'responsive_breakpoint' => array(
+				'type'        => 'number',
+				'label'       => __( 'Responsive Breakpoint', 'so-widgets-bundle' ),
+				'default'     => ( ! function_exists( 'siteorigin_panels_setting' ) ? 520 : false ), // We want to default to the global setting if Page Builder is enabled so we're not going to set anything
```
braamgenis: As above, I don't think we should add the complexity of checking for PB here. Just use the previous default of 520px. I think there's also some risk here that we change the value for existing users.

AlexGStapleton:
How does that look?

PR #651, opened 2018-01-29
Features: Better feature padding removal on row ends

AlexGStapleton:
This fixes an overflow issue on full width stretched features widgets. This issue only occurs in themes that don't hide overflow on their theme/content container.

[Related thread](https://wordpress.org/support/topic/page-width-exceeds/#post-9913254)

### DIFF
--- a/widgets/features/styles/default.less
+++ b/widgets/features/styles/default.less
@@ -16,6 +16,7 @@
 @more_text_font_weight: 400;
 @more_text_size: default;
 @more_text_color: default;
+@per_row: 3;
 
 @container_size: 84px;
 @icon_size: 24px;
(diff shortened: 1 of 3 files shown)


AlexGStapleton:
https://github.com/siteorigin/so-widgets-bundle/pull/651/files#diff-fed4b4e42ad9831119a96c6cea1980d1L10 < Based on testing, this no longer appeared to be required due to the flex based system we use now.

widgets/features/styles/default.less
```diff
@@ -36,6 +34,14 @@
         padding: 0 25px;
         display: flex;
 
+        &:nth-of-type(@{per_row}n+1) {
+            padding-left: 0;
+        }
+
+        &:last-of-type, &:nth-of-type(@{per_row}n) {
```
braamgenis: This `last-of-type` selector will result in the icon being misaligned inside the `.sow-features-feature` container if you don't have an exact multiple of the number of features per row and it didn't seem to be necessary in my quick tests.

PR #653, opened 2018-02-05
Sliders: WCAG 2.4.4

AlexGStapleton:
Similiar to #595 

This PR will allow for the SiteOrigin Slider to comply with [WCAG 2.4.4](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-refs.html) by using [ARIA8](https://www.w3.org/TR/WCAG20-TECHS/ARIA8.html). At least the base slider will comply. The contents may not comply but that's due to the selected widgets. Tested using [WAVE](http://wave.webaim.org/).

### DIFF
--- a/base/inc/widgets/base-slider.class.php
+++ b/base/inc/widgets/base-slider.class.php
@@ -305,7 +305,7 @@ abstract class SiteOrigin_Widget_Base_Slider extends SiteOrigin_Widget {
 	 */
 	function video_code( $videos, $classes = array() ){
 		if( empty( $videos ) ) return;
-		$video_element = '<video class="' . esc_attr( implode( ',', $classes ) ) . '" autoplay loop muted>';
+		$video_element = '<video class="' . esc_attr( implode( ',', $classes ) ) . '" autoplay loop muted playsinline>';
 
 		foreach( $videos as $video ) {
 			if( empty( $video['file'] ) && empty ( $video['url'] ) ) continue;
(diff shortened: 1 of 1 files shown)


base/inc/widgets/base-slider.class.php
```diff
@@ -172,18 +172,18 @@ function render_template_part( $part, $controls, $frames ) {
 				?>
 				<ol class="sow-slider-pagination">
 					<?php foreach($frames as $i => $frame) : ?>
-						<li><a href="#" data-goto="<?php echo $i ?>"><?php echo $i+1 ?></a></li>
+						<li><a href="#" data-goto="<?php echo $i ?>" aria-label="<?php printf( __( 'open slide %s', 'so-widgets-bundle' ), $i+1 ) ?>"><?php echo $i+1 ?></a></li>
```
braamgenis: "Open slide _x_" feels a bit ambiguous and might imply opening the slide's image or content somehow.

AlexGStapleton: What would you recommend instead?

braamgenis: Perhaps "Show slide _x_" or "Display slide _x_"?

AlexGStapleton: I like both of those, but... hm... display sounds nicer to me.

PR #657, opened 2018-02-13
Hero: Add responsive height settings

AlexGStapleton:
This PR adds responsive height settings for the hero widget. The breakpoint is controlled via the global widget settings.

### DIFF
--- a/widgets/hero/hero.php
+++ b/widgets/hero/hero.php
@@ -162,6 +162,12 @@ class SiteOrigin_Widget_Hero_Widget extends SiteOrigin_Widget_Base_Slider {
 						'default' => 'default',
 					),
 
+					'height_responsive' => array(
+						'type' => 'measurement',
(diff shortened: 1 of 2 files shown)


AlexGStapleton:
@braamgenis is there a good way to check if something is set when it was removed due to being set to default? I ask because currently if the responsive height isn't set, it'll still output the media query.

Also, what are your thoughts on having the responsive height only be selectable if a normal height is set? That way we won't have people mixing the automatic height and the hard height.

braamgenis:
I use `& when( condition )` to prevent outputting empty styles. So wrapping the media query in something like `& when ( isnumber( @slide_height_responsive ) )` should do it.

braamgenis:
And unless it specifically breaks something, I'd leave the responsive height even if they don't have a normal fixed height.

AlexGStapleton:
Oh! I must have misplaced the notification email for this - sorry!

Okay cool. I'll make the recommended changes shortly.

PR #663, opened 2018-03-09
Release 1.11.5

braamgenis:
* Features: Better feature padding removal on row ends.
* Sliders: WCAG 2.4.4 compliance.
* Corrected video attributes.
* Tabs: Hide widget title when no title is set.
* TinyMCE field: Added setting for `wpautop` processing which is on by default.
* Contact: When Gradient disabled, set basic background.
* Beaver Builder compat: Only set SOWB widget form values when editing a SOWB widget.
* Contact: Option to log submitter's IP address.
* Add random number and set `more_entropy` to increase chance of unique form ids.
* Contact: Added 'tel' field type which should show numeric keyboard on mobile.
* Media field: Trigger change event when removing selected image.
* Renamed the PHP LESS parser to `SiteOrigin_LessC` to avoid conflicts.
* Date range field: Prevent initializing date range fields multiple times and ensure date format consistent.
* Register pikaday as common script and enqueue as needed in fields.
* Google Map: Show satellite map type.
* Translation: Add context to From strings.
* Add missing semicolons to Ionicons codes.


base/inc/fields/js/date-range-field.js
```diff
@@ -1,23 +1,40 @@
...
-
-		if ( $( this ).find( '[class*="sowb-specific-date"]' ).length > 0 ) {
+		if ( $dateRangeField.find( '[class*="sowb-specific-date"]' ).length > 0 ) {
 			var createPikadayInput = function ( inputName, initVal ) {
-				var $field = $( this ).find( '.' + inputName + '-picker' );
+				var $field = $dateRangeField.find( '.' + inputName + '-picker' );
 				var picker = new Pikaday( {
 					field: $field[ 0 ],
 					blurFieldOnSelect: false,
+					toString( date, format ) {
```
gregpriday: For consistency, we could stick to `toString: function(...) {` here

widgets/contact/contact.php
```diff
@@ -121,6 +121,11 @@ function get_widget_form() {
 							'required_fields[hide]' => array( 'hide' ),
 						)
 					),
+					'log_ip_address' => array(
```
gregpriday: I assume this defaults to false either way, but we should have a `'default' => false` just to clarify.

base/inc/fields/js/date-range-field.js
braamgenis: 👍

widgets/contact/contact.php
braamgenis: 👍

PR #674, opened 2018-03-21
Release 1.11.7

braamgenis:
* Image: Add title alignment setting.
* Button: Add max-width to ensure buttons are responsive.
* Hero: New filter for frame content 'siteorigin_hero_frame_content'
* Features: Don't set margin for one column left/right feature.
* Updated widget icons.
* Updated google web fonts.


widgets/accordion/js/accordion.js
```diff
@@ -99,6 +100,13 @@ jQuery( function ( $ ) {
 				} else {
 					updateHash();
 				}
+				if ( initialScrollPanel > 0 ) {
```
gregpriday: Couldn't this just be `if ( initialScrollPanel ) {`, or is that `> 0` needed?

braamgenis: The `> 0` is just a bit of extra protection against silly values like `-1` etc. :suspect:

PR #697, opened 2018-04-17
Accordion/Tabs: Unable to use Masonry widget if not visible by default

AlexGStapleton:
This issue is similar to siteorigin/siteorigin-premium#133, and siteorigin/siteorigin-premium#129. Basically, add an accordion or tab. Create a non-visible (by default) item and set the content type to Layout Builder. Add a Simple Masonry widget and set it up as desired. Save and view the results. The simple masonry won't appear as the Simple Masonry widget calculates sizing on load so we need to add a check/trigger for Simple Masonry to avoid this, which this PR does. I wasn't sure about the comment to leave so I just copied the existing one used in the Google Maps widget.

### DIFF
--- a/widgets/simple-masonry/js/simple-masonry.js
+++ b/widgets/simple-masonry/js/simple-masonry.js
@@ -64,6 +64,16 @@ jQuery( function($){
 				});
 			});
 		};
+		
+		// If the Simple Masonry container is hidden it won't display properly. This is an attempt to make it display by
+		// calling resize when a custom 'show' event is fired. The 'show' event is something we fire in a few widgets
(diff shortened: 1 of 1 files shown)


braamgenis:
@AlexGStapleton Not sure what I was thinking with that change in the maps widget. So I've made it somewhat simpler and more consistent with how we're handling widgets in various other contexts. See ba8c5576492a3fbfcd59e67d3bc603270bacac8a The accordion and tabs widgets now rather trigger `setup_widgets` when they've opened/shown a panel and widgets now just check whether they're visible before actually running their setup. That should fix this and all the other related issues.

PR #706, opened 2018-05-14
Layout Slider: Prevent header color from affecting SiteOrigin Headline

### LINKED ISSUE #432: Layout Slider Overrides Headline color setting
The layout slider will override the SiteOrigin Headline widget's headline and sub-headline colors. Ideally, the layout slider should only be able to override things like the text widget and the editor widget and shouldn't be able to override the headline widget.

AlexGStapleton:
Resolves #432

### DIFF
--- a/widgets/layout-slider/styles/default.less
+++ b/widgets/layout-slider/styles/default.less
@@ -8,7 +8,6 @@
 @slide_padding_sides: 10px;
 @slide_width: 1280px;
 @slide_height: default;
-@heading_size: 38px;
 @text_size: 16px;
 @text_color: #F6F6F6;
 @heading_shadow: 50;
@@ -25,11 +24,13 @@
 			max-width: @slide_width;
 			height: @slide_height;
 
-			h1,h2,h3,h4,h5,h6{
-				line-height: 1.375em;
-				color: @heading_color;
-				text-shadow: 0 2px 2px rgba( 0, 0, 0, @heading_shadow/100 );
-				margin: 0.1em 0;
+			.widget:not(.widget_sow-headline) {
+				h1,h2,h3,h4,h5,h6{
+					line-height: 1.375em;
+					color: @heading_color;
+					text-shadow: 0 2px 2px rgba( 0, 0, 0, @heading_shadow/100 );
+					margin: 0.1em 0;
+				}
 			}
 
 			p{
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
@Misplon Can you think of any other widgets with header color options that should be excluded?

Misplon:
Not right now. Let's start with this and add further if there are additional reports.

AlexGStapleton:
That's a decent idea.

Please note that change will ensure that the header colors remain for everything _but_ the SiteOrigin Headline widget - no other widget will be affected by this change.

Misplon:
@AlexGStapleton Was there anything you needed to add here?

AlexGStapleton:
Honestly, I'm not too sure. 😕

braamgenis:
Yeah, so I was suggesting we should add an optional setting to prevent this for _all_ widgets and not just Headline, but now I'm thinking we should remove the setting altogether. My guess is this setting was meant for the Hero widget, which has a TinyMCE field as it's content editor. IMO it doesn't really make sense for the Layout Slider which could contain widgets with their own header color settings.

AlexGStapleton:
The only potential issue I see with removing this setting is that certain users may already be relying on these settings. Is that something we're okay with? I know @Misplon is typically very against these sort of setting removals.

braamgenis:
Yeah, usually I'm very against it too, but in this case I think leaving it will cause more unintented consequences down the line. You guys will have a better idea of how many users this might affect though, so if you and @Misplon think we should just stick to fixing it for Headline widget, I'm happy to merge as is.

Misplon:
I'm cool with @braamgenis suggestion here (weighing up the greatest cost).

braamgenis:
Yeah, I would say so. `heading_size` isn't even being used at all. Perhaps as a compromise, we can remove the form fields and keep using the settings if they exist for users who may have been relying on them,

AlexGStapleton:
I can't figure out a way to dynamically filter out widget forms while still providing access to the forms to those that have already set them up. I know we've spoken about this previously Braam but, do you have any ideas?

braamgenis:
@AlexGStapleton I'll have to dig around a bit for this one. Will take a proper look as soon as some other higher priority issues have been taken care of.

AlexGStapleton:
That's fine and completely understandable.

braamgenis:
Fixed in a066c334efef2449a8a812ff550cf944ebca8119

Misplon:
Thanks brother, looking good.

PR #707, opened 2018-05-15
Accordion/Tabs FitText display issue fix

### LINKED ISSUE #698: Accordion/Tabs: FitText doesn't work until browser is resized
This issue is exactly the same one as #697.

Basically, add an accordion or tab. Create a non-visible (by default) item and set the content type to Layout Builder. Add a Hero widget and set it up as desired. Save and view the results. The text won't appear until you resize the browser as the text size is calculated on load.

AlexGStapleton:
Resolves #698. This may not be the best fix as it's kind of just thrown on, but from testing, it does its job and ensures FitText is readable and correctly sized.

### DIFF
--- a/js/sow.jquery.fittext.js
+++ b/js/sow.jquery.fittext.js
@@ -35,7 +35,7 @@ var sowb = window.sowb || {};
             resizer();
 
             // Call on resize. Opera debounces their resize by default.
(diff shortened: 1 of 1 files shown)


braamgenis:
This was also fixed by ba8c5576492a3fbfcd59e67d3bc603270bacac8a

PR #709, opened 2018-05-23
Features: Rename container settings

AlexGStapleton:
A number of users have asked why the text doesn't appear inside of the container and while that would be nice, that's not what the container currently is - they're Icon Container. This PR will rename the setting labels to state that.

### DIFF
--- a/widgets/features/features.php
+++ b/widgets/features/features.php
@@ -194,14 +194,14 @@ class SiteOrigin_Widget_Features_Widget extends SiteOrigin_Widget {
 
 			'container_shape' => array(
 				'type' => 'select',
-				'label' => __('Container shape', 'so-widgets-bundle'),
+				'label' => __('Icon Container Position', 'so-widgets-bundle'),
 				'default' => 'round',
 				'options' => include dirname( __FILE__ ) . '/inc/containers.php',
 			),
 
 			'container_size' => array(
 				'type' => 'measurement',
-				'label' => __('Container size', 'so-widgets-bundle'),
+				'label' => __('Icon Container Size', 'so-widgets-bundle'),
 				'default' => '84px',
 			),
 
--- a/widgets/features/features.php
+++ b/widgets/features/features.php
@@ -194,7 +194,7 @@ class SiteOrigin_Widget_Features_Widget extends SiteOrigin_Widget {
 
 			'container_shape' => array(
 				'type' => 'select',
-				'label' => __( 'Container shape', 'so-widgets-bundle' ),
+				'label' => __( 'Icon container shape', 'so-widgets-bundle' ),
 				'default' => 'round',
 				'options' => include dirname( __FILE__ ) . '/inc/containers.php',
 			),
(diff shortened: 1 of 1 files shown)


widgets/features/features.php
```diff
@@ -52,13 +52,13 @@ function get_widget_form(){
 
 					'container_color' => array(
 						'type' => 'color',
-						'label' => __('Container color', 'so-widgets-bundle'),
+						'label' => __('Icon Container Color', 'so-widgets-bundle'),
```
braamgenis: To be consistent with other labels in this form this should us sentence case, i.e. _Icon container color_.

widgets/features/features.php
```diff
@@ -52,13 +52,13 @@ function get_widget_form(){
...
 						'default' => '#404040',
 					),
 
                     'container_position' => array(
                         'type' => 'select',
-                        'label' => __('Container Position', 'so-widgets-bundle'),
+                        'label' => __('Icon Container Position', 'so-widgets-bundle'),
```
braamgenis: Sentence case here: _Icon container position_

widgets/features/features.php
```diff
@@ -194,14 +194,14 @@ function get_widget_form(){
 
 			'container_shape' => array(
 				'type' => 'select',
-				'label' => __('Container shape', 'so-widgets-bundle'),
+				'label' => __('Icon Container Position', 'so-widgets-bundle'),
```
braamgenis: Looks like this was accidentally changed from _shape_ to _position_ and sentence case: _Icon container shape_

widgets/features/features.php
```diff
@@ -194,14 +194,14 @@ function get_widget_form(){
...
 				'default' => 'round',
 				'options' => include dirname( __FILE__ ) . '/inc/containers.php',
 			),
 
 			'container_size' => array(
 				'type' => 'measurement',
-				'label' => __('Container size', 'so-widgets-bundle'),
+				'label' => __('Icon Container Size', 'so-widgets-bundle'),
```
braamgenis: Sentence case: _Icon container size_

widgets/features/features.php
```diff
@@ -194,20 +194,20 @@ function get_widget_form(){
 
 			'container_shape' => array(
 				'type' => 'select',
-				'label' => __('Icon Container Position', 'so-widgets-bundle'),
+				'label' => __( 'Icon container position', 'so-widgets-bundle' ),
```
braamgenis: This should be _shape_, shouldn't it? I.e. _Icon container **shape**_

braamgenis:
One more thing, in future please restrict the whitespace edits to lines to which you've had to make code changes. Also, please be sure the whitespace changes are absolutely necessary. All the lines with whitespace-only changes add a lot of noise to the diffs.

widgets/features/features.php
AlexGStapleton: That would align with the setting name but I don't think we should change this however as Shape isn't as obvious as position. While it may make sense for us to change it, it will cause confusion. "What is a Top Shape?"

AlexGStapleton: Oh my mistake.

braamgenis:
Ok, all good. 👍

PR #712, opened 2018-06-07
Add condition get style hash by widget number

dvekotletki:
Hello! We got a problem with using Page Builder in batch with Widgets bundle.
When widget has styles, it generates them once a week when the page is requested.
On this way, we got some middleware before user get the page. We have FCGI and CDN, those two tools cache pages and their static to make it faster for end user.
We trying to figure out how we could fix disappearing widget styles and that what we came to:

We add an option to make `*.css` file name generated with not unique hash everytime, but with sliced md5 of widget-id hash. You can check it here: `base/siteorigin-widget-class.php:1096`

### DIFF
--- a/base/siteorigin-widget.class.php
+++ b/base/siteorigin-widget.class.php
@@ -441,7 +441,7 @@ abstract class SiteOrigin_Widget extends WP_Widget {
 
 		// Filter the instance specifically for the form
 		$instance = apply_filters('siteorigin_widgets_form_instance_' . $this->id_base, $instance, $this);
-		
+
 		// `more_entropy` adds a period to the id.
 		$id = str_replace( '.', '', uniqid( rand(), true ) );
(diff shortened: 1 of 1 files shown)


dvekotletki:
Request with the same purposes you could find in so-widget-bundle here:
https://github.com/siteorigin/siteorigin-panels/pull/543

If you have any questions or comments please write down here.
Thanks in advance.

gregpriday:
Thanks for the submission @dvekotletki. So what I'd prefer to see is a more generic filter that you can make use of in your own code. For example, how about something like this:

```
$style_hash = apply_filters('siteorigin_widgets_widget_style_hash', '', $this);
if( empty( $style_hash ) ) {
    // Generate the style hash the standard way
}
return $style_hash
```

This means you could create your own plugin that creates a hash that fits your needs, without adding a new setting for every single Page Builder user.

PR #718, opened 2018-06-25
Accordion: Expand items automatically when user is printing

AlexGStapleton:
This PR will automatically expand all Accordion items when the user prints the page.


braamgenis:
No, it seems to happen with text-only content too.

AlexGStapleton:
That's odd. Hm. Do you have Background graphics enabled?

braamgenis:
I did not know that was a thing. :suspect:  Ok, all good. 👍

PR #721, opened 2018-06-28
Remove old style after update widget

dvekotletki:
1. Hello, looks like I found a bug. When update widget, old css doesn’t deleted, because to delete_css function transferred `$new_instance` data not old one. Could you check this please ?
2. Also I have one question. Do you have a way to disable clear_file_cache() method ?


dvekotletki:
@gregpriday Plz review

PR #727, opened 2018-08-06
Plugin Action Links: Check if edit link is present before removal

AlexGStapleton:
This PR prevents a fatal error from occurring if another plugin has already removed the editor link.

`HP Fatal error: Uncaught Error: Cannot unset string offsets in wp-content/plugins/so-widgets-bundle/so-widgets-bundle.php:745
`
[Related](https://siteorigin.com/thread/siteorigin-widgets-bundle-vs-wp-editor/)

### DIFF
--- a/so-widgets-bundle.php
+++ b/so-widgets-bundle.php
@@ -742,7 +742,9 @@ class SiteOrigin_Widgets_Bundle {
 	 * Add action links.
 	 */
 	function plugin_action_links($links){
-		if ( isset( $links['edit'] ) ) unset( $links['edit'] );
+		if ( isset( $links['edit'] ) ) {
+			unset( $links['edit'] );
+		}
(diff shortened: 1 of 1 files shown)


so-widgets-bundle.php
```diff
@@ -742,7 +742,7 @@ function load_missing_widget( $the_widget, $class ){
 	 * Add action links.
 	 */
 	function plugin_action_links($links){
-		unset( $links['edit'] );
+		if ( isset( $links['edit'] ) ) unset( $links['edit'] );
```
braamgenis: It's a minor thing, but as per [WordPress style guide](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#brace-style) we should use braces, even when there's only one line in the block.

PR #729, opened 2018-08-13
Social Media Buttons: Add Phone and resolve #725

### LINKED ISSUE #725: Social Media Buttons: Change title text for email button
The Email network option doesn't really make much sense with the default title text - `SITENAME on Envelope`. We should change it to something more logical like, `Email SITENAME`.

[Related](https://siteorigin.com/thread/social-media-buttons-3/)

AlexGStapleton:
This PR adds a new Phone button and resolves #725 while also accounting for the addition of the Phone "network".

### DIFF
--- a/widgets/social-media-buttons/tpl/default.php
+++ b/widgets/social-media-buttons/tpl/default.php
@@ -6,7 +6,19 @@
 		if( !empty($instance['design']['hover']) ) $classes[] = 'ow-button-hover';
 		$classes[] = "sow-social-media-button-" . sanitize_html_class( $network['name'] );
 		$classes[] = "sow-social-media-button";
-		$title = empty( $network['icon_title'] ) ? sprintf( __( '%s on %s', 'so-widgets-bundle' ), get_bloginfo( 'name' ), ucwords( str_replace( '-', ' ', $network['name'] ) ) ) : $network['icon_title'];
+
+		if( ! empty( $network['icon_title'] ) ) {
+			$title = $network['icon_title'];
+		} else {
+			if ( $network['name'] == 'email' || $network['name'] == 'phone' ) {
+				if ( $network['title'] == 'envelope' ) $network['name'] == 'email';
+
+				/* translators: Social Media Button with a verb and then site name. As an example, Call Eastlink Windows */
+				$title = sprintf( __( '%s %s', 'so-widgets-bundle' ), ucfirst( $network['name'] ), get_bloginfo( 'name' ) );
+			} else {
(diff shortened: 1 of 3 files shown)


widgets/social-media-buttons/tpl/default.php
```diff
@@ -6,7 +6,19 @@
 		if( !empty($instance['design']['hover']) ) $classes[] = 'ow-button-hover';
 		$classes[] = "sow-social-media-button-" . sanitize_html_class( $network['name'] );
 		$classes[] = "sow-social-media-button";
-		$title = empty( $network['icon_title'] ) ? sprintf( __( '%s on %s', 'so-widgets-bundle' ), get_bloginfo( 'name' ), ucwords( str_replace( '-', ' ', $network['name'] ) ) ) : $network['icon_title'];
+
+		if( ! empty( $network['icon_title'] ) ) {
+			$title = $network['icon_title'];
+		} else {
+			if ( $network['name'] == 'email' || $network['name'] == 'phone' ) {
+				if ( $network['title'] == 'envelope' ) $network['name'] == 'email';
```
braamgenis: This is quite confusing. Where does the `$network['title']` come from? And then it's just doing a check to see if `$network['name']` is equal to 'email' without assigning it to anything?

widgets/social-media-buttons/tpl/default.php
```diff
@@ -6,7 +6,19 @@
...
+
+				/* translators: Social Media Button with a verb and then site name. As an example, Call Eastlink Windows */
+				$title = sprintf( __( '%s %s', 'so-widgets-bundle' ), ucfirst( $network['name'] ), get_bloginfo( 'name' ) );
```
braamgenis: I think the below format could work for email and phone as is, e.g. Eastlink Windows on Email, but regardless, perhaps we should add a field to let users specify the format they want. Then we can just have our one as a default.

AlexGStapleton: We do actually already have a way for users to specify a custom title, icon_title. The issue with that though is we can't expect users to always fill that field in just to correct the title - for example, "Envelope on Eastlink Windows" is obviously wrong.

With that said, I do like the Sitename on type name a little more than "Email Eastlink Windows", but the problem with that is that both buttons use tel and mailto. By clicking that link you're not taken to somewhere, you're doing something - be it calling or emailing.

widgets/social-media-buttons/tpl/default.php
AlexGStapleton: That actually contains two major typos. 😞 
It's not email, it's Envelope. It's also not `$network['title']`, it's `$network['name']`. I had thought I fixed that before committing it but I guess I didn't - sorry about that!

Issue #731, opened 2018-08-27
Google Maps: should use geocoding only in backend (and coordinates in frontend)


vollyimnetz:
Hi there,
today severeal of my customers informed me that a layer is visible above there google-maps areas.
As i investigate the problem i noticed that your plugin might be causing the problem.

It seems that the plugin use the geocoding api everytime it is loaded. Meaning that the plugin requests the coordinates for the address everytime it renders the map, instead of doing this once in the backend and then use the coordinates in the frontend.
Can you confirm that behaviour?

As the geocoding api has a way lower contingent than the normal javascript api, this leads to the problem of the overlay once the contingent is exceeded.

AlexGStapleton:
Hi Vollyimetz,

Thank you for your report. This is @braamgenis department so I can't really help with this, but out of interest, can you please show me a screenshot of the overlay you're referring too?

AlexGStapleton:
That's odd. I can't seem to replicate that overlay. Would it be possible for you to send me an export for the pictured widgets? To do this please open up the linked page in the editor and click the Layout button in the Page Builder toolbar. Then click import/export and then click download. Please upload the export to a 3rd party file hosting site such as <a href="https://upload.teknik.io/">teknik.io</a>.

vollyimnetz:
Nevermind - i think I found the problem elsewhere. It seems anyone uses our api-key to generate trafic on the google api - so the quota exceeded.

Misplon:
Awesome, glad to hear you were able to find the problem. All the best.

vollyimnetz:
Sorry, i have to reopen this.
Turns out, that this plugin indeed calls serveral hundert times the GeocodeService and by this exceeds my quota.

![image](https://user-images.githubusercontent.com/6512765/44714292-3e8c8880-aab5-11e8-9e4a-9c2210a98278.png)

vollyimnetz:
So the question remains: Why use the geocode service in the frontend? Even if its not hang in a loop - its totally unnessesary. Do this in a backend and you dont get any problems.

AlexGStapleton:
Can you please try #735 to see if that helps?

AlexGStapleton:
Let me know if you need help with testing the above PR and I'll provide further instructions.

braamgenis:
[...]

PR #735, opened 2018-08-28
Google Maps - limit geocode rate limit attempts

AlexGStapleton:
Related to #731. Don't consider merge until feedback from @vollyimnetz.

### DIFF
--- a/js/sow.google-map.js
+++ b/js/sow.google-map.js
@@ -348,7 +348,7 @@ sowb.SiteOriginGoogleMap = function($) {
 			var locationPromise = new $.Deferred();
 			var location = { address: inputLocation };
 			//check if address is actually a valid latlng
-			var latLng;
+			var latLng, geocodetimer;
 			if ( inputLocation && inputLocation.indexOf( ',' ) > -1 ) {
 				var vals = inputLocation.split( ',' );
 				// A latlng value should be of the format 'lat,lng'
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
This change will ensure the timeout only runs once - in other words, it will ensure there's only one more additional attempt at loading the rest of the markers. Clearing the timeout before it had its chance to run would result in the code not doing anything. Also, clearing would occur pretty much instantly unless I add another timeout as the code will do another run through pretty much instantly.

I've run a few tests with the above changes and there are no longer any additional requests for me. Can you please try the PR to confirm if this is (or isn't ) the case for you?

vollyimnetz:
a sorry - i mistook timeout for interval.

vollyimnetz:
As far as i can see, the problem is fixed with this code (i replaced the minified file in plugin).

js/sow.google-map.js
```diff
@@ -378,9 +380,16 @@ sowb.SiteOriginGoogleMap = function($) {
...
+								if ( geocodeIteration <= 2 ) {
+									geocodeIteration++;
+									this.getGeocoder().geocode.call( this, location, onGeocodeResults );
+								} else {
+									clearInterval( geocodeTimer );
+								}
+							}.bind( this ), 100 );
+						}
 					} else if ( status === google.maps.GeocoderStatus.ZERO_RESULTS ) {
 						locationPromise.reject( status );
```
AlexGStapleton: @braamgenis Does this stop processing the rest of markers? If so, maybe it's best to do that as if the API is over the query limit and nothing can process there's really no point in proceeding.

js/sow.google-map.js
```diff
@@ -378,9 +380,14 @@ sowb.SiteOriginGoogleMap = function($) {
 						locationPromise.resolve( results[ 0 ].geometry.location );
 					} else if ( status === google.maps.GeocoderStatus.OVER_QUERY_LIMIT ) {
 						//try again please
-						if ( geocodetimer != null ) {
-							geocodetimer = setTimeout( function () {
-								this.getGeocoder().geocode.call( this, location, onGeocodeResults );
+						if ( geocodeTimer == null ) {
+							geocodeTimer = setInterval( function () {
```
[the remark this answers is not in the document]
AlexGStapleton: I attempted that initially but the problem is that it ran way too fast. I know it's not ideal, but it might be best if you attempt that as I feel I may have been overlooking something obvious.

js/sow.google-map.js
braamgenis: Sorry, I forgot to reply to this comment. This is just for the current marker, so requests for other markers will still occur. That's a good suggestion, but it will require a larger refactor to stop all other requests if we hit the `OVER_QUERY_LIMIT` status, which should perhaps be part of the work to move geocoding requests to the server-side. I'll try look into this soon.

PR #741, opened 2018-09-04
Widget CSS: Use inline CSS as a fallback if siteorigin-widgets directory or CSS file wasn't able to be created

AlexGStapleton:
[Slack](https://siteorigin.slack.com/archives/C02SYBBH2/p1535963127000100)

### DIFF
--- a/base/siteorigin-widget.class.php
+++ b/base/siteorigin-widget.class.php
@@ -731,7 +731,7 @@ abstract class SiteOrigin_Widget extends WP_Widget {
 			if ( WP_Filesystem() ) {
 				global $wp_filesystem;
 				$upload_dir = wp_upload_dir();
-						$add_cache = true;
+				$add_cache = true;
 
 				if ( ! $wp_filesystem->is_dir( $upload_dir['basedir'] . '/siteorigin-widgets/' ) ) {
 					$directory_created = $wp_filesystem->mkdir( $upload_dir['basedir'] . '/siteorigin-widgets/' );
--- a/base/siteorigin-widget.class.php
+++ b/base/siteorigin-widget.class.php
@@ -731,22 +731,26 @@ abstract class SiteOrigin_Widget extends WP_Widget {
 			if ( WP_Filesystem() ) {
 				global $wp_filesystem;
 				$upload_dir = wp_upload_dir();
-
-				if ( ! $wp_filesystem->is_dir( $upload_dir['basedir'] . '/siteorigin-widgets/' ) ) {
-					$store_css = $wp_filesystem->mkdir( $upload_dir['basedir'] . '/siteorigin-widgets/' );
+				
+				$dir_exists = $wp_filesystem->is_dir( $upload_dir['basedir'] . '/siteorigin-widgets/' );
+				
+				if ( empty( $dir_exists ) ) {
+					// The 'siteorigin-widgets' directory doesn't exist, so try to create it.
+					$dir_exists = $wp_filesystem->mkdir( $upload_dir['basedir'] . '/siteorigin-widgets/' );
 				}
 				
-				if ( empty( $store_css ) ) {
+				if ( ! empty( $dir_exists ) ) {
+					// The 'siteorigin-widgets' directory exists, so we can try to write the CSS to a file.
 					$wp_filesystem->delete( $upload_dir['basedir'] . '/siteorigin-widgets/' . $name );
-					$store_css = $wp_filesystem->put_contents(
+					$file_put_success = $wp_filesystem->put_contents(
 						$upload_dir['basedir'] . '/siteorigin-widgets/' . $name,
 						$css
 					);
 				}
 			}
 
-			// Check if we're able to store CSS (based on the above checks) and if we are, store a record of it
-			if ( empty( $store_css ) ) {
+			// We couldn't write to file, so let's use cache instead.
+			if ( empty( $file_put_success ) ) {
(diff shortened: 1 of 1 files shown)


base/siteorigin-widget.class.php
```diff
@@ -727,23 +727,34 @@ private function save_css( $instance ){
 
 		$css = $this->get_instance_css($instance);
 
-		if( !empty($css) ) {
-
+		if ( !empty( $css ) ) {
 			if ( WP_Filesystem() ) {
 				global $wp_filesystem;
 				$upload_dir = wp_upload_dir();
+				$add_cache = true;
```
gregpriday: Is this variable necessary? I can't see a situation where this is anything but true?

AlexGStapleton: Whoops - it was for debugging.

base/siteorigin-widget.class.php
```diff
@@ -727,23 +727,33 @@ private function save_css( $instance ){
...
 			if ( WP_Filesystem() ) {
 				global $wp_filesystem;
 				$upload_dir = wp_upload_dir();
 
 				if ( ! $wp_filesystem->is_dir( $upload_dir['basedir'] . '/siteorigin-widgets/' ) ) {
-					$wp_filesystem->mkdir( $upload_dir['basedir'] . '/siteorigin-widgets/' );
+					$directory_created = $wp_filesystem->mkdir( $upload_dir['basedir'] . '/siteorigin-widgets/' );
+					if ( ! $directory_created ) {
+						$add_cache = true;
+					}
```
braamgenis: This could be simplified to avoid all the conditionals here and below. Just assign the result of the various file write attempts to the `$add_cache` variable and check the result at the end. I.e.
`$add_cache = ! $wp_filesystem->mkdir( $upload_dir['basedir'] . '/siteorigin-widgets/' );`
and
`$add_cache = ! $wp_filesystem->put_contents(...`

base/siteorigin-widget.class.php
```diff
@@ -727,23 +727,33 @@ private function save_css( $instance ){
...
+					$file_created = $wp_filesystem->put_contents(
+						$upload_dir['basedir'] . '/siteorigin-widgets/' . $name,
+						$css
+					);
+					if ( ! $file_created ) {
+						$add_cache = true;
+					}
+				}
 			} else {
+				$add_cache = true;
```
braamgenis: This whole `else` block can be removed.

base/siteorigin-widget.class.php
AlexGStapleton: Oh whoops. For some reason, I had thought this was causing errors to be logged. I just double checked the initial thread and I was mistaken.

base/siteorigin-widget.class.php
```diff
@@ -727,23 +727,25 @@ private function save_css( $instance ){
...
-		if( !empty($css) ) {
-
+		if ( ! empty( $css ) ) {
 			if ( WP_Filesystem() ) {
 				global $wp_filesystem;
 				$upload_dir = wp_upload_dir();
 
 				if ( ! $wp_filesystem->is_dir( $upload_dir['basedir'] . '/siteorigin-widgets/' ) ) {
-					$wp_filesystem->mkdir( $upload_dir['basedir'] . '/siteorigin-widgets/' );
+					$add_cache = $wp_filesystem->mkdir( $upload_dir['basedir'] . '/siteorigin-widgets/' );
```
braamgenis: `$wp_filesystem->mkdir` returns `false` if it fails, so this should be `$add_cache = ! $wp_filesystem->mkdir(...` as we only want to use the cache if we fail to write the file, correct? Alternatively, if you want to avoid the negation, you should rename `$add_cache`.

base/siteorigin-widget.class.php
```diff
@@ -727,23 +727,25 @@ private function save_css( $instance ){
...
 				}
 
-				$wp_filesystem->delete( $upload_dir['basedir'] . '/siteorigin-widgets/' . $name );
-				$wp_filesystem->put_contents(
-					$upload_dir['basedir'] . '/siteorigin-widgets/' . $name,
-					$css
-				);
+				if ( !isset( $add_cache ) ) {
```
braamgenis: Assuming you make the changes suggested above, `$add_cache` will be set to `false` if the call to `mkdir` succeeded, then this check should rather be `if ( empty( $add_cache ) )`.

base/siteorigin-widget.class.php
```diff
@@ -727,23 +727,26 @@ private function save_css( $instance ){
...
 			if ( WP_Filesystem() ) {
 				global $wp_filesystem;
 				$upload_dir = wp_upload_dir();
 
 				if ( ! $wp_filesystem->is_dir( $upload_dir['basedir'] . '/siteorigin-widgets/' ) ) {
-					$wp_filesystem->mkdir( $upload_dir['basedir'] . '/siteorigin-widgets/' );
+					$store_css = $wp_filesystem->mkdir( $upload_dir['basedir'] . '/siteorigin-widgets/' );
 				}
+				
+				if ( empty( $store_css ) ) {
```
braamgenis: The logic needs a bit more tweaking. This will skip trying to write the file if the `siteorigin-widgets` directory already existed and it will try to write the file if the attempt to create the directory failed.

AlexGStapleton:
That code looks good and works for all of the use cases I tried.

PR #742, opened 2018-09-04
Fixed Accordion deeplinking issue

AlexGStapleton:
Resolves siteorigin/siteorigin-premium#213

### DIFF
--- a/widgets/accordion/js/accordion.js
+++ b/widgets/accordion/js/accordion.js
@@ -33,6 +33,10 @@ jQuery( function ( $ ) {
 					);
 					$panel.addClass( 'sow-accordion-panel-open' );
 					openPanels.push( panel );
+					parentPanel = $( panel ).parents( '.sow-accordion-panel' );
+					if ( parentPanel.length && openPanels.indexOf( parentPanel.get(0) ) == -1 ) {
+						openPanels.push( parentPanel.get(0) );
+					}
 					if ( ! preventHashChange ) {
 						updateHash();
 					}


braamgenis:
Just did a bit of testing for functionality and there's some weird behavior. I've attached the layout: [so-premium-accordion.zip](https://github.com/siteorigin/so-widgets-bundle/files/2394417/so-premium-accordion.zip)

Steps to reproduce:
1. When you initially view the page everything seems fine.
2. Click on any item in the child accordion.
3. From this point on the state of both accordions seems a bit scrambled. Clicking on any headers in parent or child accordion has mixed results.

AlexGStapleton:
Can you show me a video of what you're seeing? I ask as I can't seem to replicate this with the provided layout. I do notice that the ids appear to be added a tad too quickly so it can result in both of them temporarily being present in the address bar. [Here's a quick video](https://doc-0c-4g-docs.googleusercontent.com/docs/securesc/b090cql7fqrrvb6ljitnujvtlmh9ijlg/l3539340gmlcjl03sbpffmsli8n9cngp/1537336800000/15522570231455504901/15522570231455504901/1KnhwHWaumD0AxQIsj-UTVV16yiGNCcC8?h=02211706589536887182&nonce=1b9d910c81ev2&user=15522570231455504901&hash=5gh46f2hhj8d7ni6dudjvnn8ftl0f8gn).

braamgenis:
I don't have enough bandwidth right now to do videos, but some steps to reproduce from a new page are:
1. Add an accordion widget.
2. Add a panel and set it's initial state to Open and set to _Layout builder_ and open it.
3. Add a child accordion widget
4. Add a panel to the child accordion widget, add some content and set it's initial state to Open.
5. In the child accordion set _Maximum number of simultaneous open panels_ to 1
6. Update and view the page.
7. Try close and open the child accordion panel.

That's about as much as I can simplify it to reproduce the issue I'm seeing.

AlexGStapleton:
Oh. Funny story, the Accordion Layout Builders aren't being included with the export - I'm not too sure why that's happening

Taking a look at this now.

braamgenis:
The panels seem to be behaving themselves when opening and closing now.

Could you perhaps give an overview of how this is generally supposed to work?  I don't see the anchors changing in the URL when clicking on child accordion items, which would be necessary to be able to deep link to a child item.

AlexGStapleton:
Did you enable **Use anchor tags in the URL** for the child accordion widget? I ask as it's working for me so I'm not too sure why it's not working for you.

Here's a reference video: https://drive.google.com/a/siteorigin.com/uc?id=1ymvcae0riKyyTUdCBL4vJpSPrz_MgQtZ

braamgenis:
Oh, wow. 🤦‍♂️ That facepalm emoji isn't facepalmy enough. Yeah, didn't have that enabled for the child. Sorry for the false alarm.

Ok, so it has the anchors in the URL now, but seeing another issue. Seems like if you have **Maximum number of simultaneous open panels** set to 1 in both the parent accordion and the child, it works for the child, but not the parent.

widgets/accordion/js/accordion.js
```diff
@@ -33,6 +33,10 @@ jQuery( function ( $ ) {
 					);
 					$panel.addClass( 'sow-accordion-panel-open' );
 					openPanels.push( panel );
+					parentPanel = $( panel ).parents( '.sow-accordion-panel' );
+					if ( parentPanel.length && openPanels.indexOf( parentPanel.get(0) ) == -1 ) {
+						openPanels.push( parentPanel.get(0) );
```
braamgenis: I don't think we should be mixing parent panels in with child panels here. Seems like this could cause the max open panels issue I'm seeing with parent accordion. How are you supposed to tell the parent and child panels apart? What if they have the same names? Just thinking that could also cause confusion in the URL, if anchors are the same for parent and child.

widgets/accordion/js/accordion.js
```diff
@@ -65,7 +69,26 @@ jQuery( function ( $ ) {
...
+
+							if ( skippedPanels != maxOpenPanels) {
+								skippedPanels++;
+								return true;
+							} else {
+								closePanel( openPanels[ openPanels.indexOf( el ) ] );
+							}
+						} );
+					} else {
+						closePanel( 0 );
```
braamgenis: This might actually be the real culprit for the max simultaneous panels issue. `closePanel` takes the HTML element as an argument.

braamgenis:
Ok, that's working. Now the anchors in the URL aren't clearing for the child accordion when the parent accordion panel closes.

AlexGStapleton:
How does that look?

widgets/accordion/js/accordion.js
```diff
@@ -64,18 +69,34 @@ jQuery( function ( $ ) {
...
 					var anchors = [];
-					for ( var i = 0; i < openPanels.length; i++ ) {
-						var anchor = $( openPanels[ i ] ).data( 'anchor' );
+					allOpenPanels = $( '.sow-accordion-panel-open' ).toArray();
+					for ( var i = 0; i < allOpenPanels.length; i++ ) {
+						var anchor = $( allOpenPanels[ i ] ).data( 'anchor' );
 						if ( anchor ) {
-							anchors[ i ] = anchor;
+							$parentPanel = $( allOpenPanels[ i ] ).parents( '.sow-accordion-panel' );
+							if ( $parentPanel.length ) {
```
braamgenis: Both code paths are doing the same thing, so this can just be a single condition. I.e. something like 
`if( ! $parentPanel.length || ( $parentPanel.length && $parentPanel.hasClass( 'sow-accordion-panel-open' ) ) )`

PR #744, opened 2018-09-09
Slider: don't hide video on mobile

AlexGStapleton:
This CSS was added when autoplaying on mobile just wasn't possible so it was better just to up and hide the video.

@Misplon Can you please try this on iOS? I can confirm this is working on Android (for at least devices where autoplaying videos are possible)

### DIFF
--- a/base/inc/widgets/base-slider.class.php
+++ b/base/inc/widgets/base-slider.class.php
@@ -100,7 +100,7 @@ abstract class SiteOrigin_Widget_Base_Slider extends SiteOrigin_Widget {
 
 			'background_video_mobile' => array(
 				'type' => 'checkbox',
-				'label' => __( 'Show Slide Background Videos on mobile', 'so-widgets-bundle' ),
+				'label' => __( 'Show slide background videos on mobile', 'so-widgets-bundle' ),
 				'description' => __( 'Ticking this setting will allow for slide backgorund videos to appear on devices that support autoplaying videos on mobile.', 'so-widgets-bundle' ),
 			)
 		);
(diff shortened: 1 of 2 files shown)


Misplon:
This is working correctly in iOS Chrome and Safari. Will be a nice improvement to get out.

braamgenis:
I agree that we should make this possible, but it should be optional. Mobile users are often concerned about data usage, so we should at least add a checkbox which defaults to the current behavior.

Misplon:
I think that would work for all of the users that reach out about this and all of the users that want everything to stay as is. @AlexGStapleton, keen? Can you roll that out for us?

base/inc/widgets/base-slider.class.php
```diff
@@ -96,6 +96,12 @@ function control_form_fields(){
 				'label' => __( 'Swipe Control', 'so-widgets-bundle' ),
 				'description' => __( 'Allow users to swipe through frames on mobile devices.', 'so-widgets-bundle' ),
 				'default' => true,
+			),
+
+			'background_video_mobile' => array(
+				'type' => 'checkbox',
+				'label' => __( 'Show Slide Background Videos on mobile', 'so-widgets-bundle' ),
```
braamgenis: Should either be sentence case (_Show slide background videos on mobile_) or title case, but not a mix of the two. IMO sentence case is better.

base/inc/widgets/base-slider.class.php
```diff
@@ -271,7 +277,10 @@ function render_frame( $i, $frame ){
 			<?php
 			$this->render_frame_contents( $i, $frame );
 			if( !empty( $background['videos'] ) ) {
-				$this->video_code( $background['videos'], array('sow-' . $background['video-sizing'] . '-element') );
+
+				$classes = array( 'sow-' . $background['video-sizing'] . '-element' );
+				$classes[] = ! empty( $controls['background_video_mobile'] ) ? 'sow-mobile-video_enabled' : '';
```
braamgenis: Wrather wrap this in an `if` block to avoid outputting an empty class, i.e.
```
if ( ! empty( $controls['background_video_mobile'] ) ) {
    $classes[] = 'sow-mobile-video_enabled';
}
```

braamgenis: _Wrather wrap_... Wow, that's a special kind of typo. 😂

PR #747, opened 2018-09-19
Remove slider slide number from output

AlexGStapleton:
This PR removes the slide numbers from the slider output. I can't see any use for these numbers but I do see a valid use for removing them - they appear in the automatic WordPress excerpts.

### DIFF
--- a/base/inc/widgets/base-slider.class.php
+++ b/base/inc/widgets/base-slider.class.php
@@ -179,7 +179,7 @@ abstract class SiteOrigin_Widget_Base_Slider extends SiteOrigin_Widget {
 				?>
 				<ol class="sow-slider-pagination">
 					<?php foreach($frames as $i => $frame) : ?>
-						<li><a href="#" data-goto="<?php echo $i ?>" aria-label="<?php printf( __( 'display slide %s', 'so-widgets-bundle' ), $i+1 ) ?>"><?php echo $i+1 ?></a></li>
+						<li><a href="#" data-goto="<?php echo $i ?>" aria-label="<?php printf( __( 'display slide %s', 'so-widgets-bundle' ), $i+1 ) ?>"></a></li>
 					<?php endforeach; ?>
 				</ol>
 


Misplon:
@braamgenis We're working through an open support topic on this at the moment. It would be awesome to have fixed in the next (not upcoming) WB release. Thanks!

braamgenis:
The numbers are there for accessibility. It's arguably the correct behavior to include the numbers in excerpts, but if it's absolutely necessary, isn't there a filter for excerpts we could use to remove the numbers?

Misplon:
Thanks, I’ll look for a filter next time I’m on.

(From a normal user perspective, when viewed in an excerpt, the numbers
don’t make sense. The excerpt reads “1 2 Text begins”. At the moment, given
what I understand, I’m unable to explain to users why those numbers should
be there. The slider isn’t present, usually just the featured image
followed by text.)

braamgenis:
Yeah, that's not ideal. :finnadie:

So I'd guess there's some filtering going on already if the slider isn't present. Ideally we'd use that same action/filter to remove the navigation numbers.

Misplon:
Sweet. I'm sure @AlexGStapleton can take a look for us on Monday and report back.

Misplon:
@braamgenis Would you be ok with proceeding with Alex's original fix here in the PR?

AlexGStapleton:
It's not ideal, but siteorigin/siteorigin-premium#221 could work if we do wish to retain this by default.

Misplon:
@AlexGStapleton Please, can you ask @gregpriday how he feels about https://github.com/siteorigin/siteorigin-premium/issues/221 being free or premium. Thanks.

AlexGStapleton:
Done. :)

braamgenis:
Ok, so if I followed correctly, we're going with siteorigin/siteorigin-premium#221 as the solution for this issue?

braamgenis:
Issue moved to siteorigin/siteorigin-panels#581

Misplon:
Sounds great, thanks guys.

PR #748, opened 2018-09-20
Image widget - if no alt text is supplied add an empty alt attribute

reefnetalex:
If no alt text is supplied for an image widget add an empty alt
attribute (alt="") for WCAG compliance. Nearly all images should have an
alt attribute to meet WCAG 2.1 Level A.

See

https://www.w3.org/WAI/WCAG21/Techniques/failures/F38

"This describes a failure condition for text alternatives for images
that should be ignored by AT. If there is no alt attribute at all
assistive technologies are not able to ignore the non-text content. The
alt attribute must be provided and have a null value (i.e., alt="" ) to
avoid a failure of this Success Criterion."

and

https://webaim.org/techniques/alttext/

"Every image must have an alt attribute. This is a requirement of HTML
standard (with perhaps a few exceptions in HTML5). Images without an alt
attribute are likely inaccessible. In some cases, images may be given an
empty or null alt attribute (e.g., alt="")."

Prevents image widget from failing accessibility audit through lack of
alt attribute.


AlexGStapleton:
@braamgenis This looks fine to me and I agree with the change based on the provided accessibility sources.

PR #750, opened 2018-10-02
Tabs: Allow tabs to change tabs

AlexGStapleton:
This PR will allow the user to tab through the tabs. This PR is for accessibility reasons.

### DIFF
--- a/widgets/tabs/js/tabs.js
+++ b/widgets/tabs/js/tabs.js
@@ -33,9 +33,11 @@ jQuery( function ( $ ) {
 					var $prevTab = $tabs.filter( '.sow-tabs-tab-selected' );
 					$prevTab.removeClass( 'sow-tabs-tab-selected' );
 					var prevTabIndex = $prevTab.index();
+					$tabPanels.eq( prevTabIndex ).attr( 'aria-hidden', 'true' );
 					$tabPanels.eq( prevTabIndex ).fadeOut( 'fast',
 						function () {
 							$( this ).trigger( 'hide' );
+							$tabPanels.eq( selectedIndex ).removeAttr( 'aria-hidden' );
 							$tabPanels.eq( selectedIndex ).fadeIn( 'fast',
 								function () {
 									$( this ).trigger( 'show' );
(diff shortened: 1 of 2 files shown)


widgets/tabs/js/tabs.js
```diff
@@ -52,7 +52,7 @@ jQuery( function ( $ ) {
 				}
 			};
 			
-			$tabs.click( function () {
+			$tabs.on('keyup click', function () {
 				selectTab( this );
```
braamgenis: This automatically selects the tab as soon as the Tab button is released, which isn't usual Tab button behavior. Pressing Tab should cycle focus to the tab and then it should only be selected when the Enter key is pressed.

braamgenis: This automatically selects the tab as soon as the Tab button is released, which isn't usual Tab button behavior. Pressing Tab should cycle focus to the tab and then it should only be selected when the Enter key is pressed.

AlexGStapleton: [It looks like both methods are accepted](https://www.w3.org/TR/wai-aria-practices/#tabpanel), but displaying the contents automatically upon focus is the general recommendation if there's no latency delay during the tab change (as in no requests to the server in general) ([source](https://www.w3.org/TR/wai-aria-practices/#kbd_selection_follows_focus))

Regardless, there are a few other recommendations (such as the tabindex and keyboard control) that should be made.

braamgenis: Oh, that's interesting. In my general experience tab components use manual activation, at least, I can't recall seeing one use automatic activation. Anyway, all good then. 👍 

>Regardless, there are a few other recommendations (such as the tabindex and keyboard control) that should be made.

Not sure what you mean by this last bit? Recommendations? Was there something else you still wanted to change or add?

AlexGStapleton: The tabindex needs to only be set on the active tab and a tabindex must be set for the tab content.

You navigate through the tabs using arrows.

I would try [the example tab](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html) for how the tab is ideally meant to work. Thoughts?

braamgenis: Ah, ok. Yeah, I agree, let's go with the recommended functionality for better accessibility. 👍

AlexGStapleton:
How does that look?

widgets/tabs/tpl/default.php
```diff
@@ -27,7 +28,7 @@
 	<div class="sow-tabs-panel-container">
 	<?php foreach ( $tabs as $i => $tab ) : ?>
 		<div class="sow-tabs-panel">
-			<div class="sow-tabs-panel-content">
+			<div class="sow-tabs-panel-content" role="tabpanel" <?php echo $i != $initial_tab_index ? 'aria-hidden="true"' : ''; ?>>
```
braamgenis: Doesn't this also need `tabindex=0` when selected? Or perhaps the parent container `sow-tabs-panel-container`?

braamgenis:
All good! 👍

PR #752, opened 2018-10-09
Post Selector: Add option to adjust taxonomies relationship

AlexGStapleton:
This PR will add a new setting to the posts form field that will allow users to change the taxonomies relationship. Previously, it was only possible to adjust the taxonomies relationship filtering the [siteorigin_widgets_posts_selector_query](https://github.com/siteorigin/so-widgets-bundle/blob/1.13.0/base/inc/post-selector.php#L106) hook and adjusting the value using PHP.

### DIFF
--- a/base/inc/fields/posts.class.php
+++ b/base/inc/fields/posts.class.php
@@ -38,6 +38,16 @@ class SiteOrigin_Widget_Field_Posts extends SiteOrigin_Widget_Field_Container_Ba
 				'description' => __( 'Taxonomies are groups such as categories, tags, posts and products.', 'so-widgets-bundle' ),
 			),
 
+			'tax_query_relation' => array(
+				'type' => 'radio',
+				'label' => __( 'Taxonomies Relationship', 'so-widgets-bundle' ),
+				'options' => array(
+					'OR' => __( 'OR', 'so-widgets-bundle' ),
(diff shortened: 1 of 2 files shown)


braamgenis:
Was there an issue related to or a specific request for this? I was just looking at the docs [here](https://codex.wordpress.org/Class_Reference/WP_Query#Taxonomy_Parameters) and it seems we could allow for more complex queries and keep things simple at the same time. IMO we should rather support the AND and OR relations directly in the field. I.e. the _Taxonomies_ field would allow simple entries like

>_OR(category:quotes,post_format:post-format-quote)_

which would be equivalent to what we have currently as

>_category:quotes,post_format:post-format-quote_

and we could allow more complex nested queries like 

>_OR(category:quotes,AND(post_format:post-format-quote,category:wisdom))_

This is just the first idea I came up with, there might be a nicer way to it, but the general idea is to not add another field and just allow a slightly more complex syntax for advanced users.

AlexGStapleton:
This is a request that happens once or twice a month.

The more advanced syntax is a good idea. No one has directly asked for that but I can definitely understand why someone may want it.

Misplon:
Guys, I'm going to close this for the moment. I've logged the "more advanced syntax" route.

PR #753, opened 2018-10-16
Contact: Altered required message and indicator. Resolve #567

### LINKED ISSUE #567: Contact: Don't Default To and From Email to the Same Email
When a new contact form is added, we need to prevent the To and From email from being the same (which will happen if the user has the site email set to their email also) as it'll be definitely be flagged as spam.

I think we should change the from email to wordpress@domain.com in instances like this. if the user doesn't like it, they can adjust it as desired anyway.

AlexGStapleton:
- Resolves #567
- Don't output required text message if field is empty
- Wrap required indicator in span with class to allow for easier styling

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -867,7 +867,7 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 
 		$indicate_required_fields = $instance['settings']['required_field_indicator'];
 
-		if ( ! empty( $indicate_required_fields ) ) {
+		if ( ! empty( $indicate_required_fields ) && ! empty( $instance['settings']['required_field_indicator_message'] ) ) {
 			?>
             <p><em><?php echo esc_html( $instance['settings']['required_field_indicator_message'] ) ?></em></p>
 			<?php
@@ -891,9 +891,6 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
             <div class="sow-form-field sow-form-field-<?php echo sanitize_html_class( $field['type'] ) ?>"><?php
 
 			$label = $field['label'];
-			if ( $indicate_required_fields && ! empty( $field['required']['required'] ) ) {
-				$label .= '*';
-			}
 			$is_text_input_field = ( $field['type'] != 'select' && $field['type'] != 'radio' && $field['type'] != 'checkboxes' );
 			// label should be rendered before the field, then CSS will do the exact positioning.
 			$render_label_before_field = ( $label_position != 'below' && $label_position != 'inside' ) || ( $label_position == 'inside' && ! $is_text_input_field );
@@ -902,6 +899,9 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 			}
 
 			$show_placeholder = $label_position == 'inside';
+			if ( $show_placeholder && $indicate_required_fields && ! empty( $field['required']['required'] ) ) {
+				$label .= '*';
+			}
 
 			if ( ! empty( $errors[ $field_name ] ) ) {
 				?>
@@ -956,7 +956,7 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 			}
 			?><label<?php if ( ! empty( $label_class ) ) {
 				echo $label_class;
-			} ?> for="<?php echo esc_attr( $field_id ) ?>"><strong><?php echo esc_html( $label ) ?></strong></label>
+			} ?> for="<?php echo esc_attr( $field_id ) ?>"><strong><?php echo esc_html( $label ) ?><span class="sow-form-required">*</span></strong></label>
 			<?php
 		}
 	}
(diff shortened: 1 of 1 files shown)


widgets/contact/contact.php
```diff
@@ -956,7 +956,7 @@ function render_form_label( $field_id, $label, $position ) {
 			}
 			?><label<?php if ( ! empty( $label_class ) ) {
 				echo $label_class;
-			} ?> for="<?php echo esc_attr( $field_id ) ?>"><strong><?php echo esc_html( $label ) ?></strong></label>
+			} ?> for="<?php echo esc_attr( $field_id ) ?>"><strong><?php echo esc_html( $label ) ?><span class="sow-form-required">*</span></strong></label>
```
braamgenis: This is now always displaying an asterisk for required fields.

widgets/contact/contact.php
```diff
@@ -1227,6 +1227,12 @@ function send_mail( $email_fields, $instance ) {
 			$instance['settings']['from'] = get_option( 'admin_email' );
 		}
 
+		if ( $instance['settings']['from'] == $instance['settings']['from'] ) {
```
braamgenis:
```suggestion
		if ( $instance['settings']['from'] == $instance['settings']['to'] ) {
```

widgets/contact/contact.php
```diff
@@ -1227,6 +1227,12 @@ function send_mail( $email_fields, $instance ) {
...
+			$domain = parse_url( esc_url( get_site_url() ), PHP_URL_HOST );
+			$domain = str_replace( 'www.', '', $domain );
+			$instance['settings']['from'] = "wordpress@$domain";
```
braamgenis: We shouldn't do this here. We should rather check this when we're setting up the default values of the form, so that users have the opportunity to see and change it before any emails are sent.

AlexGStapleton: To clarify, are you suggesting that form validation should be implemented or after form submission?

widgets/contact/contact.php
```diff
@@ -948,15 +949,20 @@ function render_form_fields( $fields, $errors = array(), $instance ) {
...
 			$label_class = '';
 			if ( ! empty( $position ) ) {
 				$label_class = ' class="sow-form-field-label-' . $position . '"';
 			}
-			?><label<?php if ( ! empty( $label_class ) ) {
-				echo $label_class;
-			} ?> for="<?php echo esc_attr( $field_id ) ?>"><strong><?php echo esc_html( $label ) ?><span class="sow-form-required">*</span></strong></label>
+
+			if ( $required ) {
+				$required = '<span class="sow-form-required">*</span>';
```
braamgenis: Don't hijack the `$required` variable. Rather declare a new one for the output, maybe `$required_indicator`.

widgets/contact/contact.php
braamgenis: Hehehe, nope. Form validation would be nice, but that's a pretty big addition for another day. 😜 

I mean when the form is rendered we populate some fields, e.g. the from and to fields, with defaults. At that point we should be checking whether `from` is the same `to` and if it is, set the `from` field to wordpress@domain.com.

widgets/contact/contact.php
```diff
@@ -948,15 +949,20 @@ function render_form_fields( $fields, $errors = array(), $instance ) {
...
-			?><label<?php if ( ! empty( $label_class ) ) {
-				echo $label_class;
-			} ?> for="<?php echo esc_attr( $field_id ) ?>"><strong><?php echo esc_html( $label ) ?></strong></label>
+
+			if ( $required ) {
+				$required = '<span class="sow-form-required">*</span>';
+			}
+			?>
+			<label<?php echo ! empty( $label_class ) ? $label_class : ''; ?> for="<?php echo esc_attr( $field_id ) ?>">
+				<strong><?php echo esc_html( $label ) . $required; ?></strong>
```
braamgenis: Or just keep it all inline:
```suggestion
				<strong><?php echo esc_html( $label ) . ( $required ? '<span class="sow-form-required">*</span>' : '' ) ; ?></strong>
```

widgets/contact/contact.php
braamgenis: So just after this:
https://github.com/siteorigin/so-widgets-bundle/blob/af811014ff18290d304f9d9eb36f4f4359251e0c/widgets/contact/contact.php#L686-L692

AlexGStapleton: Ah. Should the following be moved also?

https://github.com/siteorigin/so-widgets-bundle/blob/af811014ff18290d304f9d9eb36f4f4359251e0c/widgets/contact/contact.php#L1226-L1234

braamgenis: Hmm... yeah, but it seems we should have some level of redundancy between here and the `modify_instance` function, where the defaults are set. So perhaps extract all the related email field value sanitization we're doing into a function, which will then be called both here and in `modify_instance`.

widgets/contact/contact.php
```diff
@@ -682,14 +683,8 @@ function sanitize_multiple_emails( $value ) {
 	}
 
 	function modify_instance( $instance ) {
-		// Use this to set up an initial version of the
-		if ( empty( $instance['settings']['to'] ) ) {
-			$current_user               = wp_get_current_user();
-			$instance['settings']['to'] = $current_user->user_email;
```
AlexGStapleton: @braamgenis I removed this (it now uses `default_domain_email` instead) as it's unlikely this email won't match up with the site admin email and it will prevent running an additional check for something that will likely be changed anyway. Should I revert this change?

widgets/contact/contact.php
```diff
@@ -853,6 +848,28 @@ static function name_from_label( $label, & $ids ) {
 		return $id;
 	}
 
+	function email_validation( $instance ) {
```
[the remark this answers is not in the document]
braamgenis: `sanitize_emails` is fine. It actually describes what the function does, i.e. it checks that values are "sane" and if not, it makes them "sane". Comments are good, but in an ideal world, function names describe exactly what they do, but also need to be as succinct as possible. We don't want ridiculously long function names.

Also there's no need for concern about conflicts with `sanitize_email`, as this is in it's own class. This should really be a `private` function as it's only for internal use in this class.

PR #760, opened 2018-10-22
Features: Remove flex from each feature on mobile

AlexGStapleton:
This PR will remove flex from each feature on mobile to prevent text from being (horizontally) cut off on mobile. This change will only apply when the user has the features widget to be enabled.

Before:

After:

[Test layout](https://drive.google.com/a/siteorigin.com/uc?id=1bQAhYEq1fkgmOCIqgqbNzZiTM98awPUJ).

### DIFF
--- a/widgets/features/styles/default.less
+++ b/widgets/features/styles/default.less
@@ -192,6 +192,7 @@
                 width: 100% !important;
                 float: none;
                 margin-bottom: 40px;
+                display: block;
 
                 &:last-child {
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
@braamgenis Just pinging you about this PR. I've found this comes up at least once or twice every other week.

braamgenis:
@AlexGStapleton Cool. It's on my list, will get to it azap.

Issue #313, opened 2016-08-30
Google Maps: Add Mobile Zoom setting


AlexGStapleton:
A semi-common complaint about the Google Maps widget is that the zoom the user is using on desktop doesn't look good on mobile due to it typically being very zoomed in. As a result, the user typically has to use a comprise setting of a good looking mobile zoom and a good looking desktop zoom. We should add a Mobile Zoom setting to prevent users from having to comprise with such an important aspect of the map.

rtibarewala:
I wish to know the same?
Need a different zoom level for mobile urgently.
It would be great to have this feature inbuilt in SiteOrigin Maps Widget.

Issue #174, opened 2016-04-11
Maps: Default to a height if no height is set


AlexGStapleton:
When no height is set, the Map will still be output it just won't be visible. This isn't exactly clear so to some it might seem as though the widget isn't working. It might be a good idea to "default" to a height to avoid confusion.

braamgenis:
The maps widget height field has a default value of 480px. So a user would have to clear it intentionally for it to be blank. It would be nice if we could replace that field with the measurement field to allow any units for height, but I think it's somewhat dependent on the Google Maps JS API.

AlexGStapleton:
Yeah, that's the weird thing. A user did clear it and was wondering why it wasn't working. :confused: I think we should at least enforce a height if the field is empty.

PR #761, opened 2018-10-23
Resolve Google Maps: #690 #313 #174

### LINKED ISSUE #690: Google Maps: Add option to ctrl + scroll to zoom (gestureHandling)
A few users have recently requested the ability to enable `ctrl + zoom`. As per [this](https://developers.google.com/maps/documentation/javascript/interaction), this can be done by setting `gestureHandling` to `cooperative`. As this is a simple change, we should add this as an optional setting.

AlexGStapleton:
This PR resolves Resolve #690 #313 and #174.

Regarding #690, [Draggable](https://developers.google.com/maps/documentation/javascript/reference/map#MapOptions.draggable) has been deprecated in favour of [gestureHandling](https://developers.google.com/maps/documentation/javascript/reference/map#MapOptions.gestureHandling). [scrollwheel](https://developers.google.com/maps/documentation/javascript/reference/map#MapOptions.scrollwheel) is no longer recommended in favour of [gestureHandling](https://developers.google.com/maps/documentation/javascript/reference/map#MapOptions.gestureHandling). They don't want to to mix and match these settings.


widgets/google-map/google-map.php
```diff
@@ -89,7 +89,6 @@ function get_widget_form(){
 					),
 					'height'      => array(
 						'type'    => 'text',
-						'default' => 480,
```
braamgenis: Why remove this?

js/sow.google-map.js
```diff
@@ -16,13 +16,15 @@ sowb.SiteOriginGoogleMap = function($) {
 			var zoom = Number(options.zoom);
 
 			if ( !zoom ) zoom = 14;
+			if( window.innerWidth < 768 ) {
```
braamgenis: Mobile width shouldn't be hardcoded here.

widgets/google-map/google-map.php
```diff
@@ -122,6 +121,21 @@ function get_widget_form(){
 						'integer'     => true,
 
 					),
+
+					// Mobile Zoom is hidden for static maps as there's no reliable way to check for mobile devices that will bypass caching
```
braamgenis: Couldn't we check for this in the JS and adjust the static image's URL accordingly?

widgets/google-map/google-map.php
```diff
@@ -505,11 +531,20 @@ function get_template_variables( $instance, $args ) {
 				}
 			}
 
+			// Work out what we need to set gestureHandling to
+			if ( ( ! $settings['draggable'] && ! $settings['scroll_zoom'] ) || ! $settings['draggable'] ) {
```
braamgenis: I just checked this very quickly, so could be wrong, but it seems the output of this is ignoring `$settings['scroll_zoom']`.
I.e. it could be simplified to just `! $settings['draggable']`

widgets/google-map/tpl/static-map.php
```diff
@@ -14,7 +14,8 @@
 <img
 	class="sowb-google-map-static"
 	border="0"
-	src="<?php echo sow_esc_url( $src_url ) ?>"
+	src="<?php echo sow_esc_url( $src_url['desktop'] ) ?>"
+	data-src="<?php echo ! empty( $src_url['mobile'] ) ? esc_attr( json_encode( $src_url ) ) : ''; ?>"
```
AlexGStapleton: @braamgenis Should I pre-escape this in google-maps.php or should I do it here?

AlexGStapleton:
Oh... sorry, I missed your comment about commits. Sorry about that! I'll do that from now on.

widgets/google-map/tpl/static-map.php
braamgenis: The general convention we follow is to try and do escaping as late as possible, so here is good.

braamgenis:
Just checked out this branch and opened a page with an existing Maps Widget and got the following error:
```
Notice | Undefined index: settings
do_action('admin_print_scripts-post.php')
wp-admin/admin-header.php:118
SiteOrigin_Panels_Admin->enqueue_admin_scripts()
wp-includes/class-wp-hook.php:286
SiteOrigin_Widget->form()
wp-content/plugins/siteorigin-panels/inc/admin.php:458
SiteOrigin_Widget_GoogleMap_Widget->modify_instance()
wp-content/plugins/so-widgets-bundle/base/siteorigin-widget.class.php:439
wp-content/plugins/so-widgets-bundle/widgets/google-map/google-map.php:466
```

AlexGStapleton:
I can't seem to replicate that error. I tried three other SiteOrgin Google Maps widgets I had set up locally and tried [the Dentist Prebuilt layout](http://layouts.siteorigin.com/layout/dentist/). :(

Can you please send me an export of the page that this occurred on?

braamgenis:
Just checked again and this is happening while editing any page. Basically the Maps widget just needs to be activated in Plugins > SiteOrigin Widgets, so that the `modify_instance` method is called while the admin page is being loaded. It's on PHP 7.1 and the error shows up in Query Monitor and the debug log.

AlexGStapleton:
That's how the development site is setup and unfortunately, I can't replicate this error message. :/

widgets/google-map/google-map.php
```diff
@@ -441,10 +450,34 @@ function get_settings_form() {
...
+				'type'        => 'number',
+				'label'       => __( 'Responsive breakpoint', 'so-widgets-bundle' ),
+				'default'     => '780',
+				'description' => __( 'This setting controls when the map will use the mobile zoom. This breakpoint will only be used if a mobile zoom is set in the SiteOrigin Google Maps settings. The default value is 780px', 'so-widgets-bundle' )
 			)
 		);
 	}
 
+	function modify_instance( $instance ) {
+		if ( empty( $instance['settings']['mobile_zoom'] ) ) {
```
AlexGStapleton: @braamgenis I'm unsure why this check would trigger the error you're getting. 

This same check is done with the contact form and that doesn't have an error... wait, do you get an error message with that widget?

https://github.com/siteorigin/so-widgets-bundle/blob/1.13.3/widgets/contact/contact.php#L686

braamgenis: `$instance` can be an empty `array`. E.g. there are times when we run widgets' `form` functions with an empty `array` simply to enqueue their scripts.

AlexGStapleton: For future reference, should this be done with `modify_instance` in general?

AlexGStapleton:
How does that look?

widgets/google-map/google-map.php
```diff
@@ -462,17 +464,19 @@ function get_settings_form() {
 	}
 
 	function modify_instance( $instance ) {
-		if ( empty( $instance['settings']['mobile_zoom'] ) ) {
-			// Check if a zoom is set, and if it is, set the mobile zoom to that
-			if ( empty( $instance['settings']['zoom'] ) ) {
-				$instance['settings']['mobile_zoom'] = 12;
-			} else {
-				$instance['settings']['mobile_zoom'] = $instance['settings']['zoom'];
+		if ( ! empty( $instance ) ) {
```
braamgenis: This works, but it would feel "safer" to me to use `if ( ! empty( $instance['settings'] )` as it's possible for `$instance` to be a non-empty array, but not have a `settings` key which would present the same issue.

braamgenis:
Well this is a tad awkward, but I just checked the Static Maps API and it seems there are a few things we should have tried before adding a separate mobile zoom setting. Firstly, there's a `mobile=true` parameter which sounds like it should adjust tiles accordingly. Secondly, they suggest setting the [scale](https://developers.google.com/maps/documentation/maps-static/dev-guide#scale_values) parameter to 2 for mobile devices. @AlexGStapleton Could you give that a try and see what effect it has? It would simplify this code a lot and reduce potential issues later.

widgets/google-map/google-map.php
```diff
@@ -551,6 +611,7 @@ public function enqueue_widget_scripts( $instance ) {
 					'geocode' => array(
 						'noResults' => __( 'There were no results for the place you entered. Please try another.', 'so-widgets-bundle' ),
 					),
+					'breakpoint' => $breakpoint,
```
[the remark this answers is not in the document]
AlexGStapleton: The breakpoint is purely being used in JavaScript. I can't seem to find anywhere `get_template_variables` has been used to pass data specifically for use with JavaScript. Did you want me to add a data attribute to the templates to pass the breakpoint?

AlexGStapleton:
Damn. I didn't see that parameter in the docs previously. I can't seem to find mobile so I'm not too sure what that specifically does.

widgets/google-map/google-map.php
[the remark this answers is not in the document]
braamgenis: Check the `get_template_variables` function in this very widget. For interactive maps we have the `map_data` variable which is specifically for use in JS and for static maps we pass the fallback image as a data attribute, so you can just add the breakpoint variable as another data attribute or combine with the existing data attribute.

braamgenis:
Hmmm... not sure where I saw the `mobile=true` option and can't find it now. Perhaps test it out to see what effect it has?

AlexGStapleton:
Just merged the recent changes to the Google Maps widget to this PR. Let me know if there were any other changes you would like me to do make.

braamgenis:
GitHub is still saying there are conflicts. Probably the last few changes before releasing. Please merge them and then I'll check if there's anything else.

AlexGStapleton:
Oh sorry, I forgot about the pending changes made in 1.15.2 or I would have held off on the merge commit. Checking now.

braamgenis:
Seems the merge went awry. Getting fatals in this branch now.

AlexGStapleton:
Ah crap, sorry about that!

widgets/google-map/google-map.php
```diff
@@ -455,6 +478,14 @@ function get_template_variables( $instance, $args ) {
 
 		$styles = $this->get_styles( $instance );
 
+		$global_settings = $this->get_global_settings();
+		if ( empty( $instance['api_key_section']['api_key'] ) ) {
+			if ( ! empty( $global_settings['api_key'] ) ) {
+				$instance['api_key_section']['api_key'] = $global_settings['api_key'];
+			}
+		}
```
braamgenis: This was removed in the develop branch, so it seems it was accidentally reintroduced in the merge.

widgets/google-map/google-map.php
```diff
@@ -492,13 +525,23 @@ function get_template_variables( $instance, $args ) {
...
-			
+      
+      // Work out what we need to set gestureHandling to
+			if ( ! $settings['draggable'] ) {
+				$gestureHandling = 'none';
+			} elseif ( ! $settings['scroll_zoom'] ) {
+				$gestureHandling = 'cooperative';
+			} else {
+				$gestureHandling = 'greedy';
+			}
```
braamgenis: Looking at this again with fresh eyes it occurred to me that with `draggable` and `scrollwheel` being deprecated we should migrate those to a single `gestureHandling` radio field in the form to stay consistent with the Maps API. This code is effectively migrating those settings anyway.

widgets/google-map/google-map.php
AlexGStapleton: How does that look? Should I add something similar to `get_api_key` for the breakpoint?

widgets/google-map/google-map.php
AlexGStapleton: That's an interesting idea. How do these options look:

- Enabled (greedy)
- Dynamically change based on device (auto)
- ctrl key or ⌘ click to scroll (cooperative)
- Disabled (none)

The problem with this change is that it makes it quite tricky to add an inline description without it being really large. Hm. Actually, we could just link to the documentation. Something as "For information on what these settings do, click here.". It's not ideal but it doesn't matter how long the documentation is.

widgets/google-map/google-map.php
braamgenis: Did you make a change? It looks the same to me. Perhaps you still need to push.
Just to clarify, what I meant was this is supposed to be removed, as it's not being used anymore. I'm not sure how the API key relates to the breakpoint setting.

widgets/google-map/google-map.php
braamgenis: Yeah, I think we can label the options exactly as they are in the Google Maps docs [here](https://developers.google.com/maps/documentation/javascript/interaction#gestureHandling) e.g: 

- Greedy
- Cooperative
- None
- Auto

and then that description looks good _"For information on what these settings do, click here."_ to link to the Google Maps docs, as an explanation to users.

AlexGStapleton:
How does that look?

widgets/google-map/google-map.php
```diff
@@ -750,6 +746,15 @@ public function modify_instance( $instance ) {
...
 			}
 
+			// Migrate draggable and scroll_zoom to gesture_handling
+			if ( isset( $instance['settings']['draggable'] ) && ! $instance['settings']['draggable'] ) {
+				$instance['settings']['gesture_handling'] = 'none';
+			} elseif ( isset( $instance['settings']['scroll_zoom'] ) && ! $instance['settings']['scroll_zoom'] ) {
+				$instance['settings']['gesture_handling'] = 'cooperative';
+			} else {
+				$instance['settings']['gesture_handling'] = 'greedy';
+			}
```
braamgenis: When migrating settings, to keep things clean we should `unset` the old settings that will no longer be used. I.e.
```
	unset( $instance['settings']['draggable'] );
	unset( $instance['settings']['scroll_zoom'] );
```

braamgenis:
Cool, just need to remove the old settings and then if you've tested everything and are happy that it's working as expected, we can merge. 👍

AlexGStapleton:
Done. Just tried 5.3 also and I'm not experiencing any issues.

PR #765, opened 2018-10-30
Move SiteOrigin Editor TinyMCE Enqueue to widget

AlexGStapleton:
[Thread](https://siteorigin.com/thread/toolbar-has-white-icons/)

While SiteOrigin Widgets Bundle is activated, active buttons are white rather than black. This PR changes this.

### DIFF
--- a/widgets/editor/editor.php
+++ b/widgets/editor/editor.php
@@ -22,6 +22,8 @@ class SiteOrigin_Widget_Editor_Widget extends SiteOrigin_Widget {
 			false,
 			plugin_dir_path(__FILE__)
 		);
+
+		add_action( 'siteorigin_widgets_enqueue_admin_scripts_sow-editor', array( $this, 'enqueue_tinymce' ) );
 	}
 
 	function get_widget_form() {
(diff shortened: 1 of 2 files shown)


widgets/editor/editor.php
```diff
@@ -115,6 +117,10 @@ function get_style_name($instance) {
 		// We're not using a style
 		return false;
 	}
+
+	function enqueue_tinymce() {
+		wp_enqueue_style( 'editor-buttons' );
```
braamgenis: I did a quick check and it looks like this isn't necessary at all anymore.

AlexGStapleton: That's odd. I initially attempted removing just [this code](https://github.com/siteorigin/so-widgets-bundle/pull/765/commits/d80902c3304db783330ff9dd997247b44774904c#diff-42c27934512039924cc8e53dcad763cdL249) from base/base.php without adding the highlighted code but I received SiteOrigin Page Builder generic error message stating that the widget styles and/or JavaScript weren't able to loaded (I can't remember the exact error message). I can't seem to replicate that error again, however.

AlexGStapleton: It might be okay to revert the editor.php changes, but I feel it's safer to include it anyway. If you're not worried about it then I'll revert those changes so the only adjustment is to base.php.

braamgenis: This code was first introduce in 2015 before the core Text Widget was an instance of TinyMCE editor and it seems the editor button styles are now being enqueued properly by WordPress since that change. I prefer not to have redundant code around, so I think we should remove it and in the unlikely event that it was somehow useful, we can add it back with a decent comment stating the exact purpose.

PR #777, opened 2018-11-12
SiteOrigin Widgets block: Fix Layout Slider not being editable

AlexGStapleton:
This PR removes the `is_admin()` check from the Layout Slider widget. This check prevents the form from displaying in the SiteOrigin Widgets block. During testing, I wasn't able to find any issues with removing it. With that said, if you have a replacement you would like to be used, let me know and I'll add that.

### DIFF
--- a/widgets/layout-slider/layout-slider.php
+++ b/widgets/layout-slider/layout-slider.php
@@ -192,7 +192,7 @@ class SiteOrigin_Widget_LayoutSlider_Widget extends SiteOrigin_Widget_Base_Slide
 	}
 
 	function form( $instance, $form_type = 'widget' ) {
-		if( is_admin() && defined('SITEORIGIN_PANELS_VERSION') ) {
+		if( defined('SITEORIGIN_PANELS_VERSION') ) {
 			parent::form( $instance, $form_type );
 		} else {
 			?>
--- a/widgets/layout-slider/layout-slider.php
+++ b/widgets/layout-slider/layout-slider.php
@@ -192,7 +192,7 @@ class SiteOrigin_Widget_LayoutSlider_Widget extends SiteOrigin_Widget_Base_Slide
 	}
 
 	function form( $instance, $form_type = 'widget' ) {
-		if( defined('SITEORIGIN_PANELS_VERSION') ) {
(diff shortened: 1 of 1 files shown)


widgets/layout-slider/layout-slider.php
```diff
@@ -192,7 +192,7 @@ function get_widget_form(){
 	}
 
 	function form( $instance, $form_type = 'widget' ) {
-		if( is_admin() && defined('SITEORIGIN_PANELS_VERSION') ) {
```
braamgenis: This was added to prevent outputting the _Open Builder_ button in some contexts where it wouldn't work. Specifically it was in the front-end editor of some page builders, which lacked some of the scripts and styles required for the builder field to work.

AlexGStapleton: Ah. I wasn't aware of that. What would you recommend using to detect Gutenberg? I ask as I had a look around and I can't find anywhere, where we're specifically checking for Gutenberg, is active on the current page.

braamgenis: As far as I can tell at the moment it's necessary to use two checks to stay compatible with both the Gutenberg plugin block editor and the integrated block editor in WP 5. See:
https://github.com/siteorigin/siteorigin-panels/blob/82627ae22b38f5ce831ca36d77de67b54e2b06ff/compat/gutenberg-block.php#L32-L38

AlexGStapleton: Annoyingly, that doesn't work as `is_current_screen` doesn't exist at the time. For whatever reason, it looks like WordPress hates conditionals now - a simple `is_guttenberg()` would be really convenient right about now. :/

braamgenis: Ah, right. For a moment I forgot this is in the block editor context. 😕

widgets/layout-slider/layout-slider.php
```diff
@@ -192,7 +192,7 @@ function get_widget_form(){
...
+		if( defined('SITEORIGIN_PANELS_VERSION') ) {
```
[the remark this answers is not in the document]
braamgenis: Ok, so here's how I handled the same thing for Premium Accordion and Tabs:
https://github.com/siteorigin/siteorigin-premium/commit/23579a7f6405edfc59f9fa7599f71020009a43d4

I think just having `defined('REST_REQUEST' )` is enough. No need for the additional `&& REST_REQUEST`. Then checking whether `register_block_type` function is defined indicates the block editor context is present.

PR #786, opened 2018-12-04
Accordion: account for duplicate accordion titles

AlexGStapleton:
This PR prevents the need for unique titles. Previously, if you used the same title in more than one place, all instances of that title would opened upon clicking one. This PR ensues the anchors are unique by appending the array index id if the title isn't unique.

This PR also prevents an instance where a partial match title won't open cause other titles to open - opening the `Other Example` item won't open `Example`.

### DIFF
--- a/widgets/accordion/accordion.php
+++ b/widgets/accordion/accordion.php
@@ -176,7 +176,8 @@ class SiteOrigin_Widget_Accordion_Widget extends SiteOrigin_Widget {
 		if( empty( $instance ) ) return array();
 		
 		$panels = empty( $instance['panels'] ) ? array() : $instance['panels'];
-		
+
+		$anchor_list = array();
 		foreach ( $panels as $i => &$panel ) {
 			if ( empty( $panel['before_title'] ) ) {
 				$panel['before_title'] = '';
@@ -196,6 +197,12 @@ class SiteOrigin_Widget_Accordion_Widget extends SiteOrigin_Widget {
 			} else {
 				$panel['anchor'] = $panel['title'];
 			}
+
+			if ( isset( $anchor_list[ $panel['anchor'] ] ) ) {
+				$panel['anchor'] .= '-' . $i;
+
+			}
+			$anchor_list[ $panel['anchor'] ] = true;
 		}
 		
 		
--- a/widgets/accordion/accordion.php
+++ b/widgets/accordion/accordion.php
@@ -194,14 +194,13 @@ class SiteOrigin_Widget_Accordion_Widget extends SiteOrigin_Widget {
 					$id .= '-' . $args['widget_id'];
 				}
 				$panel['anchor'] = $id . '-' . $i;
+			} else if ( isset( $anchor_list[ strtolower( $panel['title'] ) ] ) ) {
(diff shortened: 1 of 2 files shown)


widgets/accordion/accordion.php
```diff
@@ -196,6 +197,12 @@ public function get_template_variables( $instance, $args ) {
 			} else {
 				$panel['anchor'] = $panel['title'];
 			}
+
+			if ( isset( $anchor_list[ $panel['anchor'] ] ) ) {
```
braamgenis: This can just be an `else if` statement in the above conditions. I.e.
```
			if ( empty( $panel['title'] ) ) {
			    // There isn't a title so create a unique string for the anchor.
			} else if ( isset( $anchor_list[ $panel['title'] ] ) ) {
			    // There's another anchor with the same title, so add the index here.
			} else {
				// Just set the title as the anchor.
			}
```

braamgenis: And please add a short comment, so we know what this is doing and why.

widgets/accordion/js/accordion.js
```diff
@@ -108,7 +108,8 @@ jQuery( function ( $ ) {
 					for ( var i = 0; i < panels.length; i++ ) {
 						var panel = panels[ i ];
 						var anchor = $( panel ).data( 'anchor' );
-						if ( anchor && window.location.hash.indexOf( anchor ) > -1 ) {
+						var windowHash = window.location.hash.substring(1).split( ',' );
```
braamgenis: The term 'hash' is quite overloaded, so rename this to something a bit more specific. Perhaps just `anchors`?

AlexGStapleton: Anchors seems a tad confusing due to `anchor` (which is a list of anchors) being present. Maybe `windowAnchors`?

braamgenis: `anchor` isn't a list, it's just one anchor name (a string). So `anchors` makes perfect sense for the array which contains many anchor names (an array of strings).

AlexGStapleton: Ah, my b. Didn't pick up on that.

widgets/accordion/accordion.php
```diff
@@ -193,9 +194,14 @@ public function get_template_variables( $instance, $args ) {
 					$id .= '-' . $args['widget_id'];
 				}
 				$panel['anchor'] = $id . '-' . $i;
+			} else if ( isset( $anchor_list[ strtolower( $panel['title'] ) ] ) ) {
```
braamgenis: This doesn't work when you have titles with uppercase letters, because here it's checking using `strtolower`, but the anchors are being added to `$anchor_list` below as is. Perhaps when you make changes just do a quick test again to make sure things are working as expected.

AlexGStapleton:
How does that look?

braamgenis:
Still some funky behaviour. Check the attached layout,
[wb-accordion.zip](https://github.com/siteorigin/so-widgets-bundle/files/2716646/wb-accordion.zip)

AlexGStapleton:
Ah. Didn't think of that use case.

How does that look?

braamgenis:
Looks good! 👍

PR #794, opened 2018-12-20
Tabs: Account for accented characters

AlexGStapleton:
This PR will fix an issue with Anchor Tags that prevented tabs with accented characters from opening automatically on load. For an example of this:

Set a Tab title to Le Réflexologie and then save.
Open the page, select the tab and then reload - the default tab will be opened.

This PR also updates updateSelectedTab to use, in my opinion, the more robust solution that the Accordion widget is using.

[Layout](https://drive.google.com/a/siteorigin.com/uc?id=1BDk8EszSfQCspYTuULcEIeWLPbuN3bd6)

### DIFF
--- a/widgets/tabs/js/tabs.js
+++ b/widgets/tabs/js/tabs.js
@@ -113,13 +113,15 @@ jQuery( function ( $ ) {
 			if ( useAnchorTags ) {
 				var updateSelectedTab = function () {
 					if ( window.location.hash ) {
-						var anchors = window.location.hash.replace( '#', '' ).split( ',' );
-						anchors.forEach( function ( anchor ) {
-							var tab = $tabs.filter( '[data-anchor="' + anchor + '"]' );
-							if ( tab ) {
-								selectTab( tab, true );
+						var tabs = $tabs.toArray();
+						for ( var i = 0; i < tabs.length; i++ ) {
+							var panel = tabs[ i ];
+							var anchor = $( panel ).data( 'anchor' );
+							var anchorHash = decodeURI( window.location.hash.replace( '#', '' ) ).split( ',' );
+							if ( anchor && $.inArray( decodeURI( anchor ), anchorHash ) > -1 ) {
+								selectTab( panel, true );
 							}
-						} );
+						}
 					}
 				};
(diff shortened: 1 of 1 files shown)


braamgenis:
I wasn't able to reproduce this issue. I tried the latest Chrome, Firefox and even Internet Explorer worked without issue. 😛 What browsers etc are you using?

AlexGStapleton:
Whoops! I prepared a layout for this and I'm about 90% sure I even uploaded it. Just didn't attach it. :(

[Here's the layout](https://drive.google.com/a/siteorigin.com/uc?id=1BDk8EszSfQCspYTuULcEIeWLPbuN3bd6).

I'll add it to the initial issue also.

braamgenis:
Still not seeing the issue with the provided layout in either Chrome or Firefox. :finnadie:
Regardless, I don't think it will hurt to apply `decodeURI` to the anchors. However, I don't think we should change the way the selected tab is being set unless there was a specific issue. The tabs are similar to accordion panels, but not exactly the same.

AlexGStapleton:
Weird stuff. Not too sure why you're unable to replicate it. :(

Outside of `decodeURI`, is there a specific alternative you would recommend? I know it's really hard to make a general recommendation as you're unable to replicate it, but in the hypothetical situation that you did, is there a different method you would prefer to be used? Or would you rather drop this for the time being? I would rather we didn't as another user and myself have replicated this but I understand.

braamgenis:
`decodeURI` is exactly the right thing to use here. My issue isn't with that, it's with the other changes. I think we should keep the change to use `decodeURI`, but revert the other changes, unless they're fixing a specific bug.

AlexGStapleton:
Ah. Right. Okay so the reason why I did this is due to:

> var tab = $tabs.filter( '[data-anchor="' + anchor + '"]' );

I could never encode/decode anchor in a manner that would allow for it to be detected. I agree the changes aren't ideal.

braamgenis:
the `filter` method also takes a function as an argument, so you could replace the above with:
```
var tab = $tabs.filter( function ( index, element ) {
	return decodeURI( anchor ) === decodeURI( $( element ).data( 'anchor' ) );
} );
```

PR #797, opened 2018-12-24
Tabs: Horizontal tabs

AlexGStapleton:


widgets/tabs/styles/default.less
```diff
@@ -116,19 +115,35 @@
...
+		display: flex;
+		flex-direction: row-reverse;
+
+		.sow-tabs-panel-container {
+			flex: 1;
+		}
+
+		.sow-tabs-tab {
+			display: block !important;
 		}
```
braamgenis: Some unnecessary duplication which can be removed. I.e. the only difference between `left` and `right` is `flex-direction: row-reverse;`

AlexGStapleton: Oddly, when I initially tried this it didn't work but I just tried it again and it did. 🤷‍♂️

widgets/tabs/styles/default.less
```diff
@@ -42,7 +43,24 @@
...
 		padding: @tabs_container_padding;
-		text-align: @tabs_container_tabs_align;
+
+		& when ( @tabs_container_tabs_position = left ), ( @tabs_container_tabs_position = right ) {
+			& when ( @tabs_container_tabs_align = middle ) {
+				display: flex;
+				justify-content: center;
+			    flex-direction: column;
+			}
+			& when ( @tabs_container_tabs_align = bottom ) {
```
braamgenis: Quite a bit of duplication here. Only difference between `middle` and `bottom` is the `justify-content` prop.

widgets/tabs/styles/default.less
```diff
@@ -113,4 +131,26 @@
 			}
 		}
 	}
+
+	& when ( @tabs_container_tabs_position = bottom ) {
+		& {
```
braamgenis: This can be removed.

widgets/tabs/styles/default.less
```diff
@@ -113,4 +131,26 @@
...
+			display: flex;
+		    flex-direction: column-reverse;
+		}
+	}
+
+	& when ( @tabs_container_tabs_position = left ), ( @tabs_container_tabs_position = right ) {
+		display: flex;
```
braamgenis: It seems like all options are using `display: flex;`, in which case it only needs to be set once.

AlexGStapleton: Top doesn't. Regardless, I've used `when not` to apply this when Top isn't being used.

PR #798, opened 2018-12-25
Widget Documentation Links

### LINKED ISSUE #678: Widgets Admin Page: Add documentation links
Should we add links in each widget block at `Plugins > SiteOrigin Widgets` like we do for SiteOrigin Premium addons? If so, is there a way for third-party plugins to do the same?

AlexGStapleton:
Resolves #678

Third party widgets can add links to their documentation by adding a documentation header.  A valid example of this is:
```
Documentation: https://siteorigin.com/widgets-bundle/example-widget/
```

### DIFF
--- a/so-widgets-bundle.php
+++ b/so-widgets-bundle.php
@@ -617,6 +617,7 @@ class SiteOrigin_Widgets_Bundle {
 			'AuthorURI' => 'Author URI',
 			'WidgetURI' => 'Widget URI',
 			'VideoURI' => 'Video URI',
+			'Documentation' => 'Documentation',
 		);
 
 		$widgets = array();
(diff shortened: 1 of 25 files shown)


so-widgets-bundle.php
```diff
@@ -617,6 +617,7 @@ function get_widgets_list(){
 			'AuthorURI' => 'Author URI',
 			'WidgetURI' => 'Widget URI',
 			'VideoURI' => 'Video URI',
+			'Documentation' => 'Documentation',
 		);
 
 		$widgets = array();
```
braamgenis: Not really related to this PR, but while this is fresh please add the following after the `get_file_data` call in the loop below to make the file headers translatable: 
```
// Need to explicitly call `translate` for file headers.
foreach ( array( 'Name', 'Description' ) as $field ) {
	$widget[ $field ] = translate( $widget[ $field ], 'so-widgets-bundle' );
}
```

PR #799, opened 2018-12-28
Feature testimonial widget

unknys:
Is testimonial and other new widgets stopped up from developing ? Livemesh, wpinked is developing fast into your markets. You always had the best widgets.

Will it come something new widgets soon ? Gutenberg dosent exacty has the "page builder" power like siteorigin, and i would love to see new widgets.


AlexGStapleton:
The SiteOrigin Widgets Bundle is a widget development framework that allows other developers (such as WPInked and Livemesh) to easily create widgets. Our main focus at the moment is on maintaining our core development framework to ensure no compatibility issues arise and they work well with the Block Editor. We've also introduced two new blocks that allow you to use our plugins (and by extension Livemesh SiteOrigin Widgets and Widgets for SiteOrigin plugins) with the Block Editor. Our blocks are:

The SiteOrgin Layout block is the full SiteOrigin Page Builder interface in block form
The SIteOrigin Widgets Block allows you to add widgets directly to the block editor.

This is not to suggest we won't introduce additional widgets in the future, it's just not what we're going working on.

Misplon:
Thanks for your support @unknys. As questions arise, please let us know on the forum: https://siteorigin.com/thread or email if you're using SiteOrigin Premium. We're always happy to lend a hand :)

Issue #808, opened 2019-01-08
custom icons issue in icon.class.php in get_value_parts()


andrejarh:
Hi I just noticed that custom icon packs are causing PHP warnings. They are trying to include a non-existing file due to an error in get_value_parts function of icon.class.php class.

The file which is trying to get included is /plugins/so-widget-bundle/icons/[custom icon pack]/filter.php

**Example of the warning:**
![screenshot 2019-01-08 at 11 12 25](https://user-images.githubusercontent.com/10053621/50825701-f40bfa00-1339-11e9-879e-c9924441bc72.png)

**How to reproduce the issue:**
1. Load a custom icon pack inside a theme/plugin using the siteorigin_widgets_icon_families filter as written in this tutorial: https://siteorigin.com/docs/widgets-bundle/form-building/icons-and-fonts/
2. Add define('WP_DEBUG', true); in the wp-config.php
3. Install Debug Bar plugin, to see all the warnings, notices on the fly.

**What causes the issue?**
I think this line of code (icon.class.php line 113):
```
if ( ! empty( $value_family ) ) {
	include_once plugin_dir_path(SOW_BUNDLE_BASE_FILE) . 'icons/' . $value_family . '/filter.php';
}
```

AlexGStapleton:
This appears to be caused by [this line](https://github.com/siteorigin/so-widgets-bundle/blob/1.14.0/icons/icons.php#L16). When I changed that line, this error no longer occured for me. Not too sure why it would be triggering something now but I'm not very familiar with the overall changes made.

Does changing the highlighted line to the following resolve this issue for you?

```
		if ( file_exists( plugin_dir_path(__FILE__) . $font . '/filter.php' ) ) {
			include_once plugin_dir_path(__FILE__) . $font . '/filter.php';
		}
```

andrejarh:
No,
but it does resolve if i change the line 113 in icon.class.php to
```
		if ( ! empty( $value_family ) ) {
			if ( file_exists( plugin_dir_path(SOW_BUNDLE_BASE_FILE) . 'icons/' . $value_family . '/filter.php' ) ) {
				include_once plugin_dir_path(SOW_BUNDLE_BASE_FILE) . 'icons/' . $value_family . '/filter.php';
			}
		}
```

AlexGStapleton:
Odd stuff. Would you be willing to [submit a PR](https://help.github.com/articles/about-pull-requests/) with that? If not, I can submit one.

[...]

PR #809, opened 2019-01-08
Font Field: Prevent warnings when including non-existing filter.php

andrejarh:
Linked to issue #808


braamgenis:
@andrejarh Thanks for submitting the PR. Your comment [here](https://github.com/siteorigin/so-widgets-bundle/issues/808#issuecomment-452285335) is quite right though, that it's not ideal to be hardcoding the path there. So I've made a slightly different quick fix which simply calls the `siteorigin_widgets_icon_families` filter to ensure the required filters for bundled icons are loaded, but avoids requiring that third party icons include a `filter.php` file. It's still not quite ideal, so at some point we'll have to relook at how loading icon families and their filters is handled.

PR #810, opened 2019-01-08
Update enqueue dependencies for the Block Editor

AlexGStapleton:
[Reported here](https://wordpress.org/support/topic/uncaught-typeerror-cannot-read-property-reduce-of-undefined/)

```
widget-block.min.js?ver=1.14.0:1 Uncaught TypeError: Cannot read property 'reduce' of undefined
    at widget-block.min.js?ver=1.14.0:1
    at widget-block.min.js?ver=1.14.0:1
```


braamgenis:
@AlexGStapleton Even though they're _very_ similar, this is actually using `lodash` which is a dependency of the block editor. Regardless, I think a better fix for this is to avoid using lodash/underscore altogether. The `reduce` function is available on plain JS arrays and well supported, so I've commited a change to just use that instead.

PR #815, opened 2019-01-14
Slider always show nav on mobile

### LINKED ISSUE #543: Sliders: Add option to always show slider navigation on mobile
On mobile devices, the slider navigation will only appear on touch. This can be problematic for certain layout sliders and Hero widgets as it may not be very clear that it works in this manner. As such, we should add the option of allowing for the Slider Navigation to always be visible on mobile.

This has come up a number of times in the past - typically with people stating that the navigation doesn't work.

AlexGStapleton:
Resolves #543. This adds an option for the Hero and Layout Sliders. The Slider widget doesn't currently have an option.

The Slider widget doesn't feature a responsive breakpoint setting. 
@braamgenis Do you want me to add one for this setting and if so, do you want me to add a dedicated Design (like the Hero and Layout Slider widgets) settings group or just add it to the base settings?


widgets/hero/styles/default.less
```diff
@@ -112,6 +113,14 @@
 		}
 	}
 
+	& when ( @nav_always_show_mobile = true ) {
+		@media (max-width: @responsive_breakpoint) {
+			& .sow-slider-pagination, & .sow-slide-nav {
+				display: block !important;
```
braamgenis: This should rather be implemented in the JS where the showing/hiding of navigation is currently being handled, instead of adding CSS with `!important` which is hard to change.

braamgenis:
As for the setting, I assume you mean the global widget setting? If so, then yeah, it makes sense to have a setting for the Image Slider too.

AlexGStapleton:
That's not exactly what I was asking (I was asking about a dedicated setting in the SiteOrigin Slider widget settings itself) but if you're completely okay with adding a global Responsive Breakpoint setting but without a dedicated setting in the widget itself, let's take that path.

braamgenis:
> I was asking about a dedicated setting in the SiteOrigin Slider widget settings itself
The Hero and Layout Slider widgets don't have a _Responsive Breakpoint_ setting in their forms, they both have a global setting for that. Not sure why we'd want to do this differently for the Image Slider?

AlexGStapleton:
The changes make the Slider widget compatible with this setting but it prevents users from changing the breakpoint when using the SiteOrigin Slider. The Hero and Layout Sliders have a dedicated breakpoint setting while the Slider doesn't, do you want to introduce a dedicated Breakpoint setting for the Slider widget just for this setting?

js/slider/jquery.slider.js
```diff
@@ -182,23 +182,28 @@ jQuery( function($){
...
+							.mouseleave(function(){
+								toHide = true;
+								setTimeout(function(){
+									if( toHide ) {
+										$p.add($n).clearQueue().fadeOut(150);
+									}
+									toHide = false;
+								}, 750);
+							});
+					} else if ( settings.nav_always_show_mobile && window.matchMedia('(max-width: ' + settings.breakpoint + ')').matches) {
```
braamgenis: Hmmm... so it seems I forgot that the Slider is already detecting mobile devices, by using `wp_is_mobile` on the back end. Does it makes to use both that and then have a breakpoint too? What if we have mobile devices which have more than 780px width?

AlexGStapleton: Ideally, `wp_is_mobile` shouldn't ever be used as it won't work when the user is using a cache plugin (and that's _very_ common). We should ideally remove `wp_is_mobile` in favor of using the breakpoint.

braamgenis: Ah, that's no good. 😕 My issue here is that this is trying to detect whether or not a device uses touch input, by screen size. For example, this won't work on a touch screen device with >780px width. It seems there's no reliable method for this though, so breakpoint will have to do. :finnadie:

widgets/hero/hero.php
```diff
@@ -477,6 +477,18 @@ function wrapper_data_filter( $data, $instance ) {
 		}
 		return $data;
 	}
+
+  function modify_instance( $instance ) {
+		if ( !empty( $instance['controls']['nav_always_show_mobile'] ) && $instance['controls']['nav_always_show_mobile'] ) {
+			$global_settings = $this->get_global_settings();
+
+			if ( ! empty( $global_settings['responsive_breakpoint'] ) ) {
+				$instance['controls']['breakpoint'] = $global_settings['responsive_breakpoint'];
```
braamgenis: There are a few problems with this:
* This exact same code is being duplicated in each of the widgets that extend the base slider widget class.
* It's a bit hacky and obscure to create a field on the `$instance` which might look like it comes from the controls section of the widget form, but doesn't. This would likely not be necessary if the duplicated code was in the base slider class.

You could perhaps retrieve the global setting for the breakpoint in the base slider's `slider_settings` function. No need to pass it through from the widget.

js/slider/jquery.slider.js
AlexGStapleton: Do you want to move away from `wp_is_mobile` or stick with it for the time being? It's not ideal for users using caching, but we haven't received many reports.

widgets/hero/hero.php
AlexGStapleton: That... that makes a lot more sense. Done.

AlexGStapleton:
How does that look? The last two commits move the breakpoint global settings to the slider widget and all of the `nav_always_show` settings.

js/slider/jquery.slider.js
braamgenis: Yeah, rather don't change what we're already using for now. I think we can relook at this later if it becomes a big issue (gets a lot of support requests).

braamgenis:
Cool, I think that's good. Did you check that it maintains the global settings from before the changes? If so, good to merge.

AlexGStapleton:
I didn't run into any issues with existing values after removing the global widget forms from the widget and adding it directly to the slider widget class. The altered values carried over without issue for me.

PR #818, opened 2019-01-21
Social Media Buttons - add Four networks

### LINKED ISSUE #689: Update FontAwesome and add additional Social Media options
A major update for FontAwesome has been released ([change log](https://github.com/FortAwesome/Font-Awesome/blob/master/CHANGELOG.md)) and it introduces a ton of new icons. There are additional icons for Social Media such as [Strava](https://www.strava.com/) so we should also work out what (if any) icons to add to SiteOrigin Social Media Buttons widget.

AlexGStapleton:
Add Bandcamp, Goodreads, Strava, and Telegram networks.

Resolve #689

### DIFF
--- a/widgets/social-media-buttons/data/networks.php
+++ b/widgets/social-media-buttons/data/networks.php
@@ -118,6 +118,12 @@ return array(
 		'icon_color' => '#FFFFFF',
 		'button_color' => '#205081'
 	),
+	'blogger-b'   => array(
+		'label'    => __( 'Blogger', 'so-widgets-bundle' ),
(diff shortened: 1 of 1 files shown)


widgets/social-media-buttons/data/networks.php
```diff
@@ -226,12 +238,24 @@
 		'icon_color' => '#FFFFFF',
 		'button_color' => '#171A21'
 	),
+	'Strava'   => array(
```
braamgenis: Lowercase please.

PR #823, opened 2019-02-04
Post carousel - load new posts based on possible number of visible posts

### LINKED ISSUE #133: Post Carousel Widget: Try improve next slide load time
To replicate users setup, ensure only 6 posts are added to a Carousel widget. Click the next arrow. Ideally, the 5th thumbnail should appear as soon as there is a space available. At the moment it takes three clicks of the right arrow to load the 5th slide, ideally, it would be one, as soon as there is a space it would load.

AlexGStapleton:
Resolve #133.

This will allow for faster posts loading as the carousel won't wait until only a single post is visible before attempting to load new posts.

### DIFF
--- a/widgets/post-carousel/js/carousel.js
+++ b/widgets/post-carousel/js/carousel.js
@@ -23,12 +23,20 @@ jQuery( function ( $ ) {
 				complete = numItems === totalPosts,
 				itemWidth = ( $firstItem.width() + parseInt( $firstItem.css( 'margin-right' ) ) ),
 				isRTL = $postsContainer.hasClass( 'js-rtl' ),
-				updateProp = isRTL ? 'margin-right' : 'margin-left';
+				updateProp = isRTL ? 'margin-right' : 'margin-left',
+				possibleVisibleItems = Math.ceil( $$.outerWidth() / itemWidth );
+
+				// Work out when to load new items based on number of posts on load and max number of visible items
+				// If numItems is less than 1, default to 1 to prevent issues with post loading
+				if ( numItems - possibleVisibleItems < 0 ) {
+					loadMoreTrigger = numItems - possibleVisibleItems;
+				} else {
+					loadMoreTrigger = 1;
+				}
 	
 			var updatePosition = function () {
 				if ( position < 0 ) position = 0;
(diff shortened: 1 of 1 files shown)


widgets/post-carousel/js/carousel.js
```diff
@@ -23,12 +23,20 @@ jQuery( function ( $ ) {
...
+				updateProp = isRTL ? 'margin-right' : 'margin-left',
+				possibleVisibleItems = Math.ceil( $$.outerWidth() / itemWidth );
+
+				// Work out when to load new items based on number of posts on load and max number of visible items
+				// If numItems is less than 1, default to 1 to prevent issues with post loading
+				if ( numItems - possibleVisibleItems < 0 ) {
+					loadMoreTrigger = numItems - possibleVisibleItems;
+				} else {
+					loadMoreTrigger = 1;
+				}
```
braamgenis: Currently this is fetching the next page of posts every time `updatePosition` is called. I think it's also a bit more complex than necessary as we already track the position of the carousel below.

widgets/post-carousel/js/carousel.js
```diff
@@ -23,12 +23,20 @@ jQuery( function ( $ ) {
...
 	
 			var updatePosition = function () {
 				if ( position < 0 ) position = 0;
-				if ( position >= $$.find( '.sow-carousel-item' ).length - 1 ) {
-					position = $$.find( '.sow-carousel-item' ).length - 1;
+				if ( position >= loadMoreTrigger ) {
```
braamgenis: The below code should be all that's necessary for this to work as intended. You just need to offset the position used to trigger the next page fetch by the number of visible items
```suggestion

				// Do this calc here to take possible resizing into account.
				var numVisibleItems = Math.ceil( $$.outerWidth() / itemWidth );
				// Offset position by numVisibleItems to trigger the next fetch before the view is empty.
				if ( position + numVisibleItems >= $$.find( '.sow-carousel-item' ).length - 1 ) {
```

widgets/post-carousel/js/carousel.js
AlexGStapleton: Yeah, you're right - I was over complicating this. 😞

Issue #666, opened 2018-03-12
Accordion: Add option to automatically add paragraphs


AlexGStapleton:
We should add an option to automatically add paragraphs to accordion items. For example, the following HTML:

```
Test

&nbsp;

Test

&nbsp;

Test

&nbsp;

Test
<h1>Test</h1>
Test
```

Will look like this on output:

![](https://i.imgur.com/4yrd2Cp.png)

Misplon:
@AlexGStapleton Please, can you chat with Braam and ask him if he's like to proceed with this or just get his thoughts.

braamgenis:
So we're already supposed to be doing this by default for TinyMCE fields when using the Visual editing tab, but it turns out there was a bug preventing it from happening. IMO it makes sense that we don't automatically do this for the Text (HTML) tab to try avoid surprising the user with unexpected results.

We added this option in the Editor widget to try stay aligned with how the core editor worked, but they since removed the option and my guess would be for similar reasons, i.e. switching between the Visual and Text tabs had unexpected results.

There is perhaps more to discuss around how we could go about adding the ability to do things like `wp_make_content_images_responsive`. Feel free to hit me up on Slack if/when you want to chat about it.

braamgenis:
So we're already supposed to be doing this by default for TinyMCE fields when using the Visual editing tab, but it turns out there was a bug preventing it from happening ([fixed now](https://github.com/siteorigin/so-widgets-bundle/commit/fc5a7eb8033eb127b99e623a36825dc1fab4bce1)). IMO it makes sense that we don't automatically do this for the Text (HTML) tab to try avoid surprising the user with unexpected results.

We added this option in the Editor widget to try stay aligned with how the core editor worked, but they since removed the option and my guess would be for similar reasons, i.e. switching between the Visual and Text tabs had unexpected results.

There is perhaps more to discuss around how we could go about adding the ability to do things like `wp_make_content_images_responsive`. Feel free to hit me up on Slack if/when you guys want to chat about it.

AlexGStapleton:
Ah. I wasn't aware of that issue. I'm completely fine with everything you've stated.
Should we instead just directly [wpautop](https://github.com/siteorigin/so-widgets-bundle/pull/734/files#diff-1296e8b8daa9763a58d7eef32b8322a4R228) against the accordion item content? I ask because without it new lines aren't rendered.

braamgenis:
[...]

PR #824, opened 2019-02-05
Accordion: Keyboard navigation

AlexGStapleton:
This PR is a version of #750 for the Accordion widget. It's built with WAI-ARIA in mind. Please refer to [this section](https://www.w3.org/TR/wai-aria-practices/#accordion) for information on accordions.

### DIFF
--- a/widgets/accordion/styles/default.less
+++ b/widgets/accordion/styles/default.less
@@ -57,6 +57,10 @@
 				color: @title_hover_color;
 				border-color: @heading_border_hover_color;
 			}
+			
+			&:focus {
+				outline: 1pox solid @heading_border_hover_color;
+			}
 
 			.sow-accordion-title {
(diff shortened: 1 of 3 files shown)


widgets/accordion/styles/default.less
```diff
@@ -57,6 +57,10 @@
 				color: @title_hover_color;
 				border-color: @heading_border_hover_color;
 			}
+			
+			&:focus {
+				outline: 1pox solid @heading_border_hover_color;
```
braamgenis: Typo. Should be `px`.

AlexGStapleton:
How does that look? I opted to use the typical outline of `#666` or the Chrome outline color because it was, in hindsight, a bad idea to use the other border colors as they won't stand out. I couldn't find one for FireFox or Safari.

PR #832, opened 2019-02-16
Tabs: Set default tab position to top

### LINKED ISSUE #831: Tabs Widget: Default alignment isn't working
To recreate the issue, deactivate the SiteOrigin Premium Tabs addon. If necessary, add a new Tabs widget.

**Result:**

tabs_ _siteorigin

**Cause**

The issue appears to be caused by `.sow-tabs` having `display: flex` added without anything further. I haven't worked on this widget or looked further. If `display: flex` needs to remain on that div then further declarations would be needed for the tabs container and the panel container.

AlexGStapleton:
Resolves #831

### DIFF
--- a/widgets/tabs/styles/default.less
+++ b/widgets/tabs/styles/default.less
@@ -9,7 +9,7 @@
 @tabs_container_border_radius: default;
 @tabs_container_padding: 12px 10px 0px 10px;
 @tabs_container_tabs_align: default;
-@tabs_container_tabs_position: default;
+@tabs_container_tabs_position: top;
 
 @tabs_background_color: default;
 @tabs_background_hover_color: default;


Misplon:
This doesn't appear to work if the Tabs addon in SO Premium is deactivated.

AlexGStapleton:
I can't seem to replicate this. SiteOrigin Premium is disabled for me and it's working as intended.

AlexGStapleton:
Did you save the widget after applying this change? The less file needs to be regenerated and that doesn't happen unless you save (or create a new tabs widget)?

Misplon:
Looks good now. I re-saved, without Premium, seems fine. I also did a build and tested it on https://demo.siteorigin.com/premium/tabs-addon/ without Premium. Seems fine.

PR #846, opened 2019-03-18
Google Maps: Detect draggable or scroll_zoom before running setting migration

AlexGStapleton:
This PR resolves an issue with the **draggable** and **scroll_zoom** setting migration code that results in it constantly running. This results in  the **Gesture Handling** being reset to **Greedy** upon viewing/opening the widget.

### DIFF
--- a/widgets/google-map/google-map.php
+++ b/widgets/google-map/google-map.php
@@ -758,17 +758,19 @@ class SiteOrigin_Widget_GoogleMap_Widget extends SiteOrigin_Widget {
 			}
 
 			// Migrate draggable and scroll_zoom to gesture_handling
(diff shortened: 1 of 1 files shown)


braamgenis:
@AlexGStapleton Could you be more specific in the PR description? There's no clear indication in the code what this is doing and there are no added comments. Seeing as you've already done the work of reproducing the issue, it should take almost no extra effort for you and then it removes the need for me to duplicate the work of figuring out how to reproduce so that I can verify that this does what it should. Unless there's an associated issue which already includes this info, that you could link to, it's generally ideal to include the following for all PRs:

- steps to reproduce
- expected behavior
- actual behavior

AlexGStapleton:
Sorry, I wasn't very clear. I've updated the PR description to be clearer.

PR #859, opened 2019-04-24
Correct name of SiteOrigin Widget Bundle object in admin.js

kallehauge:
If you go to a page/post and insert a layout/widget and press `preview` then it is possible to end up in a scenario where the block will return `This block has encountered an error and cannot be previewed.` because the `sowb` object do not exist.

Looking at other component like `widgets/accordion/js/accordion.js` then it defines the following at the top: `var sowb = window.sowb || {};`.

based on the way `sowb` is normally used, I'm assuming the object is meant to be named `sowbForms` in `admin.js`.

### DIFF
--- a/base/js/admin.js
+++ b/base/js/admin.js
@@ -1318,7 +1318,7 @@ var sowbForms = window.sowbForms || {};
 	if ( $body.hasClass('block-editor-page') ) {
 		// Setup new widgets when they're previewed in the block editor.
 		$(document).on('panels_setup_preview', function () {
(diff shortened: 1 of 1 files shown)


braamgenis:
Hi @kallehauge It is correctly using the `sowb` variable to trigger the `setup_widgets` event. It's a (admittedly, rather hacky) way we use to notify all widgets that they need to reinitialize themselves in some way.

Please feel free to [open a new issue](https://github.com/siteorigin/so-widgets-bundle/issues/new) with steps to reproduce the issue you're experiencing and we'll take a look as soon as we get a chance.

PR #865, opened 2019-05-05
Tabs and Accordion: Account for Accented Characters

AlexGStapleton:
This is a follow up to #794. This PR uses the provided feedback and also accounts for accented characters for the accordion widget also.

[Accordion Test Layout](https://drive.google.com/a/siteorigin.com/uc?id=1J7Avs2zold9rSt6032r0Kb61ZL1bkJwq)
[Tab Test Layout](https://drive.google.com/a/siteorigin.com/uc?id=1pggjX64KJ6LAIrohlCGD3w6222_i090a)

### DIFF
--- a/widgets/accordion/js/accordion.js
+++ b/widgets/accordion/js/accordion.js
@@ -137,16 +137,20 @@ jQuery( function ( $ ) {
 				};
 				
 				var updatePanelStates = function () {
-					var panels = $accordionPanels.toArray();
-					for ( var i = 0; i < panels.length; i++ ) {
-						var panel = panels[ i ];
-						var anchor = $( panel ).data( 'anchor' );
-						var anchors = window.location.hash.substring(1).split( ',' ); 
-						if ( anchor && $.inArray( anchor.toString(), anchors ) > -1 ) {
-							openPanel( panel, true );
-						} else {
-							closePanel( panel, true );
-						}
+					if ( window.location.hash ) {
+						var anchors = window.location.hash.replace( '#', '' ).split( ',' );
+						anchors.forEach( function ( anchor ) {
+							var $panels = $accordionPanels.toArray();
+							var panel = $panels.filter( function ( index, element ) {
+								return decodeURI( anchor ) === decodeURI( $( $panels[ element ] ).data( 'anchor' ) );
+							} );
+
+							if ( anchor ) {
+								openPanel( panel, true );
+							} else {
+								closePanel( panel, true );
+							}
+						} );
 					}
 				};
 				$( window ).on( 'hashchange', updatePanelStates );
--- a/widgets/accordion/js/accordion.js
+++ b/widgets/accordion/js/accordion.js
@@ -137,18 +137,21 @@ jQuery( function ( $ ) {
 				};
 				
 				var updatePanelStates = function () {
-					if ( window.location.hash ) {
-						var anchors = window.location.hash.replace( '#', '' ).split( ',' );
-						anchors.forEach( function ( anchor ) {
-							var $panels = $accordionPanels.toArray();
-							var panel = $panels.filter( function ( index, element ) {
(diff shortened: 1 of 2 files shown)


braamgenis:
I still can't reproduce, even with the layouts you provided. :finnadie:  I assume this has been reported again? If so, it would be great to see the thread for more info. Also, as [I mentioned](https://github.com/siteorigin/so-widgets-bundle/pull/794#issuecomment-450594946) in the previous PR, it should be fine to use `decodeURI` but let's not change anything else.

AlexGStapleton:
I'm not too sure why you're unable to replicate this.
How does that look?

widgets/tabs/js/tabs.js
```diff
@@ -121,7 +121,9 @@ jQuery( function ( $ ) {
 					if ( window.location.hash ) {
 						var anchors = window.location.hash.replace( '#', '' ).split( ',' );
 						anchors.forEach( function ( anchor ) {
-							var tab = $tabs.filter( '[data-anchor="' + anchor + '"]' );
+							var tab = $tabs.filter( function ( index, element ) {
+								return decodeURI( anchor ) === decodeURI( $( element ).data( 'anchor' ) );
+							} );
 							if ( tab ) {
```
braamgenis: The return value of the jQuery filter method is a jQuery object, so `tab` will always be truthy. A more accurate test here is perhaps `if ( tab.length > 0 )`.

widgets/accordion/js/accordion.js
```diff
@@ -142,7 +142,12 @@ jQuery( function ( $ ) {
 						var panel = panels[ i ];
 						var anchor = $( panel ).data( 'anchor' );
 						var anchors = window.location.hash.substring(1).split( ',' ); 
-						if ( anchor && $.inArray( anchor.toString(), anchors ) > -1 ) {
+						
+						var anchor = anchors.filter( function ( index, element ) {
+							return decodeURI( anchor.toString() ) === decodeURI( index );
```
braamgenis: This is decoding the integer index of the element and comparing it to the decoded `anchor` string . It should probably be something more like `return decodeURI( anchor.toString() ) === decodeURI( $( element ).data( 'anchor' ) );`.

widgets/accordion/js/accordion.js
```diff
@@ -142,7 +142,12 @@ jQuery( function ( $ ) {
...
+						} );
+
+						if ( anchor.length ) {
```
braamgenis:
```suggestion
						if ( panelOpen ) {
```

widgets/accordion/js/accordion.js
```diff
@@ -142,7 +142,12 @@ jQuery( function ( $ ) {
 						var panel = panels[ i ];
 						var anchor = $( panel ).data( 'anchor' );
```
braamgenis:
```suggestion
						var panelAnchor = $( panel ).data( 'anchor' );
```

widgets/accordion/js/accordion.js
```diff
@@ -142,7 +142,12 @@ jQuery( function ( $ ) {
...
 						var anchors = window.location.hash.substring(1).split( ',' ); 
-						if ( anchor && $.inArray( anchor.toString(), anchors ) > -1 ) {
+						
+						var anchor = anchors.filter( function ( index, element ) {
```
braamgenis:
```suggestion
						var panelOpen = anchors.some(function (anchor) {
```

widgets/accordion/js/accordion.js
braamgenis:
```suggestion
							return decodeURI(anchor) === decodeURI(panelAnchor);
```

widgets/tabs/js/tabs.js
braamgenis:
```suggestion
							if ( tab.length > 0 ) {
```

PR #872, opened 2019-05-24
Added padding to slider navigation arrows

### LINKED ISSUE #445: Hero Widget: Expand link area around nav arrows
For better usability it would be great to expand the link area around the nav arrows.

link-area

Let's apply to all applicable slider type widgets if we proceed.

Misplon:
This PR adds padding around the slider navigation arrows to hopefully, make them a little easier to select.

@braamgenis Are you happy with the selector the padding has been added to and the negative margin to maintain the arrow position? Should we rather use ems?

Which CSS property order convention are we using? [Outside In](https://webdesign.tutsplus.com/articles/outside-in-ordering-css-properties-by-importance--cms-21685)?

Resolves #445.

### DIFF
--- a/css/slider/slider.less
+++ b/css/slider/slider.less
@@ -159,7 +159,6 @@
 			display: block;
 			
 			padding: 15px;
-			position: relative;
 	
 			text-align: center;
 			text-decoration: none;
--- a/css/slider/slider.less
+++ b/css/slider/slider.less
@@ -159,6 +159,8 @@
 			display: block;
 
 			height: 1em;
+			margin: -15px;
+			padding: 15px;
 
 			text-align: center;
 			text-decoration: none;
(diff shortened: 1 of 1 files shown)


css/slider/slider.less
```diff
@@ -159,6 +159,8 @@
 			display: block;
 
 			height: 1em;
+			margin: -15px;
```
braamgenis: Yeah, I'm generally not a fan of using negative margin values. Rather change the position of the containing `div` below. I.e. in the `.sow-slide-nav-next` selector set `right: 5px;` and for `.sow-slide-nav-prev` set `left: 5px;`

braamgenis:
>Which CSS property order convention are we using?
We aren't using any convention that I'm aware of.

css/slider/slider.less
Misplon: Super, thanks for pushing for a better solution.

Misplon:
Super, thanks for checking :)

PR #890, opened 2019-08-14
Accordion: Check if if anchor is present in URL and scroll to Accordion item

### LINKED ISSUE #785: Accordion: Override Initially scroll to panel if anchor is present in URL
[Slack](https://siteorigin.slack.com/archives/C02SXRL8N/p1543390584001100)

The *Initially scroll to panel* setting should be overridden by the  *Use anchor tags in URL settings* if an anchor tag is present in the URL. This will allow for something like the [standard page jumps](https://make.wordpress.org/support/user-manual/content/pages/page-jumps/).

This could be done by changing [this line](https://github.com/siteorigin/so-widgets-bundle/blob/1.13.3/widgets/accordion/js/accordion.js#L124-L130) to:

```
var navOffset = 90;// Add some magic number offset to make space for possible nav menus etc.
if ( window.location.hash && openPanels.length ) {
	window.scrollTo( 0, $( openPanels[0] ).offset().top - navOffset );
} else if ( initialScrollPanel > 0 ) {
	var $initialScrollPanel = initialScrollPanel > $accordionPanels.length ?
		$accordionPanels.last() :
		$accordionPanels.eq( initialScrollPanel - 1 );
	window.scrollTo( 0, $initialScrollPanel.offset().top - navOffset );
}
```

That change doesn't, however, factor in the accordion items being _temporarily_ visible on load.

AlexGStapleton:
Resolve #785

### DIFF
--- a/widgets/accordion/js/accordion.js
+++ b/widgets/accordion/js/accordion.js
@@ -157,7 +157,7 @@ jQuery( function ( $ ) {
 				}
 				var initialScrollPanel = $widget.data( 'initialScrollPanel' );
 				if ( window.location.hash && openPanels.length ) {
-					setTimeout( function () {
+					setTimeout( function() {
 						scrollToPanel( $( openPanels[0] ) );
 					}, 500 );
 				} else if ( initialScrollPanel > 0 ) {
--- a/widgets/accordion/js/accordion.js
+++ b/widgets/accordion/js/accordion.js
@@ -156,7 +156,11 @@ jQuery( function ( $ ) {
 					updateHash();
 				}
 				var initialScrollPanel = $widget.data( 'initialScrollPanel' );
-				if ( initialScrollPanel > 0 ) {
+				if ( window.location.hash && openPanels.length ) {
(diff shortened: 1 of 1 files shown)


widgets/accordion/js/accordion.js
```diff
@@ -156,7 +156,11 @@ jQuery( function ( $ ) {
 					updateHash();
 				}
 				var initialScrollPanel = $widget.data( 'initialScrollPanel' );
-				if ( initialScrollPanel > 0 ) {
+				if ( window.location.hash && openPanels.length ) {
+					setTimeout( function () {
```
Misplon: Thanks, Alex. As far as I'm aware `function (` should be `function(`
https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#indentation-and-line-breaks

AlexGStapleton: Fixed.

Misplon:
I've done a functionality check and the PR is producing the result it sets out to produce.

PR #901, opened 2019-08-22
Contact: Add note to Default Subject about overriding Subject field types set to required.

AlexGStapleton:
A User emailed us and was confused why the Subject field wasn't respecting the Required setting. Turns out they had said a Default Subject set so that was being used when the user didn't input a subject. This, in hindsight, makes sense but it isn't mentioned anywhere. This PR will add a note to the Default Subject description to make this more apparent.


Misplon:
Thanks, mate. Quite tricky to make this easy to read. What do you think about the following?

```Subject to use when there isn't one supplied by the user. If you make use of this option it won't be possible to set the Subject field as required because the default subject will be used as a fallback.```

AlexGStapleton:
That works. 👍

PR #905, opened 2019-09-02
Add max number of items for repeater field

xella29:
Here are some code modifications of the plugin to add an option to limit number of repeater items for Repeater Fields.

### DIFF
--- a/base/js/admin.js
+++ b/base/js/admin.js
@@ -642,11 +642,19 @@ var sowbForms = window.sowbForms || {};
 
 			$el.find('> .siteorigin-widget-field-repeater-add').disableSelection().click(function (e) {
 				e.preventDefault();
-				$el.closest('.siteorigin-widget-field-repeater')
-					.sowAddRepeaterItem()
-					.find('> .siteorigin-widget-field-repeater-items').slideDown('fast', function () {
-					$(window).resize();
-				});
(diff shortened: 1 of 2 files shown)


base/js/admin.js
```diff
@@ -788,12 +796,18 @@ var sowbForms = window.sowbForms || {};
...
+                            $('.siteorigin-widget-field-repeater-add').removeClass('is-hidden');
+                            $('.siteorigin-widget-field-copy').show();
+						}
 					}
 				});
 				itemTop.find('.siteorigin-widget-field-copy').click(function (e) {
 					e.preventDefault();
 					var $form = $(this).closest('.siteorigin-widget-form-main');
 					var $item = $(this).closest('.siteorigin-widget-field-repeater-item');
+                    var $maxItemsRepeater = $('.siteorigin-widget-field-repeater').attr('data-max-items');
```
gregpriday: This could actually just be `$('.siteorigin-widget-field-repeater').data('max-items')`, which is slightly neater. Also, rather than casting `$maxItemsRepeater` to an integer later on, this could be done on this line.

We're also using `$` in front of a variable name to represent a jQuery object. So in this particular case, the variable should just be `maxItemsRepeater`.

Misplon:
Hi @xella29 

Thanks for submitting your PR. I'm going to close this for the time being. If you'd like to re-open and make the requested changes, that'll be great. If not, no worries :)

Thanks again.

PR #909, opened 2019-10-23
Add fallback URL for testimonial images

Misplon:
Add the ability to add a fallback/external image for Testimonial images. 


braamgenis:
Ah, unfortunately it's not that simple. You also need to check whether an image has been set and, if not, then use the fallback in the template. [This function](https://github.com/siteorigin/so-widgets-bundle/blob/2fa546c356cbeb97db0dbe1d880ed19f5e607612/widgets/testimonial/testimonial.php#L339-L351) gets called from the template to render the image. So it should take the optional fallback url as an additional argument and use that if the attachment image is missing.

Misplon:
Ahh, my bad! Will log this as a feature request.

PR #912, opened 2019-10-24
SO Features: Container Position: Swap Right and Left values

AlexGStapleton:
The SiteOrigin Features Widget Container Position Icon Horizontal options control the container position while the top and bottom settings affect the icon placement. This is problematic as left actually means right and right means left. This PR will swap the left and right values,  but will still internally refer to them as the left and right for backwards compatibility.

### DIFF
--- a/widgets/features/features.php
+++ b/widgets/features/features.php
@@ -57,6 +57,8 @@ class SiteOrigin_Widget_Features_Widget extends SiteOrigin_Widget {
 						'default' => '#404040',
 					),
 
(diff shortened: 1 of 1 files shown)


Misplon:
@braamgenis To put this another way, when you set the Icon container position to left, the icon should appear on the left. At the moment, this doesn't happen. So we're looking to correct the behavior of the Icon Container Position setting without disturbing existing user's layouts. This seemed like the simplest option. The only other option I can think of would be to migrate settings.

Misplon:
@braamgenis How's that? Let me know if you'd like any further changes. If not, I'll move onto a release pull request.

PR #925, opened 2019-12-05
Contact: Check from field for dev emails and replace it

AlexGStapleton:
Currently, dev emails are replaced when the mail is sent. For consitency, we should replace it before then.


Misplon:
@gregpriday A quick code review when able, thanks mate.

Issue #926, opened 2019-12-09
Measurement Input: Line-height issue on mobile


Misplon:
![IMG_0100](https://user-images.githubusercontent.com/789159/70450605-aef1ca00-1aac-11ea-9a7d-df5486d65fc8.png)

PR #927, opened 2019-12-09
Removed static height declaration

Misplon:
Resolves https://github.com/siteorigin/so-widgets-bundle/issues/926.

@AlexGStapleton Can you see an issue with doing this?

### DIFF
--- a/base/inc/fields/css/measurement-field.less
+++ b/base/inc/fields/css/measurement-field.less
@@ -3,6 +3,7 @@
         float: left;
         max-width: 58px;
         margin: 1px;
+        min-height: 28px;
         vertical-align: middle;
     }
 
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
I don't see any direct issues with doing this, but I'm not familiar why a height was specified in the first place so there may be something I'm oving looking.

To play it safe, it might be better to change `height` to `min-height` rather than removing the height altogether. That's what's currently done for the drop-down in the measurement field.

Misplon:
Sounds good. Added `min-height`.

AlexGStapleton:
Looks good! 👍

Misplon:
Thanks!

@gregpriday quick check when ready.

PR #934, opened 2019-12-15
Added right margin to color picker input

Misplon:
Resolves:

spacing

### DIFF
--- a/base/css/admin.less
+++ b/base/css/admin.less
@@ -111,6 +111,7 @@
 		}
 
 		input.siteorigin-widget-input-color {
+			margin-right: 6px;
 			width: auto;
 		}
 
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
This works, but it might be a better idea to apply the spacing directly to the Clear button. My reasoning for this is that the Default button already has spacing applied so if a Default button is present, it results in additional spacing when this change is present.

Before:


After:

Misplon:
Good points. Please, now check style fixes for Clear and Select Content buttons. Select Content is the button for Destination URL. Thanks.

Misplon:
Probably should have gone with another PR but thought we could squeeze it in here this time.

Misplon:
Super, thanks.

PR #935, opened 2019-12-15
SiteOrigin Features: Allow Selection of Different Title Header Levels

### LINKED ISSUE #107: SiteOrigin Features: Allow selection of different title header levels
Currently, they default to h5 and users may wish to increase this depending on the importance of the header for SEO purposes.

Reference thread: https://siteorigin.com/thread/how-to-change-siteorigin-features-title-heading/

AlexGStapleton:
This PR resolves #107.

### DIFF
--- a/widgets/features/features.php
+++ b/widgets/features/features.php
@@ -221,8 +221,8 @@ class SiteOrigin_Widget_Features_Widget extends SiteOrigin_Widget {
 			),
 
 			'title_tag' => array(
-				'type'    => 'select',
-				'label'   => __( 'Title Text HTML Tag', 'so-widgets-bundle' ),
+				'type' => 'select',
+				'label' => __( 'Title Text HTML Tag', 'so-widgets-bundle' ),
 				'default' => 'h5',
 				'options' => array(
 					'h1' => __( 'H1', 'so-widgets-bundle' ),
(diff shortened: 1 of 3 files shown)


widgets/features/features.php
```diff
@@ -220,6 +220,20 @@ function get_widget_form(){
 				'default' => false,
 			),
 
+			'title_tag' => array(
+				'type' => 'select',
+				'label' => __( 'Title Text HTML Tag', 'so-widgets-bundle' ),
```
Misplon: @AlexGStapleton SO is using Sentence case so `Title Text HTML Tag` should be changed to `Title text HTML tag`. Personally, I'd prefer Title Case but to match the standard already in use across the plugin and Widgets Bundle, it should be sentence case.

AlexGStapleton: Ah, sorry I had forgotten to check this.

Misplon: No worries, all sorted now.

Misplon:
@gregpriday I'm happy with the functionality. A quick code review when able, thanks mate.

widgets/features/tpl/default.php
```diff
@@ -38,11 +38,11 @@ class="sow-icon-container <?php echo !empty($instance['container_shape']) ? 'sow
 
 				<div class="textwidget">
 					<?php if(!empty($feature['title'])) : ?>
-						<h5>
+						<<?php echo esc_attr( $instance['title_tag'] ); ?>>
```
gregpriday: I think that `esc_html` might be a better choice here, but both work.

PR #939, opened 2019-12-26
"Pre-fill" hero if height is set

### LINKED ISSUE #937: Hero: If Height set, add height to Hero widget container to prevent "jump"
The Hero widget doesn't have a proper sentinel as it's instead applied to a container within the slider itself which isn't visible until after the DOM has loaded. This can result in the Hero widget "jumping" into view. This can be avoided by applying the Height, if set, to a container outside of the slider but still within the regular widget container.

AlexGStapleton:
Resolve #937

### DIFF
--- a/widgets/hero/styles/default.less
+++ b/widgets/hero/styles/default.less
@@ -28,6 +28,14 @@
 @responsive_breakpoint: 780px;
 @slide_height_responsive: default;
 
+& when ( isnumber( @slide_height ) ) {
+	@media (min-width: @responsive_breakpoint) {
+		& {
+			min-height: @slide_height;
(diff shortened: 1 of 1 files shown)


Misplon:
Looks good. Should we add another rule if a responsive height is set?

AlexGStapleton:
We can. I thought about doing that but honestly, I have never come across any reports of this ever being an issue on mobile.

Misplon:
Yeah, let's take care of this for mobile too while we are here.

Misplon:
@gregpriday Functionality is good, just a quick code review, please. Thanks mate.

PR #940, opened 2019-12-26
Set Font-display: swap on font imports

### LINKED ISSUE #938: Slider widgets:  Consider adding font-display for web fonts
Consider adding `font-display` for the slider web-font.

AlexGStapleton:
Resolve #938


Misplon:
Hey brother, I'm not sure we're working in the correct file here. `font-display` is only valid when applied to the `@font-face` rule.

https://css-tricks.com/almanac/properties/f/font-display/

AlexGStapleton:
Oh. I must have missed that. I've reverted the previous commit and applied the correct change

PR #946, opened 2019-12-30
Add class to each Image Grid image

### LINKED ISSUE #910: LiteSpeed Lazy Loading negatively impacts the SO Image Grid widget
To replicate this issue add the Mono prebuilt layout to a page and activate LiteSpeed Cache plugin. Scroll down to the Our Work section:

Page_Builder_–_SiteOrigin

Next, deactivate LiteSpeed Cache or just Lazy Loading at Settings > LiteSpeed Cache > Media. The Image Grid loads normally:

Banners_and_Alerts_and_Page_Builder_–_SiteOrigin

Perhaps the solution is to exclude Lazy Load from the grid images as we've done with the [Corp logo](https://github.com/siteorigin/siteorigin-corp/blob/develop/inc/template-tags.php#L132-L147) or perhaps there is a wider solution. The issue we've run into with the Corp logo is each lazy loader needs its own function to exclude. Jetpack, LiteSpeed, Smush, they all need a function.

AlexGStapleton:
Resolves #910
Related #945

I used sow-image-grid-image_html instead of sow-image-grid-image due to that being using as a container. `image_html` was used due to it being the array key of where the image HTML is stored prior to output.

### DIFF
--- a/widgets/image-grid/image-grid.php
+++ b/widgets/image-grid/image-grid.php
@@ -134,7 +134,7 @@ class SiteOrigin_Widgets_ImageGrid_Widget extends SiteOrigin_Widget {
 
 			if ( empty( $image['image'] ) && ! empty( $image['image_fallback'] ) ) {
 				$alt = ! empty ( $image['alt'] ) ? $image['alt'] .'"' : '';
-				$image['image_html'] = '<img src="'. esc_url( $image['image_fallback'] ) .'" alt="'. esc_attr( $alt ) .'" title="'. esc_attr( $title ) .'">';
+				$image['image_html'] = '<img src="'. esc_url( $image['image_fallback'] ) .'" alt="'. esc_attr( $alt ) .'" title="'. esc_attr( $title ) .'" class="sow-image-grid-image_html">';
 			} else {
 				$image['image_html'] = wp_get_attachment_image( $image['image'], $instance['display']['attachment_size'], false, array(
 					'title' => $title,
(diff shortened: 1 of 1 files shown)


widgets/image-grid/image-grid.php
```diff
@@ -134,7 +134,7 @@ function get_template_variables( $instance, $args ) {
 
 			if ( empty( $image['image'] ) && ! empty( $image['image_fallback'] ) ) {
 				$alt = ! empty ( $image['alt'] ) ? $image['alt'] .'"' : '';
-				$image['image_html'] = '<img src="'. esc_url( $image['image_fallback'] ) .'" alt="'. esc_attr( $alt ) .'" title="'. esc_attr( $title ) .'">';
+				$image['image_html'] = '<img src="'. esc_url( $image['image_fallback'] ) .'" alt="'. esc_attr( $alt ) .'" title="'. esc_attr( $title ) .'" class="sow-image-grid-image_html">';
```
Misplon: Due to the conditional this change is contained in, the class will only be added if a fallback is used and an image is not added via the media button.

Misplon:
@gregpriday Do you have any feedback regarding the image class name used? Thanks.

widgets/image-grid/image-grid.php
AlexGStapleton: Oh sorry about that. I was using a prebuilt Layout and had forgotten to check this.

PR #949, opened 2020-01-04
Account for false values when setting default values rather than just empty values

### LINKED ISSUE #919: SO Widget Block: Features defaults aren't present
In the SO Widget Block, the SiteOrigin Features defaults aren't present. 

Icon container size
Icon size
Features per row

Empty, empty and zero are incorrect defaults for those fields.

AlexGStapleton:
Resolves #919.

### DIFF
--- a/base/siteorigin-widget.class.php
+++ b/base/siteorigin-widget.class.php
@@ -406,7 +406,7 @@ abstract class SiteOrigin_Widget extends WP_Widget {
 				$instance[$id] = $this->add_defaults( $field['fields'], $instance[$id], $level + 1 );
 			}
 			else if( $field['type'] == 'measurement' ) {
-				if( ! isset( $instance[$id] ) ) {
+				if( ! isset( $instance[$id] ) || $instance[$id] === false ) {
 					$instance[$id] = isset( $field['default'] ) ? $field['default'] : '';
 				}
 				if ( empty( $instance[ $id . '_unit' ] ) ) {
@@ -424,7 +424,7 @@ abstract class SiteOrigin_Widget extends WP_Widget {
 				}
 			}
 			else {
-				if( !isset( $instance[$id] ) ) {
+				if( !isset( $instance[$id] ) || $instance[$id] === false ) {
 					$instance[$id] = isset( $field['default'] ) ? $field['default'] : '';
 				}
 			}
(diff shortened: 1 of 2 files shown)


Misplon:
In the Classic Editor, from what I can see, the Editor widget "Automatically add paragraphs" settings is default enabled. Currently, in the development branch, this doesn't appear to be working in the SO Widget Block but it looks like this PR fixes that issue and enables the "Automatically add paragraphs" by default.

base/siteorigin-widget.class.php
```diff
@@ -406,7 +406,7 @@ function add_defaults( $form, $instance, $level = 0 ){
 				$instance[$id] = $this->add_defaults( $field['fields'], $instance[$id], $level + 1 );
 			}
 			else if( $field['type'] == 'measurement' ) {
-				if( ! isset( $instance[$id] ) ) {
+				if( ! isset( $instance[$id] ) || $instance[$id] === false ) {
```
braamgenis: Something's off if a measurement field can have it's value set to `false`. This might be treating the symptom, but there's probably an underlying issue that should be addressed too.

base/siteorigin-widget.class.php
```diff
@@ -424,7 +424,7 @@ function add_defaults( $form, $instance, $level = 0 ){
 				}
 			}
 			else {
-				if( !isset( $instance[$id] ) ) {
+				if( !isset( $instance[$id] ) || $instance[$id] === false ) {
```
braamgenis: Same here. This is basically just sweeping a deeper issue under the rug. Default values should only be used for fields that have not been set. This also prevents a valid value of `false` from being used.

Misplon:
Thanks Braam, let us know what you find.

braamgenis:
Ok, so the underlying issue here is that there's an attempt to sanitize values the first time the widget form is rendered, when the widget instance is just an empty array. This leads to keys incorrectly being set to `false`. The offending code is here:

https://github.com/siteorigin/so-widgets-bundle/blob/1.15.9/base/inc/routes/siteorigin-widgets-resource.class.php#L93

A quick fix here is to simply avoid the call to `update` for empty instances.
```
if ( ! empty( $widget_data ) ) {
	$widget_data = $widget->update( $widget_data, $widget_data );
}
```

Misplon:
Thanks, guys, looking good in `develop`, I did a quick test now.

PR #963, opened 2020-01-10
Added line-height to section heading

Misplon:
The section heading alignment relies on a line-height that changes in the Block Editor. The result is seen below.

widget-block

By stating the line height we're expecting in the Classic Editor we can fix this.

@AlexGStapleton Is `.siteorigin-widget-field-type-widget` used in Beaver Builder? I've only tested this targeting `.siteorigin-widget-field-type-section`. Can you test this PR in BB?

### DIFF
--- a/base/css/admin.less
+++ b/base/css/admin.less
@@ -428,6 +428,7 @@
 		> label {
 			background: #F0F0F0;
 			border: 1px solid #D0D0D0;
+			line-height: 1.4;
 			padding: 10px;
 			display: block;
 			margin-bottom: 0;


AlexGStapleton:
`.siteorigin-widget-field-type-widget` isn't used in Beaver Builder.

Misplon:
@AlexGStapleton Can you test something that is using `siteorigin-widget-field-type-widget`? Put another way, I've only tested `.siteorigin-widget-field-type-section`.

AlexGStapleton:
Ah, I was checking the wrong widget previously and had overlooked its usage.
`siteorigin-widget-field-type-widget ` is used, but the `.siteorigin-widget-field-type-widget  > label` is never used - this is the case for Beaver Builder, SiteOrigin Page Builder, the SiteOrigin Layout Block, and the SiteOrigin Widgets Block. It's always, `.siteorigin-widget-field-type-widget > .siteorigin-widget-widget >.siteorigin-widget-field > label`. More containers were added at some point and the old and no longer valid selector, `.siteorigin-widget-field-type-widget > label`, wasn't.

Misplon:
Super, thanks for checking.

PR #964, opened 2020-01-10
Switch media field and external URL margin

Misplon:
When the Block Editor is under about 690px in width, the external URL field collapses.

image-grid

I'm not sure how to detect the width of the Block Editor with CSS to set breakpoints for different behavior so I've just switched the margin from the external URL field to the media field. If you have any better ideas, let me know. I think this is sufficient but I am open to ideas.

### DIFF
--- a/base/inc/fields/css/media-field.less
+++ b/base/inc/fields/css/media-field.less
@@ -6,6 +6,7 @@
 		position: relative;
 		display: block;
 		float: left;
+		margin-right: 25px;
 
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
> I'm not sure how to detect the width of the Block Editor with CSS to set breakpoints for different behavior

Without JavaScript, you can't. The most you can do is make an assumption, or do what you did - which is fine imo.

PR #970, opened 2020-01-15
SO Premium Tabs: CSS for Custom icons

AlexGStapleton:
Requires siteorigin/siteorigin-premium/pull/344


braamgenis:
This breaks the `Tabs position` setting when set to left or right.

PR #971, opened 2020-01-15
SO Premium Accordion: CSS for Custom icons

AlexGStapleton:
Requires: siteorigin/siteorigin-premium/pull/345

### DIFF
--- a/widgets/accordion/styles/default.less
+++ b/widgets/accordion/styles/default.less
@@ -39,11 +39,17 @@
 			padding: @heading_padding;
 			background-color: @heading_background_color;
 			color: @title_color;
-			text-align: @heading_title_align;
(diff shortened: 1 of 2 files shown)


widgets/accordion/styles/default.less
```diff
@@ -64,20 +70,33 @@
...
-
-				& when ( @show_open_close_icon = false ) {
-					display: none;
+			& when ( @show_open_close_icon = false ) {
+				.sow-accordion-open-close-button {
+						display: none;
 				}
 			}
+
+			.sow-accordion-icon-image {
```
Misplon: I know there is a big mix of formatting around but if possible, let's organize selector alphabetically when adding new rules. Thanks.

PR #976, opened 2020-01-21
Testimonial: Enforce Image Size for Square Container

### LINKED ISSUE #953: Testimonials: Image size isn't applied to square images
When Image shape is set to Square, Image size doesn't appear to be functioning.

AlexGStapleton:
Resolve #953

### DIFF
--- a/widgets/testimonial/styles/default.less
+++ b/widgets/testimonial/styles/default.less
@@ -82,7 +82,7 @@
 		.rounded(@text_border_radius);
 	}
 
-	.sow-round-image-frame {
+	.sow-round-image-frame, .sow-image-shape-square {
 		border-radius: @image_size;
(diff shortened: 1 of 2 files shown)


Misplon:
Thanks, brother. On my end it looks like square designated images are now round.

Misplon:
This isn't creating great results for me. What if, when square, we set the container width and height to the user set image size. So if it's `50px` we set `.sow-image-wrapper` to `width: 50px`, `height: 50px;`. We then leave the image to fill that with the existing `max-width: 100%;` and `height: auto` which most themes will apply.

AlexGStapleton:
How does that look? I had to apply an extra @max-width due to .sow-image-wrapper  having a defined width in certain instances.

PR #978, opened 2020-01-27
Slider: Add class to slide images

AlexGStapleton:
This PR will allow users to exclude the SiteOrigin Slider slide images from lazy load plugins. This was directly in response to a report about a display issue the SiteOrigin Slider widget has with the Jetpack Lazy Load.

This PR also extends the `siteorigin_widgets_get_attachment_image` function by adding a `class` parameter that allows developers to easily give images a class.

### DIFF
--- a/base/inc/attachments.php
+++ b/base/inc/attachments.php
@@ -32,21 +32,18 @@ function siteorigin_widgets_get_attachment_image_src( $attachment, $size, $fallb
 	return false;
 }
 
-function siteorigin_widgets_get_attachment_image( $attachment, $size, $fallback, $class = false ){
+function siteorigin_widgets_get_attachment_image( $attachment, $size, $fallback, $atts = array() ){
 	if( !empty( $attachment ) ) {
-		if ( ! empty( $class ) ) {
-			$class = array( 'class' => $class );
-		}
-
-		return wp_get_attachment_image( $attachment, $size, false, $class );
+		return wp_get_attachment_image( $attachment, $size, false, $atts );
 	}
 	else {
 		$src = siteorigin_widgets_get_attachment_image_src( $attachment, $size, $fallback );
 		if( empty($src[0]) ) return '';
 
-		$atts = array(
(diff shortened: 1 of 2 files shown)


widgets/slider/slider.php
```diff
@@ -154,7 +154,8 @@ function render_frame_contents($i, $frame) {
 					echo siteorigin_widgets_get_attachment_image(
 						$frame['foreground_image'],
 						'full',
-						!empty( $frame['foreground_image_fallback'] ) ? $frame['foreground_image_fallback'] : ''
+						!empty( $frame['foreground_image_fallback'] ) ? $frame['foreground_image_fallback'] : '',
+						'sow-slider-forground-image'
```
Misplon: Looks like a typo here.

base/inc/attachments.php
```diff
@@ -32,9 +32,13 @@ function siteorigin_widgets_get_attachment_image_src( $attachment, $size, $fallb
 	return false;
 }
 
-function siteorigin_widgets_get_attachment_image( $attachment, $size, $fallback ){
+function siteorigin_widgets_get_attachment_image( $attachment, $size, $fallback, $class = false ){
```
gregpriday: We should really keep the arguments of `siteorigin_widgets_get_attachment_image` the same as `wp_get_attachment_image`. So rather than having a `$class` argument, it should be a general `$atts` that's processed and passed through to `wp_get_attachment_image`.

widgets/slider/slider.php
```diff
@@ -155,7 +155,7 @@ function render_frame_contents($i, $frame) {
 						$frame['foreground_image'],
 						'full',
 						!empty( $frame['foreground_image_fallback'] ) ? $frame['foreground_image_fallback'] : '',
-						'sow-slider-foreground-image'
+						array( 'class' => 'sow-slider-forground-image' )
```
Misplon: Typo was re-introduced here. Thanks.

PR #983, opened 2020-02-06
Use first element in $widget_icon_families instead of raw string.

mamarmite:
I'm currently adding fontawesome pro to this field. And I'm removing other font available to the user. So if fontawesome doesn't exist in the array, my custom font was not rendering.

Tested on debian and php 7.2

### DIFF
--- a/icons/icons.php
+++ b/icons/icons.php
@@ -4,7 +4,7 @@ define( 'SITEORIGIN_WIDGETS_ICONS', true );
 
 function siteorigin_widgets_icon_families_filter( $families ){
 	$bundled = array(
-        'fontawesome' => __( 'Font Awesome', 'so-widgets-bundle' ),//moved first to fit old behavior of default fonts.
+        'fontawesome' => __( 'Font Awesome', 'so-widgets-bundle' ),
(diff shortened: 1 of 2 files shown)


AlexGStapleton:
Hi @mamarmite,

Thank you for submitting a PR. Can you please remove the comments? The changes made are self-explanatory and for those that need additional context [the blame](https://help.github.com/en/github/managing-files-in-a-repository/tracking-changes-in-a-file) surfaces.

mamarmite:
Awesome, yes np :)

AlexGStapleton:
Thanks mate! :)

icons/icons.php
```diff
@@ -4,8 +4,8 @@
 
 function siteorigin_widgets_icon_families_filter( $families ){
 	$bundled = array(
+        'fontawesome' => __( 'Font Awesome', 'so-widgets-bundle' ),
```
gregpriday: It's minor, but can you use tab indenting here instead of space indenting.

mamarmite: Yes, np ! I missed that ;)

PR #985, opened 2020-02-11
Accordion: When hiding panels, only search current accordion and not nested accordions

AlexGStapleton:
This PR prevents potential visability issues if the user is nesting accodions inside of an accordion that's not open by default.


Misplon:
When switching to this branch, the accordion open/close buttons don't appear to be working.

AlexGStapleton:
How does that look?

Misplon:
Looks good.

PR #987, opened 2020-02-17
Remove slider slide number from output

AlexGStapleton:
Related #747.

This PR prevents the output of the slide numbers to prevent an issue where they'll appear in post excerpts. Previously, it was argued that these slide numbers should be retained for accessibility reasons but there's already `aira-label` in place which is sufficient for accessibility. The previously decided solution was to implement functionality that automatically generates an excerpt from the first SiteOrigin Editor widget. While useful, we've run into situations where an editor isn't being used.


Misplon:
The PR resolves the search results excerpt issue. Thanks.

@gregpriday please, review when able. Thanks.

PR #990, opened 2020-02-23
Fix Tab Alignment

AlexGStapleton:


Misplon:
Thanks Alex. 

Please, can you run through all of the possibilities in testing for us. All of the alignment options in combination with all of the position options. 

Here is how Top and Right currently render in the release version:

right-top

And in this branch:

Edit_Page_‹_SiteOrigin_Premium_Demo_—_WordPress

Issue #885, opened 2019-07-24
Plugin Build wp-pot: Set metadataFile to allow for widget descriptions to be translatable (edited) 


AlexGStapleton:
Related: siteorigin/plugin-build#9

SiteOrigin Premium and the SiteOrigin Widgets Bundle recently received changes to allow for addon and widget descriptions to be translatable.

siteorigin/siteorigin-premium#98
siteorigin/so-widgets-bundle@0810a64

A user has reported that the SIteOrigin Widgets Bundle plugin descriptions aren't being completely translated. Only the Hero and Layout Slider widgets are the only widget description are present in their respective pot files. All other widget descriptions are missing.

[A comment on the SiteOrigin Premium PR](https://github.com/siteorigin/siteorigin-premium/issues/98#issuecomment-450412091) mentioned that plugin build needed to be updated. Specifically, wp-pot needed [metadataFile ](https://github.com/rasmusbe/wp-pot#options) to be set allow for the descriptions to be correctly extracted

PR #993, opened 2020-03-04
#885 Make description translations more reliable.

aequitascs:
This issue is based on the fact that the widgets are given descriptions in two places:

1. In the PHP comment that comprises the Widget file header.
2. In the Widget constructor where the description is set manually in an array passed as the 3rd parameter to the parent constructor.

The POT file in the latest distribution includes both of these description settings. However, any translation will only be applied to number 1 in the list above. The problem is that there are obvious Widget headers in the POT file that come from number 2 in the list above that are different from the headers that come from number 1. Any translations associated with these number 2 descriptions will never be displayed because the $widget object passed into the admin.php file will always use the description set in number 1.

Therefore, the basic problem of the number 1 descriptions showing up in the POT file was already solved. The secondary problem of translations not showing up was being caused by translations being associated with the unused number 2 descriptions. The solution is to make number 1 and number 2 headers identical for all Widget files to eliminate the confusing extra entries in the POT file. #885

### DIFF
--- a/widgets/accordion/accordion.php
+++ b/widgets/accordion/accordion.php
@@ -14,7 +14,7 @@ class SiteOrigin_Widget_Accordion_Widget extends SiteOrigin_Widget {
 			'sow-accordion',
 			__( 'SiteOrigin Accordion', 'so-widgets-bundle' ),
 			array(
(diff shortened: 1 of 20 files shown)


aequitascs:
Cancelled in favour of creating a new pull request based on tags.

PR #995, opened 2020-03-04
#885: Improved reliability of widget description translations.

aequitascs:
Unlike what was described in #885, I was able to find all of the Widget descriptions in the POT file. The problem was that most widgets had two descriptions:

1. The PHP comment description, which is used to describe widget in the Page Builder editor.
2. The description in the post options array in the constructor that is sent to the parent constructor, which is used to describe widgets in the Wordpress Appearance/Widgets screen.

I was surprised to see that the POT file generator picks up the PHP comment description, and I have confirmed that the translations all work perfectly on that value.

The problem that I encountered when trying to test this issue was that the number 1 and number 2 descriptions from the list above were usually different. I would create a translation for what looked like the button widget and it wouldn't appear when administering SO widgets.

I've therefore made the widget descriptions uniform for each widget and they appear correctly in both views when translated. Unfortunately, both descriptions are required to service the two different views.

### DIFF
--- a/widgets/accordion/accordion.php
+++ b/widgets/accordion/accordion.php
@@ -14,7 +14,7 @@ class SiteOrigin_Widget_Accordion_Widget extends SiteOrigin_Widget {
 			'sow-accordion',
 			__( 'SiteOrigin Accordion', 'so-widgets-bundle' ),
 			array(
-				'description' => __( 'An accordion widget.', 'so-widgets-bundle' ),
+				'description' => __( 'An accordion to squeeze a lot of content into a small space.', 'so-widgets-bundle' ),
 				'help' => 'https://siteorigin.com/widgets-bundle/accordion-widget/',
 			),
 			array(),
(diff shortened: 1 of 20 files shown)


Misplon:
Thanks, Andrew :)

There are a few cases in this PR where I think we should use an alternative description. For example, we're changing:

Share your product/service testimonials in a variety of different ways.
to 
Display some testimonials.

In the above case, I think the first description is the one we should use. Would you like to first standardize the descriptions in this PR and then change a few of them in another PR or do it all in this PR?

aequitascs:
Hi Andrew

Yeah, actually it did not escape me that some of the descriptions were better in Position 2 than in Position 1. I overlooked this because I was also snagged by some grammatical inelegancies in the Position 2 descriptions, so I had assumed that the Position 1 descriptions were what everyone had agreed to.

I'm going to update my code for this ticket with what I think are the best for each of the widgets, so we can do it all in this PR.

Cheers!

Misplon:
Awesome, sounds good, thanks :)

aequitascs:
Done!

Issue #994, opened 2020-03-04
Error message when previewing with the Editor Widget


aequitascs:
With WP_DEBUG and WP_DEBUG_DISPLAY on, the following message is displayed when previewing a post containing the Editor Widget:

Notice: Trying to get property 'ID' of non-object in /Users/andrewf/EclipseProjects/so-wordpress/wp-content/plugins/so-widgets-bundle/widgets/editor/editor.php on line 118

To reproduce the problem:

Create a new post
Add an Editor Widget
Enter some text
Click _Preview_

If WP_DEBUG is enabled, the error will occur.

This was discovered while investigating #984.

Misplon:
Thanks for logging @aequitascs. I have confirmed.

aequitascs:
Further investigation shows that this message occurs in two circumstances.

1. When previewing a new post that has not yet been saved to the database.

2. When previewing the contents of the TinyMCE editor using the embedded preview button, under all circumstances.

This is because the $post object itself is _null_ under those conditions.

Misplon:
Resolved via https://github.com/siteorigin/so-widgets-bundle/pull/996.

PR #996, opened 2020-03-04
#994: Fixed PHP Notice on Editor Widget previews.

aequitascs:
The PHP Notice message in #994 is coming from the situation where the get_post() function returns a null value. This happens under two conditions:

1. The post is previewed before it has been saved for the first time, immediately after creation.
2. Every time the TinyMCE editor's preview function is used.

To resolve this I have simple wrapped the return of the get_post() function in a null check.

The $post->ID is still used later on in the code and may give a similar message. However, since this only applies to Classic Editor pages, that have "more" link functionality, I suspect that all pages hitting that code would already be saved and that the error will therefore not occur. I have, at least, not been able to reproduce the error with widgets in the Classic Editor. Since it only shows up with WP_DEBUG on, it wouldn't be a show-stopping error anyway.

### DIFF
--- a/widgets/editor/editor.php
+++ b/widgets/editor/editor.php
@@ -115,7 +115,7 @@ class SiteOrigin_Widget_Editor_Widget extends SiteOrigin_Widget {
 
 	private function process_more_quicktag( $content ) {
 		$post = get_post();
-		if ( ! empty($post) ) {
+		if ( ! empty( $post ) ) {
 			$panels_content = get_post_meta( $post->ID, 'panels_data', true );
 		}
 		// We only want to do this processing if on archive pages for posts with non-PB layouts.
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
Thank you for your PR! We really appreciate your support! :)

Can you please format your changes using the [WordPress PHP Code Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/)?

Also, please use `empty` rather than `is_null` as there are situations where `get_post_meta` will return false rather than a null value.

Kind regards,
Alex

aequitascs:
Done! Please let me know if you need any other changes, and thanks for drawing my attention to the coding standards.

AlexGStapleton:
Thanks. Can you please add spaces inside of the parentheses? ([related](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#space-usage))

aequitascs:
Sorry for missing that! It is done.

AlexGStapleton:
No worries mate. Everything looks good on my end. I've forwarded your PR to the rest of the team.

Issue #924, opened 2019-12-04
Simple Masonry: Change Item label to image Title if available


Misplon:
Currently, it's hard re-ordering the Simple Masonry and similar widgets because all you see is:

<img width="490" alt="Edit_Page_‹_SiteOrigin_Premium_Demo_—_WordPress" src="https://user-images.githubusercontent.com/789159/70155573-2c879580-16bb-11ea-9500-9e8ca24f30fe.png">

Let's change "Item" to the image Title tag if available.

AlexGStapleton:
I just went to do this and it appears to have already in place. Can you please try setting a Title, saving and then checking if a title has been set?

![](https://i.imgur.com/JFJrrst.png)

Misplon:
Thanks, you're right. Perhaps I was logged this for the incorrect widget, not sure. Closing.

aequitascs:
Hi guys. I've already sunk quite a lot of time into resolving this, so I'd like to request that this is re-opened.

When I tested it, I took a literal interpretation of the original request. The "title" field in each masonry element is different from the _image_ title that is associated with the image in the media library. As the code stands at the moment, the masonry items will all appear as "Item" if the title field is not populated.

However, I've made a change (I'm about to check it in and generate a PR) that actually uses the image's title setting in Wordpress is the title is not set in the masonry element form. By default in Wordpress, this is the image file name. The result is this:

![image](https://user-images.githubusercontent.com/23128430/76241495-be181d80-623d-11ea-99d3-fdb68abcb448.png)

In this example, all titles like "Test X" are from the "Title" field, and all others come directly from the Wordpress media title associated with the image.

So, I think this might be an improvement that addresses the original problem, which will still occur when masonry images don't have an explicit title set in their own form.

Misplon:
Thanks. Using the image title from the media library as a fallback for when the title field in the widget is not populated is perfect.

Misplon:
Resolved via https://github.com/siteorigin/so-widgets-bundle/pull/997.

PR #997, opened 2020-03-09
#924: Change item label to image title if item

aequitascs:
Implemented fallback system for masonry form item labels where the Wordpress media tag will be used as the item label when the form's title field is not populated.

### DIFF
--- a/base/js/admin.js
+++ b/base/js/admin.js
@@ -733,23 +733,21 @@ var sowbForms = window.sowbForms || {};
 				var $parentRepeater = $el.closest('.siteorigin-widget-field-repeater');
 				var itemTop = $el.find('> .siteorigin-widget-field-repeater-item-top');
 				var itemLabel = $parentRepeater.data('item-label');
-				if (itemLabel && itemLabel.selector) {
+				if ( itemLabel && ( itemLabel.hasOwnProperty( 'selector' ) || itemLabel.hasOwnProperty( 'selectorArray' ) ) ) {
 					var updateLabel = function () {
-						var functionName;
-						var txt;
-						if ( itemLabel.hasOwnProperty('selectorArray') ) {
-							var selectorRow;
+						var functionName, txt, selectorRow;
+						if ( itemLabel.hasOwnProperty( 'selectorArray' ) ) {
 							for ( var i = 0 ; i < itemLabel.selectorArray.length ; i++ ) {
-								selectorRow = itemLabel.selectorArray[i];
-								functionName = ( selectorRow.hasOwnProperty('valueMethod') && selectorRow.valueMethod ) ? selectorRow.valueMethod : 'val';
-								txt = $el.find(selectorRow.selector)[functionName]();
+								selectorRow = itemLabel.selectorArray[ i ];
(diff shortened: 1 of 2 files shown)


Misplon:
Resolves https://github.com/siteorigin/so-widgets-bundle/issues/924.

Misplon:
Super, thanks for sorting this one :)

**Functionality**
I'm struggling to get this PR to work at the moment but perhaps it's something in my local setup that's the issue. I'll ask Alex to confirm. @AlexGStapleton is this PR working for you? 

**Really minor feedback**
The WordPress community is very keen on everyone capitalizing the "P". Ideally, it wouldn't be a big deal either way but, yeah, in our experience it is.

With regards to formatting, I'm never quite sure what to do with files where we have mixed formatting. So for example `if ( itemLabel.hasOwnProperty('selectorArray') ) {` ideally, that would be `if ( itemLabel.hasOwnProperty( 'selectorArray' ) ) {` with a space after the parentheses open and before the parentheses close as per https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/#examples-of-good-spacing. Perhaps, let's format this PR as per the guidelines, we can always come back later and do another PR to update the `admin.js` file to match the guidelines. Perhaps also check whether `[functionName]` should have spaces.

Thanks.

AlexGStapleton:
This wasn't working for me either. I had to change https://github.com/siteorigin/so-widgets-bundle/pull/997/files#diff-77ab07216416f9da215526118f391bd1R736 to `if ( itemLabel && ( itemLabel.selector || itemLabel.selectorArray ) ) {` to get it to work.

aequitascs:
I am sorry! That was my error in checking in my changes. I'll fix it with the code formatting.

AlexGStapleton:
No worries. Outside of that issue, and the formatting Andrew mentioned, I'm not seeing any real issues. The most important thing is that backwards compatibility is maintained and you've accounted for that.

Regarding formatting, I agree with @Misplon although there's a caveat. It's really tempting to update the formatting of surrounding lines - don't. You should, however, update the formatting of any lines you modify to be consistent with the entire line and your overall changes. In other words, this is bad:

> if (itemLabel && ( itemLabel.selector || itemLabel.selectorArray ) ) {

(notice the lack of spacing before the itemLabel check)

This is good:

> if ( itemLabel && ( itemLabel.selector || itemLabel.selectorArray ) ) {

aequitascs:
Done!

AlexGStapleton:
Thanks, that works with no additional changes for me.

Screenshot_2020-03-11 Edit Page ‹ SiteOrigin — WordPress

Ideally, please create a commit for each change - for example, one commit for fixing the merge bug and one fix to formatting. As a general rule of thumb, if you have to include "and" in the commit summary, you likely need to do multiple commits. This just makes looking up blames easier down the line as it makes it very clear that all of the changes are specific to the commit text and not just some of them.

Unrelated, but can you please be a little more descriptive in your branch names? While the issue id is great, it's not something that is instantly recognizable. Something like `item-label-from-items-924`, or something like that, tells me exactly what the overall goal of the branch is without having to look up the issue number.

Issue #986, opened 2020-02-13
Hero: Add a new setting to vertically align contents


Misplon:
@AlexGStapleton Can you please, take a look at the widget and let me know a suggested location and wording for the setting before you get going on this. Thanks mate.

AlexGStapleton:
Setting Name: Vertically Align Slides (checkbox)
Location: Just after Responsive Height setting in the Design and Layout settings group

This location should reduce the number of people using "the old way" of centering the contents using padding - something that would actually break the vertical alignment. It might be a good idea to add a description to the setting that mentions padding being factored in, but I'm drawing a blank on how to word that and have it make sense.

Misplon:
Super, thanks. How about tweaking the setting label wording to:

**Vertically align slide contents**

Location sounds good.

Let's see how it works, we can then work on the description sentence wrt. to explaining padding.

PR #998, opened 2020-03-09
Hero: Add "Vertically align slide contents" setting

AlexGStapleton:
Resolves https://github.com/siteorigin/so-widgets-bundle/issues/986

@Misplon You mentioned you wanted to see how it works before working on the description. How does the following (very rough suggestion) sound?

> Please note that Extra top padding may cause the slide contents to be slightly off-centered.

### DIFF
--- a/widgets/hero/styles/default.less
+++ b/widgets/hero/styles/default.less
@@ -27,6 +27,7 @@
 
 @responsive_breakpoint: 780px;
 @slide_height_responsive: default;
+@vertically_align: false;
 
 // "Pre-fill" hero widget area if height is set
 & when ( isnumber( @slide_height ) ) {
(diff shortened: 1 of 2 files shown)


widgets/hero/styles/default.less
```diff
@@ -55,6 +56,12 @@
 			max-width: @slide_width;
 			height: @slide_height;
 
+			& when (@vertically_align = true) {
+				display: flex;
+				justify-content: center;
+				flex-direction: column;
```
Misplon: What do you think about moving to alphabetical declarations? _s uses that style and I have been in the themes for a long time. `display, flex-direction and then justify-content` in this case. It's a simple organization method that's easy for everyone to follow.

Misplon:
@AlexGStapleton How about:

_For perfect centering, consider setting the Extra top padding setting to 0 when using this setting._

Misplon:
Otherwise, it looks all good mate.

AlexGStapleton:
> 
> 
> @AlexGStapleton How about:
> 
> _For perfect centering, consider setting the Extra top padding setting to 0 when enabling this setting._

I think that's fine. Adding now.

widgets/hero/styles/default.less
AlexGStapleton: Sounds good. I'll do this now.

Issue #1002, opened 2020-03-19
Slider: Change "Frame" label to image title


Misplon:
Currently, Slider widget frame labels don't update based on the content inserted.

![Edit_Page_‹_SiteOrigin_Premium_Demo_—_WordPress](https://user-images.githubusercontent.com/789159/77085986-c58fb180-6a09-11ea-972e-1efbb288cdf9.png)

Change the "Frame" label to match the media title from the Media Library. I think the order of priority should be:

1. Background image title.
2. Video title.
3. Foreground image title.

To confirm, by title I mean the title field in the Media Library.

@AlexGStapleton Are you happy with the above fallback order?

AlexGStapleton:
The suggested ordering is fine.

PR #1008, opened 2020-03-22
Slider: Set frame titles based on priority list

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1002


Misplon:
Looks, good mate.

Is it possible to have the frame label revert back to "Frame" when all images/videos are removed? That isn't happening right now. The same applies to the PR we have open for the Layout Slider.

AlexGStapleton:
That looks to be how the labels currently work in general as it also happens with the SiteOrigin Simple Masonry widget. I don't think that should be fixed in this PR and instead a differnet PR should be submitted. Do you agree?

Misplon:
Sounds good.

PR #1011, opened 2020-04-02
Contact: Add placeholder support to dropdown select field

AlexGStapleton:
This PR will move the field label inside of the select if the `Label Position` is set to `Placeholder`.

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -932,9 +932,9 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 			if ( $indicate_required_fields && ! empty( $field['required']['required'] ) ) {
 				$label .= '*';
 			}
-			$is_text_input_field = ( $field['type'] != 'select' && $field['type'] != 'radio' && $field['type'] != 'checkboxes' );
+			$no_placeholder_support = ( $field['type'] != 'radio' && $field['type'] != 'checkboxes' );
 			// label should be rendered before the field, then CSS will do the exact positioning.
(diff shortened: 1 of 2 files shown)


Misplon:
Thanks, Alex.

Ideally, I'd like to be able to return the select back to a state where I don't have to submit an option. Equally, I don't want the label to be sent as an option. Any ideas?

https://codepen.io/chriscoyier/pen/CiflJ

Misplon:
Perhaps: https://www.w3docs.com/snippets/css/how-to-create-a-placeholder-for-an-html5-select-box-by-using-only-html-and-css.html. 
Example: https://www.w3docs.com/tools/code-editor/2587

AlexGStapleton:
Currently, the Select form doesn't allow a blank value and the required check always passes so I didn't consider this. I'll remove `disabled` and set a blank value.

AlexGStapleton:
I reverted the rename as the base field doesn't support required. Do you want me to add a blank field that's selected by default if `Label Position` is set to `Placeholder` in this PR, or a different PR? It'll allow for the field to be set to required regardless of `Label Position`.

Misplon:
Because the issue persisted before this PR, I reckon it'll be better form if we fix that issue in another PR.

Misplon:
As per our Slack conversation, we'll do this all in one PR now.

Misplon:
@gregpriday Below is how the change looks on the front-end:

Screenshot 2020-04-05 at 17 01 11

It isn't possible to return to the label once a selection has been made but you can always view the label when opening the drop-down.

PR #1013, opened 2020-04-05
Font Awesome: Re-order style ordering

### LINKED ISSUE #1012: Font Awesome: Change icon group order
Change the icon group order to:

Regular
Solid
Brands

At the moment, Brands is the first item in the drop-down list.

AlexGStapleton:
Resolve #1012


Misplon:
Thanks. Let's put Brands last unless you disagree.

Issue #1014, opened 2020-04-06
Layout Slider: Add a setting to automatically vertically center contents 


Misplon:
As we did with the Hero widget, add a setting to automatically vertically center the contents of the slider.

PR #1015, opened 2020-04-06
Layout Slider: Add Setting that Allows Vertical Alignment

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1014

### DIFF
--- a/widgets/layout-slider/layout-slider.php
+++ b/widgets/layout-slider/layout-slider.php
@@ -138,6 +138,12 @@ class SiteOrigin_Widget_LayoutSlider_Widget extends SiteOrigin_Widget_Base_Slide
 						'label' => __( 'Responsive Height', 'so-widgets-bundle' ),
 					),
 
+					'vertically_align' => array(
+						'type' => 'checkbox',
+						'label' => __( 'Vertically align slide contents', 'so-widgets-bundle' ),
+						'description' => 'For perfect centering, consider setting the Extra top padding setting to 0 when enabling this setting.',
+					),
+
 					'padding' => array(
 						'type' => 'measurement',
 						'label' => __('Top and bottom padding', 'so-widgets-bundle'),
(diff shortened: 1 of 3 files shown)


widgets/layout-slider/layout-slider.php
```diff
@@ -138,6 +138,12 @@ function get_widget_form(){
 						'label' => __( 'Responsive Height', 'so-widgets-bundle' ),
 					),
 
+					'vertically_align' => array(
+						'type' => 'checkbox',
+						'label' => __( 'Vertically align slide contents', 'so-widgets-bundle' ),
```
Misplon, line 143: Should we change the label? I'm not sure vertically align necessarily means vertical center. 

**Vertically center align slide contents**

That work for you?

widgets/layout-slider/styles/default.less
```diff
@@ -42,6 +43,12 @@
 
 			max-width: @slide_width;
 			height: @slide_height;
+			
+			& when (@vertically_align = true) {
+				display: flex;
+				flex-direction: column;
```
Misplon, line 49: Are there any layouts that you can think of that will break or have an issue due to `flex-direction: column;`. I can't demonstrate a case but am asking the question. Can you think of a problem here?

widgets/layout-slider/layout-slider.php
AlexGStapleton: That works.

AlexGStapleton: Do you want this applied to the Hero widget also?

widgets/layout-slider/styles/default.less
AlexGStapleton: I just ran a few tests with five existing SO Layout Slider widgets I had set up and didn't run into any issues. Can you please elaborate on what you're worried may happen? I'll see if there's any steps I can take to prevent that from happening.

widgets/layout-slider/layout-slider.php
Misplon: Yes, please, apply to the Hero too.

widgets/layout-slider/styles/default.less
Misplon: No worries, as long as we've had a go at thinking of possible issues and tested, we're ok for now.

Misplon:
Quick conflict resolution when able, I should have waited to merge the Hero fix.

AlexGStapleton:
Oh! Sorry about that! 😞 I must have gotten lost when editing the files.

Fixed.

Issue #992, opened 2020-02-29
Post Carousel: Posts per page is being enforced when blank


Misplon:
Previously, when Posts per page was blank all posts would be available to the widget and loaded as the user scrolled the carousel. 1.16.0 broke this behavior with a default of 10 posts being loaded when the Posts per page field is blank.

Misplon:
This will hopefully be resolved as part of https://github.com/siteorigin/so-widgets-bundle/issues/1020.

Misplon:
Resolved via #1024.

PR #1024, opened 2020-04-16
Post Carousel: Migrate to Slick Carousel Library

### LINKED ISSUE #1020: Post Carousel: Migrate to Slick
Migrate the Post Carousel to [Slick](https://github.com/kenwheeler/slick).

We'll do more with Slick in the future. The initial task is to migrate the Post Carousel widget with all of the same options and behavior (or as close as possible) to Slick.

AlexGStapleton:
Resolves #1020.

Description Pending


AlexGStapleton:
I marked this PR as a draft as I want to do one last quick once before it gets reviewed. @Misplon you're welcome to look at it before then though.

Misplon:
Thanks, mate. I'm stoked to be getting this one done!

* Featured images don't seem to be displaying at the moment. 
* Looping doesn't appear to be functional.

AlexGStapleton:
That's odd. I just double-checked and both are working for me. Can you please clarify what you're seeing instead of the featured image? Nothing? Hm. Are you using a build or the direct repo?

Misplon:
The image is there but isn't displaying. I tested both with the repo and a build on the Corp demo.

Please, also check that the h3 styling is being properly targetted, we seem to have lost that.

Misplon:
The following appears to be missing:

```
.sow-carousel-wrapper ul.sow-carousel-items li.sow-carousel-item .sow-carousel-thumbnail a {
    display: block;
    background-position: center center;
}
```

AlexGStapleton:
`ul` and `li` aren't being used anymore so the CSS you referenced is outdated (this is an unavoidable - weird Slack limitation).

I suspect your CSS cache hasn't update due to a proper plugin update not occurring so it's using the old CSS. Can you please either manually clear your wp-content/uploads/siteorigin-widgets or recreate the Post Carousel?

Misplon:
Thanks. I'm not seeing the old rules, just letting you know that the rule necessary for the image to show in the `a` tag isn't there. Same with the `h3` tag.

Misplon:
Yeah, seems to be caching, images and h3 working now.

Misplon:
The `h3` issue is resolved [here](https://demo.siteorigin.com/corp/) but the image and looping issue persists. I've cleared AO. Perhaps take a look.

AlexGStapleton:
It looks like the widgets LESS isn't being generated. The LESS file is in place and it's the correct version (which is generating without issue for me locally) so I'm unsure why it's not generating. Are you able to replicate this locally?

https://demo.siteorigin.com/corp/wp-content/plugins/so-widgets-bundle/widgets/post-carousel/styles/default.less

Misplon:
We're progressing, Alex found an issue with multiple carousels on a page. The PR will be updated soon to resolve.

Misplon:
Thanks, Alex.

* It seems we're still stuck with https://github.com/siteorigin/so-widgets-bundle/issues/992. If a value isn't entered into the posts per page field, the count will default to 10.

* Scrolling on mobile appears to be pretty difficult. Please, test on mobile. We'll most likely need to adjust `edgeFriction` or is it `touchThreshold` ?. 

* Should we enable the Slick `rtl` setting?

Misplon:
For interest sake, perhaps look at `carousel_elem.slick` in Livemesh to see what they've done there with their setup.

AlexGStapleton:
1. Oh. I didn't know you wanted that resolved in this PR (thought it was going to be a follow-up PR). https://github.com/siteorigin/so-widgets-bundle/pull/1024/commits/a22687af1b0df51fd27c7b434a1e4834962e0636 resolves this.

2. I have doubled the threshold (5 is the default). Let me know your thoughts.

Livemesh doesn't adjust the threshold.

3. We would need to manually implement this as we're sidestepping the standard navigation system to allow for the post carousel to work how it currently does (see [this commit](https://github.com/siteorigin/so-widgets-bundle/pull/1024/commits/dafdcc511d120f164e51b7b43f42d7a197724a61) comment). Let me know if you would like me to proceed with this.

Misplon:
1. Fixed. Good work.
2. It's set to `10`, let's move it up to `12`.
3. When I switch to RTL, there are no posts visible. Let's get this working.
4. A hash is appearing when clicking the navigation buttons.
5. I think we should consider removing the select blue focus.

Misplon:
I think we should enable `swipeToSlide`. Try dragging with your mouse as it is now, it's frustrating, the 1 count is enforced. `swipeToSlide` lets the user drag to the item they want without enforcing the count.

AlexGStapleton:
2. Done.
3. Does RTL work in the release version? I tried to enable it to see how it works and it doesn't appear to be working for me. Regardless, I've set made the changes required to get Slick working.
4. Fixed.
5. I agree. I think the zoom is enough of an indicator without changing how the image itself looks. Let me know if you want me to proceed with removing this.
6. Done.

widgets/post-carousel/js/carousel.js
```diff
@@ -25,12 +26,15 @@ jQuery( function ( $ ) {
...
-				touchThreshold: 10,
+				touchThreshold: 12,
+				swipeToSlide: true,
+				rtl: direction == 'rtl',
 			} );
 
 			// click is used rather than Slick's beforeChange or afterChange 
 			// due to the inability to stop a slide from changing from those events
-			$widget.find( '.sow-carousel-previous, .sow-carousel-next' ).on( 'click', function() {
+			$widget.find( '.sow-carousel-previous, .sow-carousel-next' ).on( 'click', function(e) {
```
Misplon, line 36: `function(e)` ideally should be `function( e )`. Thanks.

Misplon:
Thanks

2. Looks good.
3. RTL does not work in `develop`. 

* The nav arrows need to change to the left for RTL. It's working otherwise.
* Try grabbing a post with your mouse when RTL is enabled, you can skim through dozens of posts with a flick, the behavior isn't the same as LTR. `rtl` isn't' an array like `responsive` so I'm not quite following why the settings are different. [Here](https://github.com/kenwheeler/slick/issues/3641) is basically the issue I'm describing. Please, take a look and confirm. It's frustrating this is happening but we might have to live with it if you don't see a fix.

4. Looks good.
5. Proceed, thanks.
6. Looks good.

7. I know we're adjusting the carousel by one post at the moment but I'd just like to preview what that might look like at 4.

For tablet and mobile, I think the reason it still feels frustrating is because we're only swiping by one so for those two, I'd definitely like to increase the post swipe count.

Can you take a look for us and see why those options aren't working? Do we perhaps have to make adjustments outside of Slick?

AlexGStapleton:
3a. Done. 
3b. Strange. This feels like a really complicated fix. I can look into this, but I don't think I'll be able to fix it without modifying Slick. Do you want me to proceed?
5. Done.
7. Looks like that unlike rtl, slidesToScroll doesn't affect slickPrev and slickNext if called directly.

You can try this by opening carousel.js and replace:

`$items.slick( 'slickNext' );`

With:
```
if ( $items.slick( 'slickCurrentSlide' ) + 4 > lastPosition ) {
	$items.slick( 'slickGoTo', lastPosition );
} else {
	$items.slick( 'slickGoTo', $items.slick( 'slickCurrentSlide' ) + 4 );
}
```
Replace:

`$items.slick( 'slickPrev' );`

With:
`$items.slick( 'slickGoTo', $items.slick( 'slickCurrentSlide' ) - 4 );`

The above code needs to be slightly cleaned up but it should allow you to get a feel for how it looks.

Misplon:
3b, we'll accept as a limitation.

7. This works but it doesn't seem to work for mobile which is really where I want to increase the count to be more than one. Do we have to write this ourselves? I want to try 3 at 1024 and 2 at 780.

AlexGStapleton:
7. By works on mobile, do you mean when they navigate using the arrows or touch? Just want to confirm we're on the same page.

widgets/post-carousel/tpl/carousel-post-loop.php
```diff
@@ -4,15 +4,16 @@
  * @var string $default_thumbnail
  */
 while($posts->have_posts()) : $posts->the_post(); ?>
-	<div class="sow-carousel-item">
+	<div class="sow-carousel-item<?php if( is_rtl() ) echo ' rtl' ?>">
```
Misplon, line 7: Small fix here `if( is_rtl() )` changed to `if ( is_rtl() )`

AlexGStapleton: That's actually no longer required so I'm just going to remove it.

Misplon: Thanks!

Misplon:
Yeah, the arrows respond to whatever count we set using the updated code you posted above so it's touch that we're looking to have the ability to control the count on.

AlexGStapleton:
Try setting touchThreshold to 1000 (or some other really high number). How does that feel? This should help make it feel better.

I'm not seeing a way of changing the touch placement without modifying the code. :(

Misplon:
I'm looking to have `https://github.com/siteorigin/so-widgets-bundle/blob/f4c14e3747f9b122ba229c8298ccdfedb1d5a28b/widgets/post-carousel/js/carousel.js#L32-L47` work. How can we do that?

Misplon:
@AlexGStapleton Please, can your resolve the conflict and we'll proceed from there.

Issue #1023, opened 2020-04-15
SO Simple Masonry Image: Add Preloader to mask images


AlexGStapleton:
The SiteOrigin Simple Masonry Image widget relies heavily on JavaScript to work. Before the JavaScipt is able to display, the images are disabled at full size. We should introduce a CSS preloader to mask this.

AlexGStapleton:
@Misplon Is there a specific design you had in mind for the preloader, or would you prefer all of the images to be hidden and only shown once the JavaScript is able to run? The latter may harm UX if the widget is using images with large file sizes and the download takes a significant amount of time.

Misplon:
There isn't a specific design I had in mind. Do you perhaps want to try to find an example of what you're thinking design-wise? I had a look around now but nothing is jumping out at me yet.

AlexGStapleton:
I typically use one from https://loading.io/css/ as they're quite simple to add. I think the circle (row 2 column 2) is fine. Thoughts?

Misplon:
Row 2 column 2, agreed.

![Pure_CSS_Loader_-_Optimized_Spinners_for_Web_·_Loading_io](https://user-images.githubusercontent.com/789159/79355014-6f952780-7f3d-11ea-8d00-73109110a9ee.png)

AlexGStapleton:
I've started work on this and I've realized a few settings need to be added. There's nowhere to really put them currently so I want to introduce a new group called Preloader (or maybe Settings) and add three settings:

- Show Preloader (certain people may not want it - up to you)
- Preloader Color
- Preloader Height
  * The size of the preloader prior to your Masonry Images showing

Are you okay with those names?

Misplon:
Let's call the section Preloader. Otherwise sounds good.

PR #1025, opened 2020-04-20
Simple Masonry Preloader

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1023

This PR adds a preloader to the Simple Masonry widget. This prevents an issue where the images display at full size before they've finished loading and then are resized into a Masonry.

Note: you'll need to clear wp-content/uploads/siteorigin-widgets or create a new widget due to the CSS cache not being cleared when plugin files are manually replaced.


Misplon:
Looks good, thanks, mate. 

Is there a border around the icon? If so, I think we can do away with that and just have the icon. Otherwise, all good.

AlexGStapleton:
The border is the icon. The icon is a serious of rounded borders that are moved.

PR #1032, opened 2020-05-02
Changed widget remove icon v-align to baseline

Misplon:
Currently, the repeater remove icon hover looks as follows:

hover

Note the alignment, it's correct. This PR corrects that issue. I'm not seeing any negatives from this change, can you, Alex?


Misplon:
I measured the icons and they are pixel-perfect centered now so this change is for the better.

PR #1033, opened 2020-05-10
Price Table: Add Hover Text using Title Attribute

AlexGStapleton:
Related: https://github.com/siteorigin/siteorigin-premium/pull/385

### DIFF
--- a/widgets/price-table/tpl/atom.php
+++ b/widgets/price-table/tpl/atom.php
@@ -38,7 +38,7 @@
 
             <div class="ow-pt-features">
 				<?php foreach ( $column['features'] as $i => $feature ) : ?>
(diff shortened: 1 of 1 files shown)


Misplon:
I ran this through the W3C Validator and didn't find any issues with the `title` tag being present when the title field is empty. Are we sure that's ok?

AlexGStapleton:
I just had a quick look over the WCAG and don't see any mention of empty title tags causing issues. I did it this way to make that line more readable, but if you would prefer it's not output if it's empty, let me know.

Misplon:
Ok cool, thanks for checking. This fine. If it gets raised then we can re-look at it.

Misplon:
Or if Greg wants it removed if empty.

gregpriday:
I'd personally have gone for leaving out the title tag if it's empty, but it is more readable like this and it doesn't make a huge difference.

Misplon:
Got you. Thanks for the feedback.

PR #1034, opened 2020-05-12
Testimonial: Resolve Alignment issues and stretch testimonials

### LINKED ISSUE #1035: Testimonial: User image isn't following user position
To recreate this issue set the testimonial user position to the right, note that the user image remains left-aligned. The user image should follow the user position setting.

AlexGStapleton:
Resolve #1035

This PR will prevent an issue where larger Testimonials on the previous row can cause alignment issues on the next. It will also stretch the testimonials so that they're all same height.

Here's a screenshot of this issue:

I had meant to split this commit into two but I had already gone too deep before I did. 😞

### DIFF
--- a/widgets/testimonial/css/style.less
+++ b/widgets/testimonial/css/style.less
@@ -1,16 +1,23 @@
 @import "../../../base/less/mixins";
 
 .sow-testimonials {
-	margin: -10px;
-	.clearfix();
+	align-self: stretch;
+	display: flex;
+	flex-wrap: wrap;
 
 	* {
 		box-sizing: border-box;
 	}
 
 	.sow-testimonial-wrapper {
-		float: left;
+		display: flex;
 		padding: 10px;
+
+		&.sow-layout-side {
+			.sow-testimonial, .sow-testimonial-text {
+				display: flex;
+			}
+		}
 	}
 
 	.sow-testimonial-user {
@@ -61,6 +68,10 @@
 	.sow-user-right {
 		.sow-testimonial-user {
 			text-align: right;
+		}
+
+		.sow-testimonial {
+			flex-direction: row-reverse;
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


Misplon:
Thanks, mate. There is an issue with the background. 

**Develop:**

dev

**This branch:**

branch

Sorry for the hassle but let's introduce a setting for stretching. So by default, let's just clear the rows. Under the Testimonial layout setting, insert a new checkbox:

Equalize testimonial height

AlexGStapleton:
> There is an issue with the background.

I can't replicate that look when using develop or the release version of the widgets bundle. I tried two different installs. I don't remember the background ever actually being behind the image and the changes to flex shouldn't have caused this type of issue.

Misplon:
Yeah, not sure how that's happening locally. I've checked here https://demo.siteorigin.com/corp/testimonial-test/. So just my settings suggestion then.

Misplon:
Thanks.

There seems to be a difference in the column width when equalize is enabled vs disabled.

Here is a test layout to check.

[page-builder.json.zip](https://github.com/siteorigin/so-widgets-bundle/files/4624408/page-builder.json.zip)

AlexGStapleton:
How does that look?

Misplon:
Thanks, that issue is resolved.

Check the following:

User position middle
Text below user
Equalize testimonial height enabled

The result is:

Page_Builder_-_SiteOrigin

Using Equalize is breaking the Testimonial layout.

Misplon:
Try User position right, the text above user, equalize disabled. There is an issue there.

AlexGStapleton:
That looks to be https://github.com/siteorigin/so-widgets-bundle/issues/1035
Did you want me to fix that in this PR? I was going to submit a new PR for that but if you think it works in this one I'll commit it in this one.



I've tried all 19 combinations and the only issues I'm seeing now is related to https://github.com/siteorigin/so-widgets-bundle/issues/1035

Misplon:
Thanks for checking all the combos. Yeah, let's hit #1035 in this PR.

PR #1050, opened 2020-05-28
Update Packery

AlexGStapleton:
This PR updates [Packery](https://github.com/metafizzy/packery) to v2.1.2. I ran some tests and didn't notice any issues with any of the selectable options. I ran a test release build to ensure that it still works as expected. The built version also worked as expected.


Misplon:
Thanks, mate. There is a version number in so-widgets-bundle.php.

Issue #1049, opened 2020-05-27
Simple Masonry: Add origin setting


Misplon:
Add a setting just before the preview button.

**Layout origin**
Controls the horizontal flow of the layout. Items can either start positioned on the left or right.

Select field.
Default: Left

PR #1051, opened 2020-05-29
Simple Masonry: Add Layout Origin setting

AlexGStapleton:
This PR resolves https://github.com/siteorigin/so-widgets-bundle/issues/1049. It requires https://github.com/siteorigin/so-widgets-bundle/pull/1050.
This PR implements [this Packery option](https://packery.metafizzy.co/options.html#originleft). 

The setting is internally known as layout_origin_left as the value needs to be a boolean and layouts_origin = true/false makes less sense than layouts_origin_left = true/false. While this setting could logically be a checkbox, I agree with the idea of using a select as it's more approachable and obvious what the outcome will be.

### DIFF
--- a/widgets/simple-masonry/js/simple-masonry.js
+++ b/widgets/simple-masonry/js/simple-masonry.js
@@ -64,7 +64,8 @@ jQuery( function ( $ ) {
 					$gridEl.packery( {
 						itemSelector: '.sow-masonry-grid-item',
 						columnWidth: columnWidth,
-						gutter: layout.gutter
+						gutter: layout.gutter,
+						originLeft: $gridEl.data( 'layout-origin-left' ),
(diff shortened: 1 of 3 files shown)


Misplon:
Thanks, mate.

Is your widget preview text showing true or false based on the value selected for this new setting? Mine seems to be.

AlexGStapleton:
I just had another look and I can confirm I'm getting that also. That's due to SiteOrigin Page Builder picking it as the "best" description. We can prevent this by excluding true/false as viable options (we already do this for "on") from the auto widget description. Do you want to do that, or do you want to change the setting to a checkbox (Layout Origin Left)?

Misplon:
Let's exclude `true` or `false`.

PR #1054, opened 2020-05-29
Move image attributes filter

andrewgillingham:
I propose moving the image attribute filter to allow alt/title attributes to be filtered as well.

### DIFF
--- a/js/lib/packery.pkgd.js
+++ b/js/lib/packery.pkgd.js
@@ -1,308 +1,182 @@
 /*!
- * Packery PACKAGED v1.4.3
- * bin-packing layout library
+ * Packery PACKAGED v2.1.2
+ * Gapless, draggable grid layouts
  *
  * Licensed GPLv3 for open source use
- * or Flickity Commercial License for commercial use
+ * or Packery Commercial License for commercial use
  *
  * http://packery.metafizzy.co
- * Copyright 2015 Metafizzy
+ * Copyright 2013-2018 Metafizzy
  */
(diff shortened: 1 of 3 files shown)


AlexGStapleton:
Hi Andrew,

Thank you for submitting a PR. :)
I like the idea of this change, but moving the apply_filter below where it currently is may result in issues for those relying on its current location as [modifications are run after this filter](https://github.com/siteorigin/so-widgets-bundle/blob/1.17.0/widgets/image/image.php#L161-L167) is triggered.

Adding a new filter for [the return](https://github.com/siteorigin/so-widgets-bundle/blob/1.17.0/widgets/image/image.php#L161-L167) will allow you to filter the alt/title attributes and link attributes. Do you want to submit a commit that does this? If not, I can submit one for you. I think it should be called `siteorigin_widgets_image_all_attrs` but I'm open to feedback.

Kind regards,
Alex

andrewgillingham:
If I didn't do it right, feel free to close this PR and do yourself - but I removed commit moving the image filter and added the new filter `siteorigin_widgets_image_all_attrs`. I tried to do it following your code formatting and the style of your other filters.

AlexGStapleton:
Hi Andrew,

Sorry, my first reply was a bit brief. Thanks for submitting the updated commit. Unfortunately, the new filter isn't in the correct location and the original filter isn't restored.
We really appreciate your support. I'll submit a PR shortly.

Kind regards,
Alex

Issue #1060, opened 2020-06-02
<iframe> is not getting saved in the textarea, text field and even the Code field


andrejarh:
Hi,
When I was trying to build a widget for responsive iframes (video, maps, etc..), I noticed that the textarea and even the text field, are filtering out the <iframe> HTML tag.

I had to use the editor field in order to prevent escaping the content. However the solution isn't optimal from the UX perspective (a lot of controls which are useless when dealing with iframe content).

Would it be possible to loosen up the escaping on textarea and text fields? 

A possible and safe solution could be a special field option, which would escape content differently (allow more HTML tags), but  would need to be defined on the widget level. This way default escaping stays the same to prevent any backwards compatibility issues.

AlexGStapleton:
Hi Andrejarh,

The text, tinymce, and textarea fields aren't really designed to handle HTML beyond basic formatting (em, strong, etc). I recommend either using  the [the code form field](https://siteorigin.com/docs/widgets-bundle/form-building/form-fields/#heading-code), or you could [create a custom form field](https://siteorigin.com/docs/widgets-bundle/form-building/adding-custom-fields/) with minimal sanitization to do this

Kind regards,
Alex

andrejarh:
I have tested the Code field, which seems like a great choice, but it also removes the <iframe> tag and doesn't save it.

PR #1062, opened 2020-06-03
Code form field: Remove Sanitization

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1060 

This will remove the standard HTML sanitization from occurring. It does not affect sanitization elsewhere but that sanitization isn't interesting with the code form field. Developers using this form field would already be escaping on output so backwards compatibility isn't an issue.


Misplon:
@AlexGStapleton Is this definitely work for you? It doesn't appear to be working for me.

`<iframe src="https://www.w3schools.com" title="W3Schools Free Online Web Tutorials"></iframe>`

AlexGStapleton:
I just tried the iframe HTML you posted and that saved for me. I also a fresh build of this PR and that worked also.

Are you using the code widget I sent you? That widget doesn't have a template as how it displays doesn't really matter, the only thing this PR is for is to ensure the iframe is able to be saved. Sorry, I should have mentioned this in the initial PR.

Misplon:
Got you, yeah, that's fine then.

PR #1067, opened 2020-06-10
Add font-display: block;

KevinTaron:
It think it would be a good idea to add font-display: swap to all font-faces.
Lighthouse / Pagespeed insights always wants me to at the font-display swap to get a better page speed.

Not huge change but could help.


KevinTaron:
@AlexGStapleton based on your descriptions it makes sense to use block.
Therefore i changed the PR to block.

AlexGStapleton:
@KevinTaron The font awesome file appears to be vastly different to [the develop version](https://github.com/siteorigin/so-widgets-bundle/blob/develop/icons/fontawesome/style.css). What version/branch is your PR based on? I ask as I tried to resolve the conflicts with develop but it was going to result in changes to that branch rather than this one and that's not really ideal due to it messing up the blame.

KevinTaron:
uhh probably its from the master branch.
I insert the style.css for fontawesome from develop but i'm not sure if thats correct.
If not maybe just ignore the PR and add the font-display direct?

PR #1069, opened 2020-06-12
SiteOrigin Widgets Block: Add Block CSS Classes

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1068


Misplon:
@AlexGStapleton I made two tiny changes otherwise looks good from my end.

PR #1073, opened 2020-06-15
Prevent previewing issues for widgets that use Google Fonts

AlexGStapleton:
Resolves https://github.com/siteorigin/siteorigin-panels/issues/732

I do not see any issues with removing this line and removing it results in this issue no longer occurs. Do you see any @Misplon?

I can't when exactly this issue was introduced. I tried a few older versions of the SiteOrigin Widgets Bundle and this issue is present. I suspect this issue may have always been present as it's not clear why `siteorigin_widgets_inline_styles` was being cleared.

### DIFF
--- a/base/base.php
+++ b/base/base.php
@@ -37,8 +37,6 @@ function siteorigin_widget_print_styles(){
             <style type="text/css"><?php echo($widget_css) ?></style><?php
         }
     }
-
-	$siteorigin_widgets_inline_styles = array();
 }
 add_action('wp_head', 'siteorigin_widget_print_styles');
 add_action('wp_footer', 'siteorigin_widget_print_styles');


AlexGStapleton:
I found this issue the easiest to replicate while using the Block Editor.

Misplon:
I'm having trouble re-creating the issue in the Block Editor. Any ideas why that may be? I've tried other installs, our demo server, etc.

AlexGStapleton:
Honestly, this issue can be really inconsistent. If you want to avoid the prebuilt layout, try adding a series of SiteOrigin Headline widgets each with a different Google Font. The issue is triggered by the other inline CSS being cleared.

Misplon:
This change resolves https://github.com/siteorigin/siteorigin-panels/issues/732. @gregpriday Can you see any issues with this change?

base/base.php
```diff
@@ -37,8 +37,6 @@ function siteorigin_widget_print_styles(){
             <style type="text/css"><?php echo($widget_css) ?></style><?php
         }
     }
-
-	$siteorigin_widgets_inline_styles = array();
```
gregpriday, line 41: I'm not sure what removing this will do. The issue is that this function is called once in the header and once again in the footer. Technically, any inline styles added before `wp_head`, will be outputted twice. That shouldn't be an issue, but it's quite inefficient.

AlexGStapleton: Is there a specific reason why it's output for both? Based on what you mentioned the actual fix may actually be just to output it once using either wp_head or wp_footer.

gregpriday: It's hard to remember the reason, but sometimes the widgets aren't processed before `wp_head`. I think one of the cases is widgets being in the sidebar and widgets being in other builders.

We wouldn't want to penalise widgets that have pre-processed by only putting their CSS in the footer. This will cause a visual jump. Technically though, having this `$siteorigin_widgets_inline_styles = array()` clearing line should be doing the trick.

If we're just trying to solve a previewing issue, then let's rather target previewing states rather.

AlexGStapleton: I've had another look and worked out what's going on. Please note that the previous title of this issue is wrong and not correct - I'm going to rename the issue based on your reply. This issue is actually pretty complicated.

The We Love Design layout includes a Hero widget that has a button but it doesn't contain the [buttons]. This is normally _fine_ but due to [this line](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/siteorigin-widget.class.php#L243), each widget in a preview instance will output its CSS after the widget has been rendered. Due to the button not outputting and the CSS being output after every widget, the CSS for both the button and the Hero widget is never output due to it being cleared - this is why removing `$siteorigin_widgets_inline_styles = array()` helped.

I've submitted a PR here that confirms [buttons] is present in the Hero frame, before attempting to render them.

I think we should prevent the output of [widget inline CSS after widgets](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/siteorigin-widget.class.php#L243) for widgets added using the SiteOrigin Layout Block, Live Editor, and WP Preview and instead rely on wp_footer to output.  This will prevent this type of issue from happening for any other sub widget - although admittedly it is quite a situational issue. Do you agree, or are you okay with keeping just the Hero widget PR?

Please note, that the SiteOrigin Layout Block CSS is already output [here](https://github.com/siteorigin/siteorigin-panels/blob/ddd2733c4001a22515e8ff77d5817fd7705d5ab0/inc/admin.php#L1196-L1202) so it's not relying on [this line](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/siteorigin-widget.class.php#L243).

AlexGStapleton:
The current title doesn't really make sense anymore, but it applies to what I had thought the issue was at the time of the current commit. I'm going to rename this PR after we decide what to do with it.

Misplon:
@AlexGStapleton Do you perhaps want to close this PR and then log an issue for what you'd like to do next.

Issue #1074, opened 2020-06-16
Features: Ensure the icon doesn't overlap the text


Misplon:
Ensure the icon doesn't overlap the text. To replicate, set the icon to the left. Let's test all layout possibilities.

![SiteOrigin_–_Just_another_WordPress_site](https://user-images.githubusercontent.com/789159/84772297-0a13f480-afdb-11ea-90da-807a48d4c783.jpg)

[home-page.json.zip](https://github.com/siteorigin/so-widgets-bundle/files/4786451/home-page.json.zip)

Misplon:
The issue we ran into with the linked PR is it requires the **Use icon size for custom icon** setting to be enabled. If this setting isn't enabled, the custom icon added won't receive a width and height and therefore won't be visible. Options that we've considered today are:

* Force this setting. All users that aren't currently using this setting and therefore haven't adjusted the icon size as required will have their icon changed in a way that isn't desirable. I don't think we should do this.

* Duplicate the complete old Less (pretty much all of it would be required) and add a dedicated legacy stylesheet that will potentially result in overlaps but won't affect custom images with no size set.

PR #1075, opened 2020-06-17
Features: Prevent Text Overlapping Icon

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1074

Removing this CSS allows for the icon and text to convert each other when sizing. This means that the features text may be slightly smaller as a result of this change due to the icon pushing it over.

### DIFF
--- a/widgets/features/styles/default.less
+++ b/widgets/features/styles/default.less
@@ -91,34 +91,46 @@
         }
         
         .sow-icon-container {
-            width: @container_size;
-            height: @container_size;
-            font-size: @container_size;
             text-decoration: none;
 
             & when not ( @per_row = 1 ) {
                 margin: auto;
             }
 
-            [class^="sow-icon-"],
-            .sow-icon-image {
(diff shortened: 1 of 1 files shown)


Misplon:
Is there any way we can handle certain icons that overlap the icon container?

SiteOrigin_–_Just_another_WordPress_site

Misplon:
Please, try the following layout. The overlap issue seems to be persisting. 

[home-page.json.zip](https://github.com/siteorigin/so-widgets-bundle/files/4792364/home-page.json.zip)

AlexGStapleton:
Can you please show me a screenshot of what you're seeing? This is what I'm getting:

Screenshot_2020-06-18 - SiteOrigin(1)

> Is there any way we can handle certain icons that overlap the icon container?
The last commit resolves this. It also resolves a display issue with container shapes not working correctly.

Misplon:
The last commit resolves the issue, thanks.

PR #1079, opened 2020-06-26
Vertically align Post Carousel nav buttons if a title is present

Misplon:
@AlexGStapleton The goal of this PR is to better vertically align the carousel nav buttons when a title is present. 

Request: https://wordpress.org/support/topic/post-carousel-in-1-17/page/2/#post-13038229

As far as I'm aware, within CSS3, we don't have a way of counting children and then applying styles to the parent based on the count. Ideally, I'd apply Flexbox when two children are present within the `.sow-carousel-title` div (the title and the nav container). Counting in this manner will be included in CSS4.

I've set this PR to draft status. Let me know if you'd prefer to go in another direction.

### DIFF
--- a/widgets/post-carousel/css/style.less
+++ b/widgets/post-carousel/css/style.less
@@ -39,6 +39,21 @@
 }
 
 .sow-carousel-title {
+
+	&.has-title {
+		align-items: baseline;
+		display: flex;
+
+		.sow-carousel-navigation {
+			margin-left: auto;
+		}
+
+		body.rtl & .sow-carousel-navigation {
+			margin-right: auto;
+			margin-left: initial;
+		}
+	}
+
 	.widget-title {
 		display: inline-block;
 		padding-right: 15px;
--- a/widgets/post-carousel/tpl/base.php
+++ b/widgets/post-carousel/tpl/base.php
@@ -11,7 +11,7 @@
 ?>
 
 <?php if(! empty( $posts ) && $posts->have_posts() ) : ?>
-	<div class="sow-carousel-title">
+	<div class="sow-carousel-title<?php if ( ! empty( $title ) ) echo ' has-title'; ?>">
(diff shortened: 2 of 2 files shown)


AlexGStapleton:
What are your thoughts on the following CSS instead? (please try it against develop) It does remove the title margin but it doesn't require an template adjustment.

```
.sow-carousel-title {
	display: flex;
	justify-content: space-between;
	align-items: center;
}

.sow-carousel-title .widget-title {
	margin-bottom: 0;
}
```

Regardless, your changes work so if you would rather stick with the current setup that's fine.

Misplon:
Thanks. 

Have you tested the above change without a title present?

AlexGStapleton:
Ah, you're right. Try the following CSS instead:

```
.sow-carousel-navigation {
	display: flex;
	flex-direction: row-reverse;
	align-self: center;
}

.sow-carousel-title .widget-title {
	margin-bottom: 0;
}
```

Misplon:
Thanks, that solves the nav with no title issue but doesn't look to resolve the horizontal alignment issue.

Misplon:
Screenshot 2020-06-28 at 19 59 19

If you get the same result, let's consider moving on.

AlexGStapleton:
What horizontal alignment issue? Did you mean vertical?

I get the following in Firefox when using the CSS I included in my last reply:



Chrome:

Misplon:
Yeah, I mean vertical. How is that happening though? How is the nav filling the height of the title? To confirm, you've added those two rules to develop, nothing else?

Misplon:
I don't know that we should remove the margin from the title though. Tough one...

AlexGStapleton:
> How is that happening though? How is the nav filling the height of the title? How is the nav filling the height of the title?

`align-self: center;` centers the element within the parent.

> To confirm, you've added those two rules to develop, nothing else?

Correct.

AlexGStapleton:
Honestly, I'm okay with either method. I'm mainly suggesting an alternative method as you asked. I'm not suggesting either is superior.

PR #1086, opened 2020-07-12
Post Carousel: Disable scroll if number of visible posts is more than total posts

AlexGStapleton:
This PR checks if the number of visible items is larger than the total number of posts on load and after a resize.

### DIFF
--- a/widgets/post-carousel/js/carousel.js
+++ b/widgets/post-carousel/js/carousel.js
@@ -110,6 +110,23 @@ jQuery( function ( $ ) {
 				}
 			} );
 
+			// Hide/disable scroll if number of visible items is less than total posts.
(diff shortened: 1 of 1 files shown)


Misplon:
Looks good, mate, thanks. Is there anything we can do about the nav icons loading and then being removed?

AlexGStapleton:
Not really. The code has to run after Slick has inititalized, and there's always going to be a brief moment during the load where it's yet to run. I ran some tests to see if there were any hooks that ran prior to `load` in the hopes and I wasn't able to find any that made this delay any less noticeable.

We could do the inverse (hide and then selectively show navigation) but that feels less ideal.

Misplon:
Thanks for the update, all sounds good.

Issue #1087, opened 2020-07-15
Mobile: Adjust form field widths to fit


Misplon:
Look at adjusting form field widths to fit on mobile devices where possible.

PR #1088, opened 2020-07-15
WB Form Field Mobile Adjustments

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1087

### DIFF
--- a/base/inc/fields/css/date-range-field.less
+++ b/base/inc/fields/css/date-range-field.less
@@ -28,6 +28,7 @@
 
 		input[type="text"].datepicker {
 			width: 258px;
+
 			@media (max-width: 680px) {
 				width: 100%;
 			}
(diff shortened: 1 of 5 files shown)


base/inc/fields/css/date-range-field.less
```diff
@@ -14,6 +28,9 @@
 
 		input[type="text"].datepicker {
 			width: 258px;
+			@media (max-width: 680px) {
```
Misplon, line 31: Can we perhaps add new line before the media query?

Misplon:
Thanks, mate. Is this PR working for you in the Hero widget? The FitText compressor strength field for example.

AlexGStapleton:
Can you please show me a screenshot of what you're seeing? I just had a look at that field and it looks to be working as expected. It's possible I misinterpreted what you meant by full width and/or this wasn't one of the fields you wanted to be full width.

Misplon:
I have this build running on the demo server so you can check the Corp home
page Hero for example.

Misplon:
Edit_Page_‹_Corp_—_WordPress

AlexGStapleton:
@Misplon Oh. I was looking at a higher resolution so I didn't see that. I can replicate this and I'll fix this now.

AlexGStapleton:
How does that look?

Issue #1028, opened 2020-04-27
Contact: Allow input to be used in the success message


Misplon:
The Editor widget successfully outputs:

`<input type="submit" value="Submit" />`

When this same `input` is used in the contact form success message, it isn't correctly parsed and is outputted as code. If possible, allow for an `input` to be used in the success message. The use case here was a PayPal donation button.

PR #1090, opened 2020-07-16
Contact: Don't double sanitize success message

AlexGStapleton:
This PR resolves https://github.com/siteorigin/so-widgets-bundle/issues/1028.

Neither `wp_kses_post` nor `wpautop` are required here as they're already [run when the page saves](https://github.com/siteorigin/so-widgets-bundle/blob/e0deee4e80c6c27a5e944e22f814a7443569b389/base/inc/fields/tinymce.class.php#L504-L516).

### DIFF
--- a/widgets/contact/tpl/default.php
+++ b/widgets/contact/tpl/default.php
@@ -10,7 +10,7 @@ if( $result['status'] == 'success' ) {
 	// Display the success message
 	?>
 	<div class="sow-contact-form-success" id="contact-form-<?php echo esc_attr( $short_hash ) ?>">
-		<?php echo $instance['settings']['success_message'] ); ?>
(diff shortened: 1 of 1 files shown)


widgets/contact/tpl/default.php
```diff
@@ -10,7 +10,7 @@
 	// Display the success message
 	?>
 	<div class="sow-contact-form-success" id="contact-form-<?php echo esc_attr( $short_hash ) ?>">
-		<?php echo wp_kses_post( wpautop( $instance['settings']['success_message'] ) ) ?>
+		<?php echo $instance['settings']['success_message'] ); ?>
```
Misplon, line 13: The `)` will cause an error. Thanks.

Issue #1089, opened 2020-07-16
Post Carousel: Prevent current post from being included in returned posts


AlexGStapleton:
The Post Loop widget prevents the current post from appearing in the post loop widget, and the Post Carousel should do the same.

PR #1091, opened 2020-07-16
Post Selector: Add optional parameter to exclude current post

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1089

### DIFF
--- a/base/inc/post-selector.php
+++ b/base/inc/post-selector.php
@@ -4,7 +4,7 @@
  * Filter a query created from the post selector field into an array that will work properly with get_posts
  *
  * @param $query
- * @param bool $exclude_current whether to excldue the current post or not.
+ * @param bool $exclude_current whether to exclude the current post or not.
  *
  * @return mixed
  */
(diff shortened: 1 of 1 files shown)


base/inc/post-selector.php
```diff
@@ -98,6 +99,10 @@ function siteorigin_widget_post_selector_process_query( $query ){
 		}
 	}
 
+	if ( $exclude_current && get_the_ID() ) {
+		$query['post__not_in'][] = get_the_ID();
```
gregpriday, line 103: Are there situations where we'd need to initialise `$query['post__not_in']` to an empty array if it's not set yet?

AlexGStapleton: I wasn't able to come across any. I ran a few tests where `post__not_in` was, and wasn't used and didn't run into any issues/errors/warnings. I can initialize it if you would prefer.

gregpriday: Initialising is always slightly safer, so I'd say yes.

PR #1094, opened 2020-07-17
Social Media Buttons: Add Tiktok

AlexGStapleton:
Require https://github.com/siteorigin/so-widgets-bundle/pull/1093

### DIFF
--- a/widgets/social-media-buttons/data/networks.php
+++ b/widgets/social-media-buttons/data/networks.php
@@ -1,6 +1,103 @@
 <?php
 
 return array(
+	'facebook'    => array(
+		'label'    => __( 'Facebook', 'so-widgets-bundle' ),
+		'base_url' => 'https://www.facebook.com/',
+		'icon_color' => '#FFFFFF',
+		'button_color' => '#3A5795'
+	),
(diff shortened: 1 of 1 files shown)


Misplon:
Thanks.

Are you sure you want the network so high up? I'd be ok with placing it alphabetically.

AlexGStapleton:
I don't mind. I put it there because of the comment here.

https://github.com/siteorigin/so-widgets-bundle/blob/831ceaf61ee84ef330aa8fc8d47f7972263cc884/widgets/social-media-buttons/data/networks.php#L101

Going to re-order them now.

gregpriday:
The original ordering was to put the most common networks first, and the rest in alphabetical order. Lets keep the more popular networks first, and just put TikTok in the alphabetical list. We can also move Google Plus down into the alphabetical list, since that's dead.

gregpriday:
Actually, we should reassess that entire "most popular" list at some point. For this PR though, just move TikTok into the alphabetical list and keep the popular list unchanged.

Misplon:
@gregpriday Super, that's what I was asking for. Thanks.

Issue #1021, opened 2020-04-14
Accordion: Allow line breaks to be used in the Text tab


Misplon:
The Accordion widget isn't currently managing line breaks in the same way the Editor widget does. In the Editor widget, if you add in the Text tab:

```
Example

Example
```

Your line breaks will be maintained.

In the Accordion widget, it's only possible to do this in the Visual tab. If you switch to the Text tab, your line breaks will be lost.

PR #1097, opened 2020-07-20
Accordion allow line breaks dont double escape

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1021

### DIFF
--- a/widgets/accordion/accordion.php
+++ b/widgets/accordion/accordion.php
@@ -62,6 +62,12 @@ class SiteOrigin_Widget_Accordion_Widget extends SiteOrigin_Widget {
 					'content_text' => array(
 						'type'  => 'tinymce',
 						'label' => __( 'Content', 'so-widgets-bundle' ),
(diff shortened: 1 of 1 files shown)


widgets/accordion/accordion.php
```diff
@@ -226,7 +232,7 @@ public function get_template_variables( $instance, $args ) {
 	}
 	
 	public function render_panel_content( $panel, $instance ) {
-		$content = wp_kses_post( $panel['content_text'] );
```
gregpriday, line 229: Here again, are you sure we can remove this? Is filtering/sanitization handled by `wpautop`?

AlexGStapleton: No. [Sanitization is handled here](https://github.com/siteorigin/so-widgets-bundle/blob/1.17.3/base/inc/fields/tinymce.class.php#L509-L513).

gregpriday: Perfect!

PR #1100, opened 2020-07-21
Post Carousel: Improve Multiple Carousel and Mobile Support

AlexGStapleton:
This PR improves support for multiple post carousels on the same page but setting variables as needed. This ensures they're correct and relevant when used.

### DIFF
--- a/widgets/post-carousel/js/carousel.js
+++ b/widgets/post-carousel/js/carousel.js
@@ -38,9 +38,14 @@ jQuery( function ( $ ) {
 				],
 			} );
 
+			// Trigger navigation click on swipe
+			$items.on( 'swipe', function( e, slick, direction ) {
+				$$.parent().parent().find( '.sow-carousel-' + ( direction == 'left' ? 'next' : 'prev' ) ).trigger( 'touchend' );
+			} );
+
 			// click is used rather than Slick's beforeChange or afterChange 
 			// due to the inability to stop a slide from changing from those events
-			$$.parent().parent().find( '.sow-carousel-previous, .sow-carousel-next' ).on( 'click', function( e ) {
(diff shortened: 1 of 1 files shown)


Misplon:
@AlexGStapleton Please, check, can you scroll past 10 posts on mobile when posts per page isn't set?

Misplon:
I've confirmed locally, on mobile, it isn't possible to scroll past 10 posts. My test layout:

[24785.json.zip](https://github.com/siteorigin/so-widgets-bundle/files/4955764/24785.json.zip)

This layout uses the latest [unit test theme data](https://codex.wordpress.org/Theme_Unit_Test). The categories I've used are `category: block` and `cateogry: classic`, the only two categories I have with more than 10 posts.

Misplon:
@AlexGStapleton There is an edge case where the 10 hard (can't scroll past) limit is being applied. Resize your browser as small as it'll go and then grab posts with your mouse cursor, I can't get past 10 in that situation.

widgets/post-carousel/js/carousel.js
```diff
@@ -23,7 +17,7 @@ jQuery( function ( $ ) {
 				arrows: false,
 				infinite: false,
 				rows: 0,
-				rtl: direction == 'rtl',
+				rtl: $$.attr( 'dir' ) == 'rtl',
```
gregpriday, line 20: I missed this before, but shouldn't this be a data attribute? So `data-dir`?

Misplon: Resolved via https://github.com/siteorigin/so-widgets-bundle/commit/f0ae7eed63c7f5a17e2e19070f98a9cdaf6cfdf2.

PR #1107, opened 2020-07-23
Post carousel: Contextualize variables as needed

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/pull/1100#issuecomment-662064082 https://github.com/siteorigin/so-widgets-bundle/pull/1100#discussion_r459005793

This PR also finds the ItemWidth as needed rather than pre-defining it.


Misplon:
@AlexGStapleton Is looping working for you?

Issue #1101, opened 2020-07-21
Sliders: Add a setting to disable autoplay


Misplon:
Alex, do you want to add this to all three sliders? Can we make the timeout and animation settings appear when this checkbox is enabled and hide when disabled?

**Autoplay**
Change slides automatically without user interaction.

Setting type: Checkbox
State: Default enabled

Issue #982, opened 2020-02-05
Layout Slider: Pause on hover


Misplon:
Consider adding a new setting to the Layout Slider to pause on hover.

AlexGStapleton:
How does this sound?

Label: Autoplay pause on hover
Description: If enabled, the slider won't change slide while the user is hovering on the slider.

Misplon:
The label is good. Do you want to try this one without a description? We're using the word slider a few times in the proposed description. An alternative is perhaps "Pause autoplay while hovering over the slider." That's pretty similar to the label. Should we try with just the label?

AlexGStapleton:
Let's try with just the label.

Misplon:
Sounds good, thanks.

Issue #920, opened 2019-11-24
Hero Widget: Pause while focussed new setting proposal


Misplon:
Consider adding a setting to the Hero that will pause the slider when a user is focussed on a slide. This will be useful for users wanting to add a form to their slider (newsletter etc.) but still have multiple frames/slides. Once implemented the same setting could be extended to the Layout Slider widget.

AlexGStapleton:
This was resolved by https://github.com/siteorigin/so-widgets-bundle/pull/1109

PR #1109, opened 2020-07-26
Sliders: Add Autoplay and Autoplay Pause on Hover settings

AlexGStapleton:
This PR resolves https://github.com/siteorigin/so-widgets-bundle/issues/1101 https://github.com/siteorigin/so-widgets-bundle/issues/982 https://github.com/siteorigin/so-widgets-bundle/issues/920

### DIFF
--- a/base/inc/widgets/base-slider.class.php
+++ b/base/inc/widgets/base-slider.class.php
@@ -62,6 +62,16 @@ abstract class SiteOrigin_Widget_Base_Slider extends SiteOrigin_Widget {
 					),
 				),
 			),
+
+			'autoplay_hover' => array(
+				'type' => 'checkbox',
+				'label' => __( 'Autoplay pause on hover', 'so-widgets-bundle' ),
+				'default' => false,
+				'state_handler' => array(
+					'autoplay[autoplay]' => array( 'show' ),
+					'autoplay[static]' => array( 'hide' ),
+				),
+			),
 			'speed' => array(
 				'type' => 'number',
 				'label' => __('Animation speed', 'so-widgets-bundle'),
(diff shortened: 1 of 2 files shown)


AlexGStapleton:
The last PR will prevent the Animation Speed setting from being ever hidden as it affects both the autoplay and non-autoplay settings.

Misplon:
Thanks. Looks like this PR is currently removing Timeout when Autoplay isn't enabled. Timeout is currently a setting available in develop.

AlexGStapleton:
That was done due to the request made in #1101

> Can we make the **timeout** and animation settings appear when this checkbox is enabled and **hide when disabled**?

Animation speed is used regardless of autoplay, but timeout only affects the slider if autoplay is enabled so it makes sense to hide it if autoplay is disabled. Regardless, if you would prefer that timeout is always visible let me know and I'll remove the state emitter.

Misplon:
Thanks. If the timeout setting is only used when autoplay is enabled then we good.

base/inc/widgets/base-slider.class.php
```diff
@@ -162,6 +179,7 @@ function slider_settings( $controls ){
 			'pagination'               => true,
 			'speed'                    => empty( $controls['speed'] ) ? 1 : $controls['speed'],
 			'timeout'                  => $controls['timeout'],
+			'paused'                 => empty( $controls['autoplay'] ) ?: false,
```
Misplon, line 182: There is perhaps an alignment issue, a space missing.

Misplon:
Thanks.

In the Slider widget, try setting timeout to 1000 and enable pause on hover. Is pause on hover working for you, it isn't currently for me.

Misplon:
Do you want to add a slide animation or some animation when hiding and showing the setting?

AlexGStapleton:
> In the Slider widget, try setting timeout to 1000 and enable pause on hover. Is pause on hover working for you, it isn't currently for me. 

I just created a build to check is if this is working as expected and it is for. Here's a video of it working for me:

https://drive.google.com/uc?id=1GMTXm_3v8mdHJba0iStSNluevI2NEZK5

(please excuse the mouse placement - that's due to display scaling)

[Here's the build I tried](https://drive.google.com/uc?id=1GiHFkKws-hV1A-1xC8c-Pm6Q1ZQHMpSl).

> Do you want to add a slide animation or some animation when hiding and showing the setting?

Sure. That sort of change should however be in a separate PR as that change will affect all state emitters.

Misplon:
Please, log the state emitter transition task and self-assign. Thanks.

Thanks for the build, all looks good in there.

PR #1110, opened 2020-07-26
Features: Allow shortcodes to be used in the text field

AlexGStapleton:
This PR allows shortcodes to be added to the Feature text field.

### DIFF
--- a/widgets/features/tpl/default.php
+++ b/widgets/features/tpl/default.php
@@ -50,7 +50,7 @@ $last_row = floor( ( count($instance['features']) - 1 ) / $instance['per_row'] )
 					<?php endif; ?>
 
 					<?php if(!empty($feature['text'])) : ?>
-						<?php echo wp_kses_post( $feature['text'] ) ?>
+						<?php echo $feature['text']; ?>
 					<?php endif; ?>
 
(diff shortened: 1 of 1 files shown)


widgets/features/tpl/default.php
```diff
@@ -50,7 +50,7 @@ class="sow-icon-container <?php echo !empty($instance['container_shape']) ? 'sow
 					<?php endif; ?>
 
 					<?php if(!empty($feature['text'])) : ?>
-						<?php echo wp_kses_post( $feature['text'] ) ?>
```
gregpriday, line 53: Do we not need both `do_shortcode` and `wp_kses_post` in here?

Misplon: I've created double-up by asking about this on Slack and not GH, my bad. Alex replied as follows:

I asked: _Are the contents of `do_shortcode` run through `wp_kses_post`?_ For example should `script` be filtered out?

Alex replied: It depends on the role/capabilities of the account you're using as we do a unfiltered_html check.
https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/inc/fields/tinymce.class.php#L509-L513

gregpriday: Perfect, works for me.

Issue #1112, opened 2020-07-29
Contact Form: Ensure compatibility with Akismet


Misplon:
Via Twitter:

@SiteOrigin
 : looks like your Contact Form plugin has a bug that impairs its spam-fighting abilities: when sending data to the Akismet API, content should be sent as `comment_content`, not `comment_text`. We've handled it, for now, so your users get better results.

stephdau:
[There it is](https://github.com/siteorigin/so-widgets-bundle/blob/develop/widgets/contact/contact.php#L1232). :)

Misplon:
Thanks for taking a look Stephane, appreciate the help. We'll get this resolved shortly and included in the next update.

PR #1113, opened 2020-07-29
Contact: Pass correct fields to Akismet

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1112

I ran a few tests and didn't come across any issues by making this change.

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -1229,7 +1229,7 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 				$message_text[] = $m['value'];
 			}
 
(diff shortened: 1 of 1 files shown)


widgets/contact/contact.php
```diff
@@ -1229,7 +1229,7 @@ function spam_check( $post_vars, $email_fields, $instance ) {
 				$message_text[] = $m['value'];
 			}
 
-			$comment['comment_text']         = $email_fields['subject'] . "\n\n" . implode( "\n\n", $message_text );
+			$comment['comment_content']         = $email_fields['subject'] . "\n\n" . implode( "\n\n", $message_text );
```
Misplon, line 1232: Just a small alignment issue here but otherwise all looks good.

PR #1115, opened 2020-08-03
Post Carousel: Add Keyboard Navigation

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1108

- The first item is the default tab index.
- Upon tabbing to tab index slide, use left and right keyboard arrows to navigate to the next/previous item
- The tab index is the currently active item
- Add keyboard navigation for post carousel arrows by pressing Enter or Space. Will navigate entire carousel slides rather than item by item.

`build:dev` is required due to LESS changes.

### DIFF
--- a/widgets/post-carousel/js/carousel.js
+++ b/widgets/post-carousel/js/carousel.js
@@ -43,7 +43,7 @@ jQuery( function ( $ ) {
 
 			// click is used rather than Slick's beforeChange or afterChange
 			// due to the inability to stop a slide from changing from those events
-			$$.parent().parent().find( '.sow-carousel-previous, .sow-carousel-next' ).on( 'click touchend', function( e ) {
+			$$.parent().parent().find( '.sow-carousel-previous, .sow-carousel-next' ).on( 'click touchend', function( e, refocus ) {
 				e.preventDefault();
 				$items = $$.find( '.sow-carousel-items' );
 				var numItems = $items.find( '.sow-carousel-item' ).length,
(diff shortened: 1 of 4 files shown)


Misplon:
Thanks, mate.

* When using enter to scroll I can click the enter button six times I think before focus moves away from the carousel. Are you able to replicate?

AlexGStapleton:
How does that look?

I think we should add a setting in a future PR that allows you to change the navigation background/icon color and hover colors. The current hover color is honestly really hard to see.

Misplon:
> I think we should add a setting in a future PR that allows you to change the navigation background/icon color and hover colors. The current hover color is honestly really hard to see.

For sure, please, proceed.

widgets/post-carousel/js/carousel.js
```diff
@@ -106,7 +110,51 @@ jQuery( function ( $ ) {
...
+
+		// Keyboard Navigation of carousel items.
+		$( document ).on( 'keyup', '.sow-carousel-item', function( e ) {
+			// Ensure left/right key was pressed
+			if ( e.keyCode != 37 && e.keyCode != 39 ) {
+				return;
+			}
+
+			var $wrapper =  $( this ).parents( '.sow-carousel-wrapper' ),
+			$items = $wrapper.find( '.sow-carousel-items' ),
```
gregpriday, line 132: For clarity, I always prefer it if these are indented one level. Not sure what the WordPress style guide says here though.

AlexGStapleton: The Style guide actually recommends the spacing also. 😞

PR #1116, opened 2020-08-04
1.17.4

Misplon:
* Post Carousel: Improved support for multiple carousel widgets on the same page.
* Post Carousel: Improved mobile support.
* Sliders: Added `Autoplay` and `Autoplay Pause on Hover` settings to the Slider, Layout Slider and Hero widgets.
* Features: Allowed shortcodes to be used in the Text field.
* Contact Form: Fix to ensure the correct fields are passed to Akismet.


base/inc/widgets/base-slider.class.php
```diff
@@ -49,18 +49,39 @@ function initialize() {
 	 */
 	function control_form_fields(){
 		return array(
+			'autoplay' => array(
+				'type' => 'checkbox',
+				'label' => __( 'Autoplay', 'so-widgets-bundle' ),
+				'description' => __( 'Change slides automatically without user interaction.', 'so-widgets-bundle' ),
+				'default' => true,
```
Misplon, line 56: @AlexGStapleton Was this perhaps a mistake? Autoplay should most likely be defaulted to false.

Misplon:
@AlexGStapleton Try adding a carousel widget with more than 10 posts. Enter a low Posts per page value like say `2`. The nav arrows are being hidden.

base/inc/widgets/base-slider.class.php
AlexGStapleton: `Autoplay` set to `true` is how the slider currently works. `false` is new.

 Setting it to false will prevent the slide from automatically changing and will result in a lot of users reporting the slider broke because slides no longer transition automatically anymore.

Misplon: Perfect.

PR #1117, opened 2020-08-05
State Emitter: show/hide animation

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1114

### DIFF
--- a/base/js/admin.js
+++ b/base/js/admin.js
@@ -140,9 +140,15 @@ var sowbForms = window.sowbForms || {};
 										$$f = $$;
 									}
 
-									// Call the function on the wrapper we've selected
-									$$f[thisHandler[i][0]].apply($$f, typeof thisHandler[i][2] !== 'undefined' ? thisHandler[i][2] : []);
-									
+									if ( thisHandler[i][0] == 'show' ) {
+										$$f.slideDown( 'fast' );
+									} else if( thisHandler[i][0] == 'hide' ) {
+										$$f.slideUp( 'fast' );
+									} else {
+										// Call the function on the wrapper we've selected
+										$$f[ thisHandler[i][0] ].apply( $$f, typeof thisHandler[i][2] !== 'undefined' ? thisHandler[i][2] : [] );
+									}
+
 									if ( $$f.is( '.siteorigin-widget-field:visible' ) ) {
 										if ( $$f.is( '.siteorigin-widget-field-type-section' ) ) {
 											var $fields = $$f.find( '> .siteorigin-widget-section > .siteorigin-widget-field' );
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
@Misplon Let me know if you would like me to adjust the speed of the animation. I used fast as having it set to `slow` made the animation look a little too jittery once it finished animating.

base/js/admin.js
```diff
@@ -140,9 +140,15 @@ var sowbForms = window.sowbForms || {};
 										$$f = $$;
 									}
 
-									// Call the function on the wrapper we've selected
-									$$f[thisHandler[i][0]].apply($$f, typeof thisHandler[i][2] !== 'undefined' ? thisHandler[i][2] : []);
-									
+									if ( thisHandler[i][0] == 'show' ) {
+										$$f.slideDown( 'fast' );
+									} else if( thisHandler[i][0] == 'hide' ) {
```
Misplon, line 145: Perhaps update this to `} else if (`

Misplon:
Thanks, mate. I'm wondering if we can't make this smoother. What do you think about `fadeToggle`? 

https://api.jquery.com/fadeToggle/

AlexGStapleton:
Yeah, that looks better.

Misplon:
@AlexGStapleton Please, try testing the Tabs widget with this branch as your head. There is something happening with the selection of the Text vs Layout Builder field.

AlexGStapleton:
@Misplon Sorry just saw your message - GitHub was only showing the PR as approved and not that it had a new comment. 😞 
I just had a look and I'm not noticing any issues. Can you send me a video of what you're seeing?

Misplon:
This happening for you?

[tabs.mov.zip](https://github.com/siteorigin/so-widgets-bundle/files/5048315/tabs.mov.zip)

AlexGStapleton:
I can replicate that in the Block Editor. I'll run some tests and get back to you on this.

Misplon:
Thanks. The video is from the Classic Editor.

Misplon:
It looks like the animation is visible on load. Try opening the Slider widget in the Classic Editor, you should see the Pause on hover setting being hidden.

Misplon:
Autoplay should be disabled to view the issue I'm raising in the Slider widget.

The Tabs widget seems fine now.

Misplon:
Looks good.

In the case of the Tabs widget or widgets where we're switching between Text and the Layout Builder, I'm not sure this PR is an improvement. Screencast below.

[tabs.mov.zip](https://github.com/siteorigin/so-widgets-bundle/files/5059005/tabs.mov.zip)

AlexGStapleton:
That looks fine to me.

Regardless, there's nothing that can really be done to improve how that looks outside of disabling it for that specific instance. If that's something you would like, I suppose I could implement a way of selectively disabling the animation using an optional state emitter parameter, but it's debatable if that's worth the effort.

Misplon:
Cool, we can leave it as is.

PR #1118, opened 2020-08-05
Post Carousel: Use totalposts for navigation visibility conditional

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/pull/1116#issuecomment-669134860

### DIFF
--- a/widgets/post-carousel/js/carousel.js
+++ b/widgets/post-carousel/js/carousel.js
@@ -106,28 +106,27 @@ jQuery( function ( $ ) {
 				}
 			} );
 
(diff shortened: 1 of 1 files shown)


Misplon:
As discussed on Slack, it appears as though this branch is causing the Layout Block to fail after saving and reloading.

Misplon:
This branch is fine now, thanks.

Misplon:
@gregpriday This PR is required for https://github.com/siteorigin/so-widgets-bundle/pull/1116. Thanks!

Issue #1120, opened 2020-08-12
Sliders: Not visible in Firefox - WP 5.5


Misplon:
To recreate this issue, add a slider widget and view the page in Firefox using WP 5.5. 

The issue is visible on our demo site https://demo.siteorigin.com/premium/lightbox/, scroll to the bottom of the page. 

The issue doesn't appear to be present in the Cycle2 demos http://jquery.malsup.com/cycle2/demo/pause.php.

`<div class="sow-slider-base " style="display: none">`

PR #1121, opened 2020-08-12
SiteOrigin Slider: Disable 5.5 Lazy Loading

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1120

In [WordPress 5.5 Lazy Loading](https://make.wordpress.org/core/2020/07/14/lazy-loading-images-in-5-5/) was introduced. Cycle2 doesn't appear to handle the new lazy load very well as the image is never actually loaded, nor are any related image load events are triggered (for example, [this one](https://github.com/siteorigin/so-widgets-bundle/blob/1.17.4/js/slider/jquery.slider.js#L255-L262)) - at least in Firefox. This PR disables the new loaded attribute for the SiteOrign Slider widget to avoid this.


AlexGStapleton:
To be clear, this issue only affects the SiteOrigin Slider widget as the other sliders use CSS to display the image.

Issue #1122, opened 2020-08-13
Post Carousel: After all slides have loaded, prevent any blank "slides" from appearing


AlexGStapleton:
[Requested here](https://wordpress.org/support/topic/post-carousel-1-17-4/)

Prevent situations like the below screenshot.

![Screenshot_2020-08-13 SiteOrigin](https://user-images.githubusercontent.com/17275120/90101596-5af95c80-dd82-11ea-97a1-f87fb7cee2dc.png)

PR #1123, opened 2020-08-13
Post Carousel: Prevent Blank Slides after all posts are loaded

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1122

A custom Slick Go to Function was implemented to avoid https://github.com/kenwheeler/slick/issues/1006.

### DIFF
--- a/widgets/post-carousel/js/carousel.js
+++ b/widgets/post-carousel/js/carousel.js
@@ -5,6 +5,19 @@ var sowb = window.sowb || {};
 jQuery( function ( $ ) {
 
 	sowb.setupCarousel = function () {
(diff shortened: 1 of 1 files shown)


widgets/post-carousel/js/carousel.js
```diff
@@ -5,6 +5,19 @@ var sowb = window.sowb || {};
 jQuery( function ( $ ) {
 
 	sowb.setupCarousel = function () {
+		$.fn.setSlideTo = function ( slide ) {
```
Misplon, line 8: Should we perhaps format this as `function( slide )` because it's new?

Misplon:
Looking good.

PR #1124, opened 2020-08-13
Slider: Only Preload Slide Images

AlexGStapleton:
Images added to the Hero/Layout Slider contents will prevent the Hero and Layout Slider from showing due to the 5.5 lazy load. To be clear, it's not the slide background image, it's a separate image added to the slide content. 

Previously, we would preload all of the images added to the slider, but that doesn't actually appear to be intentional and looks to be more of an oversight that wasn't corrected with the introduction of the Hero and Layout Slider widgets.

### DIFF
--- a/js/slider/jquery.slider.js
+++ b/js/slider/jquery.slider.js
@@ -245,7 +245,7 @@ jQuery( function($){
 			var imagesLoaded = 0;
 			var sliderLoaded = false;
 
-			// Preload all of the slide image, when they're loaded, then display the slider
+			// Preload all of the slide images, when they're loaded, then display the slider.
 			images.each( function(){
 				var $i = $(this);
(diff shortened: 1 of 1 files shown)


Misplon:
Thanks, mate. Can you send me a test layout? The demo server is using the release version of WB and I don't seem to be able to recreate the issue in FF.

https://demo.siteorigin.com/corp/lazy-test/

AlexGStapleton:
The slider on the linked page isn't being lazy loaded for some reason. You can check this by checking for the loading attribute on the image. Can you try replicating this on a different site?

[Here's a test layout.](https://drive.google.com/uc?id=1qVXP0RGHPY10wYW-f4X0GOI4mD52zX-4)

Issue #1085, opened 2020-07-10
Prevent preview issues with widgets that attempt to render sub widgets but don't output them


AlexGStapleton:
We should prevent the output of [widget inline CSS after widgets](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/siteorigin-widget.class.php#L243) for widgets added using the SiteOrigin Layout Block, Live Editor, and WP Preview and instead rely on wp_footer to output. Doing so will prevent a potential situation where a sub widget is rendered, but never output and this results in the parent widget CSS never being output.

For example, the We Love Design layout includes a Hero widget that has a button but it doesn't contain the [buttons]. Due to the button not outputting and the CSS being output after every widget, the CSS the Hero widget CSS is never output due to it being cleared when the button was generated by the plugin. Please note that this specific issue was prevented in [this PR](
https://github.com/siteorigin/so-widgets-bundle/pull/1081) so this specific example can't be replicated anymore, but this issue applies to all sub widgets.

Please note, that the SiteOrigin Layout Block CSS is already output [here](https://github.com/siteorigin/siteorigin-panels/blob/ddd2733c4001a22515e8ff77d5817fd7705d5ab0/inc/admin.php#L1196-L1202) so it's not relying on [this line](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/siteorigin-widget.class.php#L243).

PR #1127, opened 2020-08-16
Only output widget CSS inline during if preview is full widget

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1085

This PR resolves the sub widget preview issue by only outputting widget CSS inline during if the preview is the direct widget preview - the dedicated preview widget button. A conditional is added rather than completely removing this line to allow for the widget preview button to still render as expected.

[Test widget](https://drive.google.com/uc?id=1vinLmiM6-LgEWOIiGXfdSDipd-hlAAg9). Once installed navigate to **WP Admin > plugins > SiteOrigin Widgets** and activate the Preview Error widget. Add it to a Block Editor page and then set a text color, and preview the widget. As expected, the color won't change. Load this PR and make any change to the widget (to trigger the content preview update) and then preview the widget again.

Note: You don't actually have to set the headline widget up.

### DIFF
--- a/base/siteorigin-widget.class.php
+++ b/base/siteorigin-widget.class.php
@@ -238,8 +238,8 @@ abstract class SiteOrigin_Widget extends WP_Widget {
 		echo $args['after_widget'];
 		do_action( 'siteorigin_widgets_after_widget_' . $this->id_base, $instance, $this );
 
-		if ( $this->is_preview( $instance ) ) {
-			// print inline styles if we're preview the widget.
+		// If this is a widget preview, we need to print the styling inline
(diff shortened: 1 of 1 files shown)


Misplon:
@gregpriday I've tested this PR using widgets in a theme sidebar widget area and using Page Builder in the Classic Editor and haven't found any issues.

Misplon:
To do: Test this PR with the global Copy styles setting disabled. I'll hit that in the morning before lunch.

Misplon:
@AlexGStapleton, Greg asked me to check the Copy styles setting with this PR. I don't see any issue with it enabled. Do you?

SiteOrigin_Page_Builder_‹_SiteOrigin_—_WordPress

gregpriday (approved):

Issue #988, opened 2020-02-20
Contact Form: Add a new setting to set field max-width


Misplon:
We have a setting to set the label width. Let's consider adding a setting to set field max-width. The use case is for when your form container is wider than you'd like your field width to be.

Misplon:
If the Label width setting is in fact a max-width then consider renaming that setting label to Maximum Width to be clearer.

AlexGStapleton:
Label width setting is `width`.

PR #1128, opened 2020-08-17
Contact: Add Field max width setting

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/988

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -446,6 +446,11 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 								'type'  => 'measurement',
 								'label' => __( 'Padding', 'so-widgets-bundle' )
 							),
+							'max_width'    => array(
+								'type'    => 'measurement',
+								'label'   => __( 'Max width', 'so-widgets-bundle' ),
+								'default' => '',
+							),
 							'height'        => array(
 								'type'  => 'measurement',
 								'label' => __( 'Height', 'so-widgets-bundle' )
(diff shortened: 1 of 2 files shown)


AlexGStapleton:
@Misplon How do you want to handle the datetime field? We could maybe exclude it or apply the width to its container instead.

Misplon:
Is it perhaps possible to apply the max-width to the form container?

Misplon:
Mmm, I wonder how we landed up with a label width setting? Is there a use case?

Misplon:
Let's try adding the width to the date/time container.

AlexGStapleton:
How does that look?

> Mmm, I wonder how we landed up with a label width setting? Is there a use case?

The setting was introduced [here](https://github.com/siteorigin/so-widgets-bundle/commit/ff3cbb964ec9fb837bb36b139475ec40e0c73a52). I suspect it was added to help with readability.

Misplon:
Thanks, mate.

* Should we drop the max-width at the mobile resolution? I think we should.
* Should we apply the max-width to the map in the location field?

AlexGStapleton:
The last commit disables the field width on mobile and applies the Field Max Width to the Location field container.

AlexGStapleton:
@Misplon We don't have to introduce a breakpoint widget setting, but it is consistent with other widgets (Slider, Hero, Layout Slider, and Social Media Buttons), and allows for the breakpoint to be adjusted if Page Builder isn't being used.

Misplon:
Ahh, true.

Misplon (approved):

gregpriday (approved):

PR #1129, opened 2020-08-17
Widget Settings: Avoid jQuery migrate notice by using $.get()

AlexGStapleton:
This PR prevents the [Enable jQuery Migrate Helper](https://wordpress.org/plugins/enable-jquery-migrate-helper/) plugin from showing a notice when viewing **WP Admin > Plugins > SiteOrigin Widgets**. This change isn't directly required, but this will avoid any reports.

To verify the settings form is still working as intended please navigate to**WP Admin > Plugins > SiteOrigin Widgets** and open a any widgets settings. Make a change, save, and then reload the page. Check the settings to ensure your change was saved.

### DIFF
--- a/admin/admin.js
+++ b/admin/admin.js
@@ -132,12 +132,16 @@ jQuery( function( $ ){
         var $$ = $(this);
         e.preventDefault();
 
-        dialog.find('.so-content')
+        $content = dialog.find('.so-content');
+        $content
             .empty()
             .addClass('so-loading')
-            .load( $$.data('form-url'), function(){
-                $(this).removeClass('so-loading');
-            } );
+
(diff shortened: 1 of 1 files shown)


Misplon:
Thanks, I can confirm the form still works.

I can see the updated JS in my browser but the notice remains. Do I have to remove the jQuery Migrate notice manually or does it remove itself on your end when you switch to this branch and reload the page at Plugins > SiteOrigin Widgets?

AlexGStapleton:
I must have accidentally dismissed it because upon disabling and reactivating the plugin I was getting the notice again. The last commit will prevent that - please confirm saving is working as expected again.

Misplon:
Looks good, thanks.

Issue #1126, opened 2020-08-16
Post Carousel: Add Navigation Color, Background, and Related Hover settings


AlexGStapleton:
[Related](https://github.com/siteorigin/so-widgets-bundle/pull/1115#issuecomment-674434355)

PR #1131, opened 2020-08-24
Post Carousel: Add Design Settings

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1126

### DIFF
--- a/widgets/post-carousel/post-carousel.php
+++ b/widgets/post-carousel/post-carousel.php
@@ -167,9 +167,9 @@ class SiteOrigin_Widget_PostCarousel_Widget extends SiteOrigin_Widget {
 						'label' => __( 'Navigation arrow color', 'so-widgets-bundle' ),
 						'default' => '#fff',
 					),
-					'navigation_color_hover' => array(
+					'navigation_hover_color' => array(
 						'type' => 'color',
-						'label' => __( 'Navigation arrow color hover', 'so-widgets-bundle' ),
+						'label' => __( 'Navigation arrow hover color', 'so-widgets-bundle' ),
 					),
 					'navigation_background' => array(
 						'type' => 'color',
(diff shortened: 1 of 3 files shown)


widgets/post-carousel/post-carousel.php
```diff
@@ -156,6 +156,33 @@ function get_widget_form(){
...
+				'hide' => true,
+				'fields' => array(
+					'navigation_color' => array(
+						'type' => 'color',
+						'label' => __( 'Navigation arrow color', 'so-widgets-bundle' ),
+						'default' => '#fff',
+					),
+					'navigation_color_hover' => array(
+						'type' => 'color',
+						'label' => __( 'Navigation arrow color hover', 'so-widgets-bundle' ),
```
Misplon, line 172: Let's perhaps label this "Navigation arrow hover color".

widgets/post-carousel/post-carousel.php
```diff
@@ -156,6 +156,33 @@ function get_widget_form(){
...
+					),
+					'navigation_background' => array(
+						'type' => 'color',
+						'label' => __( 'Navigation background', 'so-widgets-bundle' ),
+						'default' => '#333',
+					),
+					'navigation_background_hover' => array(
+						'type' => 'color',
+						'label' => __( 'Navigation background hover', 'so-widgets-bundle' ),
```
Misplon, line 181: Let's perhaps label this "Navigation hover background". Unless there is precedence I've missed.

Misplon:
Do you want to add a setting for the overlay hover color?

AlexGStapleton:
How does that look? I've added the overlay hover color setting.

What are your thoughts on moving Design settings above Post Query?

Misplon:
Super, thanks.

I'm ok with the order for now.

Can we bold the Post query tab heading so it matches the Design tab heading?

What do you think of changing the label from Overlay hover color to Thumbnail overlay hover color?

AlexGStapleton:
> Can we bold the Post query tab heading so it matches the Design tab heading?

We can. As that's a general change though (as the Posts Query settings group is always unbolded) I've submitted a PR for this: https://github.com/siteorigin/so-widgets-bundle/pull/1136

Misplon (approved):

gregpriday (approved):

PR #1137, opened 2020-08-30
Price Table: Remove text-shadow from feature text

AlexGStapleton:
This avoids an issue with Tooltips. There's no difference after removing this CSS.

### DIFF
--- a/widgets/price-table/styles/atom.less
+++ b/widgets/price-table/styles/atom.less
@@ -99,8 +99,6 @@
 					.gradient(#f1f1f1, #eeeeee, #f1f1f1);
 				}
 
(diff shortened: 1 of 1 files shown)


Misplon:
It's extremely subtle but the text-shadow is currently being outputted. I've made the text huge and changed the background color so you can see it.

Front_Page_–_SiteOrigin

Is there not another CSS adjustment we can make to exclude the tooltip? We could perhaps rather add `text-shadow: none;` to `.so-premium-tooltip`.

AlexGStapleton:
It's not directly possible to change the background of the features so while it is more noticeable with a black background, it's never visible like that. Regardless, I could add an exception so I'll just do that in a different PR.

Issue #875, opened 2019-06-06
Google Maps: Add setting to delay loading of Google Maps API until after user clicks button


AlexGStapleton:
For DSGVO and GDPR compliance, the Google Maps API should be loaded after the user consents to transmit certain data to Google. We should introduce a way to delay loading of the Google Maps API until _after_ the user browser clicks a button.

As this setting will be the same for all instances of the SiteOrigin Google Maps widget, it might be a good idea to add this as a global widget setting.

PR #1139, opened 2020-09-01
Google Maps: Add Optional Consent Check before loading Google Maps API

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/875


widgets/google-map/google-map.php
```diff
@@ -466,6 +466,28 @@ function get_settings_form() {
...
+			'map_consent' => array(
+				'type' => 'checkbox',
+				'label'       => __( 'Require consent before loading Maps API', 'so-widgets-bundle' ),
+				'description' => __( 'Consent is required for the Google Maps widget to comply with regulations like DSGVO, or GDPR.', 'so-widgets-bundle' ),
+				'default' => false,
+			),
+
+			'map_consent_btn_text' => array(
+				'type' => 'text',
+				'label'       => __( 'Consent button text', 'so-widgets-bundle' ),
```
Misplon, line 478: An indentation fix is required here, thanks.

widgets/google-map/google-map.php
```diff
@@ -466,6 +466,28 @@ function get_settings_form() {
...
+				'default' => __( 'I Consent', 'so-widgets-bundle' ),
+			),
+
+			'map_consent_notice' => array(
+				'type' => 'tinymce',
+				'label'       => __( 'Consent prompt text', 'so-widgets-bundle' ),
```
Misplon, line 484: Indentation check here too.

widgets/google-map/tpl/js-map.php
```diff
@@ -6,9 +6,16 @@
  * @var $fallback_image_data
  */
 ?>
+<?php if ( $map_consent ): ?>
```
Misplon, line 9: Ideally need a space before `:`.

Misplon:
- Clicking anywhere in the content message reveals the map.
- Can we perhaps have the map in the background with an opacity overlay?
- Could we perhaps simplify the message? I like Borlab's option:

_By loading the map, you agree to OpenStreetMap Foundation's privacy policy.
Learn more (linked)
Load map_

Borlabs_Cookie_Iframe_Demo_–_WordPress_Cookie_Plugin

AlexGStapleton:
> Can we perhaps have the map in the background with an opacity overlay?

Not the JavaScript version, no. We could show the static map, but that won't work for most people without them adjusting their API key, and it would almost double API usage.

We could however show a background image with a generic map image - which appears to be the case in your image. If you would like to go that route, we'll need to have a generic placeholder image created. We should also add a setting that allows users to change the background image.

> Could we perhaps simplify the message? I like Borlab's option:

Yeah, we can do that. Do you want this to be editable, or do you want it to "Learn More"?
How about we only show the Read More link if the site has a privacy policy set up? Is that text editable?

Misplon:
> If you would like to go that route, we'll need to have a generic placeholder image created. We should also add a setting that allows users to change the background image.

Super, let's do that, thanks.

> Yeah, we can do that. Do you want this to be editable, or do you want it to "Learn More"?
How about we only show the Read More link if the site has a privacy policy set up? Is that text editable?

I'd be keen to use a learn more link but make it all editable so users can change the text and destination.

AlexGStapleton:
I sent you a few questions over slack regarding settings. Based on your answer I'll make some further adjustments to the design - specifically, change the color of the prompt text.

Do we have an image we can use for the consent prompt background?

Misplon:
> Do we have an image we can use for the consent prompt background?
I'll try to find some sort of generic map illustration that we can use.

Misplon:
We might need to get a vector illustrated for us. Perhaps we can use something like the below for now?

snazzy-image-3x

Misplon:
Thanks, mate. Please, can you check up on:

`Parse error: syntax error, unexpected ',', expecting ';' in /wp-content/plugins/so-widgets-bundle/widgets/google-map/google-map.php on line 490`

AlexGStapleton:
Sorry about thar. I excluded a setting from one of the last commits and let that slip through.

Misplon:
Thanks, mate. I've made some minor styling adjustments. The Less doesn't seem to be writing for me locally. How does that look for you?

Misplon (approved):

Misplon (approved):

gregpriday (approved):

PR #1148, opened 2020-09-11
Social Media Buttons: Increase Completely rounded value

### LINKED ISSUE #1147: Social Media Buttons: Ensure Completely Rounded is working
The Completely Rounded setting doesn't appear to be working as expected.

Front_Page_–_SiteOrigin

(Send a beta build to https://siteorigin.com/thread/cant-get-rounded-icons-with-the-widget/ once ready).

AlexGStapleton:
Resolve #1147

### DIFF
--- a/widgets/social-media-buttons/social-media-buttons.php
+++ b/widgets/social-media-buttons/social-media-buttons.php
@@ -130,7 +130,7 @@ class SiteOrigin_Widget_SocialMediaButtons_Widget extends SiteOrigin_Widget {
 							'0'    => __( 'None', 'so-widgets-bundle' ),
 							'0.25' => __( 'Slightly rounded', 'so-widgets-bundle' ),
 							'0.5'  => __( 'Very rounded', 'so-widgets-bundle' ),
-							'1.5'  => __( 'Completely rounded', 'so-widgets-bundle' ),
+							'2.5'  => __( 'Completely rounded', 'so-widgets-bundle' ),
 						),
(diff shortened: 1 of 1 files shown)


Misplon:
At the moment, this change doesn't seem to be working for me. I've tried inserting a new widget. Same issue.

Front_Page_–_SiteOrigin

AlexGStapleton:
I'm having trouble replicating this after making this change. Can you please show me a screenshot of your settings? There's a bunch of setting combinations and I _feel_ like I've tried them all but I must have missed one.

Misplon:
I'm adding Facebook without changing any defaults. I'm then scrolling down and setting Completely rounded, I'm not changing anything else.

AlexGStapleton:
So the only way to really fix that is to adjust the width. The problem with doing that is that the required width is closer to Normal Padding. Medium Padding looks worse. We have two ways of proceeding:

1. Remove the predefined fields and use number fields (migration code will be added to prevent any noticeable changes). This will allow for the user to manually size/round/pad the icons.
2. Come up with new padding and width sizing.

The latter will remove the need for migration code and keep the widget relatively the same. The problem however is that there are going to be some different sized icons. An icon that's wide is going to require different values to an icon that's tall. The first route won't really avoid that but it will allow the user to find setting values they're happy with.

Misplon:
Thanks. I'll take a look over this tomorrow and get back to you.

Issue #1149, opened 2020-09-12
Widget Block: Preview styling is missing


Misplon:
Styling is currently missing from Widget Block previews. 

[Reporting thread](https://siteorigin.com/thread/cant-preview-correctly-the-widgets-anymore/).

PR #1150, opened 2020-09-12
Ensure widget CSS is output inline if WB Preview

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1149.
This is a follow up to https://github.com/siteorigin/so-widgets-bundle/pull/1127 so the following needs to be tested:

SiteOrigin Widget Block Preview
SiteOrigin Layout Block Preview
Block Editor Preview
Classic Editor Preview
Widget Preview (this is the dedicated widget button).

### DIFF
--- a/base/siteorigin-widget.class.php
+++ b/base/siteorigin-widget.class.php
@@ -239,7 +239,7 @@ abstract class SiteOrigin_Widget extends WP_Widget {
 		do_action( 'siteorigin_widgets_after_widget_' . $this->id_base, $instance, $this );
 
 		// If this is a widget preview, we need to print the styling inline
-		if ( $this->is_preview( $instance ) && isset( $_POST['action'] ) &&  $_POST['action']== 'so_widgets_preview' ) {
+		if ( $this->is_preview( $instance ) ) {
 			siteorigin_widget_print_styles();
 		}
 	}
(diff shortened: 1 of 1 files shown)


Misplon:
* When previewing the SO Widget Block > Accordion, the icons to open/close the accordion items aren't displaying in preview.
* When testing the SO Layout Block > Consultant layout the first Hero appears when I preview the first time, if I click edit and then the preview icon (eye icon) again, the Hero doesn't display in the preview.

Misplon:
> When previewing the SO Widget Block > Accordion, the icons to open/close the accordion items aren't displaying in preview.

This is only happening when I use one of the preset themes like Flat.

AlexGStapleton:
I'm having trouble replicating either of those on an existing page or a new page. Can you please clarify the exact edit you're making to the Hero widget?

Misplon:
The Hero is within a Layout Block. I'm not making an edit. Just preview and return to edit mode then back into preview mode. Screencast below. Thanks!

[block-editor-previews.mov.zip](https://github.com/siteorigin/so-widgets-bundle/files/5213080/block-editor-previews.mov.zip)

Misplon (approved):

gregpriday (approved):

Issue #1143, opened 2020-09-08
Elementor: Fix label colors


Misplon:
The overall styling needs a few fixes.

![elementor](https://user-images.githubusercontent.com/789159/92499338-0c61a580-f1fc-11ea-868d-1404c0c042f2.jpg)

PR #1153, opened 2020-09-13
Elementor styling changes

AlexGStapleton:
-  Set default form text color (Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1143)
- Prevent color picker button overlapping selected color and incorrect sizing. ([screenshot](https://i.imgur.com/2OQ9D4f.png))
- Prevent Noticeable lag when using the color picker. ([gif](https://i.imgur.com/23JpcDi.gif) - every time it moves is a single click)
- Move Color picker buttons and color field to the same line. ([screenshot](https://i.imgur.com/2OQ9D4f.png))
- Reduce size of measurement fields

### DIFF
--- a/compat/elementor/styles.less
+++ b/compat/elementor/styles.less
@@ -14,20 +14,27 @@
 			background: none;
 
 			.wp-color-result {
-				float: none;
+				float: left;
 				margin: 0;
 			}
 
 			.wp-picker-input-wrap {
 				padding: 0;
-				display: flex;
-				flex-direction: row;
-				float: right;
+				float: left;
 
 				.siteorigin-widget-input-color,
 				.button.wp-picker-default {
 					height: 34.1px;
 				}
+
+				.button.wp-picker-default {
+					float: right;
+					width: auto;
+				}
+			}
+
+			.wp-picker-holder {
+				clear: both;
 			}
 
 			.iris-picker.iris-border {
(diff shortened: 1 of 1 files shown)


Misplon:
Elementor___Elementor_Test

and 

Elementor___Elementor_Test

To solve the issue in the first screenshot, the button issue you could set the height to 32px and the font size to 12px.

The second screenshot needs a more in-depth look. The more we can rely on default color picker styling the better I reckon.

AlexGStapleton:
Oh. I didn't realize you could resize lower than I had it. 🤦‍♂️ 

While I agree in principal, it looked very disjointed before.
On that note, I can't do anything about the color picker being larger than the sidebar without forcefully resizing it and that'll require a large number of overrides. It's not ideal but I feel we should leave that. The last commit prevents the buttons and color field from displaying incorrectly.

Misplon:
**Second screenshot:**
Ok, cool, understood. Funny thing is I didn't resize the sidebar to that size, it seems to be like that by default for me.

**First screenshot:**
The first of my screenshots happens before the color picker is opened.
> To solve the issue in the first screenshot, the button issue you could set the height to 32px and the font size to 12px.

Misplon:
color-picker

Our clear where my arrow is doesn't seem to be effective enough for the 6px top margin to kick in. Are we safe to add:

```
.wp-picker-holder:before {
    clear: both;
    content: "";
    display: table;
}
```

AlexGStapleton:
I didn't see any issues with adding that CSS.

Misplon (approved):

Issue #1152, opened 2020-09-13
SiteOrigin Image: Stretching in Safari


Misplon:
Related: https://github.com/siteorigin/so-widgets-bundle/issues/1141.

Images are stretched in Safari due to flexbox. Adding `flex-direction: column;` resolves the issue but I haven't tested to see if that might introduce other problems.

Ref: https://stackoverflow.com/questions/57516373/image-stretching-in-flexbox-in-safari

AlexGStapleton:
`flex-direction: column;` vertically stretches the images in IE 11. How about `flex-direction: row;`? That doesn't result in any differences in IE 11. If that doesn't work I guess we could use [something like this](https://paper-leaf.com/insights/targeting-ie-10-11-browsers-css/) to target IE 11 directly.

Misplon:
I'd be keen to revert our IE11 fix and instead try targetting IE directly as per https://paper-leaf.com/insights/targeting-ie-10-11-browsers-css/.

Issue #1141, opened 2020-09-03
IE11 Image Widget: Image exceeds widget column unless set to full width


AlexGStapleton:
[Reported here](https://wordpress.org/support/topic/widgets-bundle-stretching-pictures-in-ie)

PR #1155, opened 2020-09-14
Image: Prevent image stretching while using Safari

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1152

This is done by applying the https://github.com/siteorigin/so-widgets-bundle/issues/1141 fix in a media query that only targets IE 11.

### DIFF
--- a/widgets/image/styles/default.less
+++ b/widgets/image/styles/default.less
@@ -24,9 +24,13 @@
 	}
 
 	> a {
-		display: flex;
+		display: inline-block;
 		width: @image_width;
 		max-width: @image_max_width;
+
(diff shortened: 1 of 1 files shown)


Misplon (approved):

Misplon:
Thanks. I can't test in IE 11 so I'll leave that you. I'm just waiting on a Safari user to confirm using a beta release before I merge.

AlexGStapleton:
I can confirm this works in IE 11.

Misplon:
No update from the user I sent the hotfix too, happy to go with what we have.

Issue #1154, opened 2020-09-14
Posts Carousel: Link target not working when multiple carousel's on the page


Misplon:

AlexGStapleton:
I'm having a hard time replicating this. Can you please send me a copy of the layout you're using?

Misplon:
Sure, here is a demo https://demo.siteorigin.com/corp/multiple-posts-carousel/. Both carousels are set to open posts in a new tab.

PR #1158, opened 2020-09-17
Post Carousel: Add link target to items with thumbnail

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1154

The second commit is more about tidying up as things aren't very readable without it.

### DIFF
--- a/widgets/post-carousel/tpl/carousel-post-loop.php
+++ b/widgets/post-carousel/tpl/carousel-post-loop.php
@@ -26,6 +26,6 @@ while($posts->have_posts()) : $posts->the_post(); ?>
 				</a>
 			<?php endif; ?>
 		</div>
-		<h3><a href="<?php the_permalink() ?>" id="sow-carousel-id-<?php echo the_ID(); ?>" tabindex="-1"><?php the_title() ?></a></h3>
+		<h3><a href="<?php the_permalink() ?>" id="sow-carousel-id-<?php echo the_ID(); ?>" <?php echo $link_target == 'new' ? 'target="_blank" rel="noopener noreferrer"': ''; ?> tabindex="-1"><?php the_title() ?></a></h3>
 	</div>
 <?php endwhile; wp_reset_postdata(); ?>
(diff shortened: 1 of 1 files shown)


Misplon:
Super, thanks. 

We don't seem to be applying the link target to the post title.

Misplon (changes requested):
Looking good, we just need to apply the link target setting to the link in the post title.

Misplon (approved):

widgets/post-carousel/tpl/carousel-post-loop.php
```diff
@@ -6,18 +6,26 @@
 while($posts->have_posts()) : $posts->the_post(); ?>
 	<div class="sow-carousel-item" tabindex="-1">
 		<div class="sow-carousel-thumbnail">
-			<?php if( has_post_thumbnail() ) : $img = wp_get_attachment_image_src(get_post_thumbnail_id(), $instance['image_size']); ?>
-				<a href="<?php the_permalink() ?>" style="background-image: url(<?php echo sow_esc_url($img[0]) ?>)" aria-labelledby="sow-carousel-id-<?php echo the_ID(); ?>" tabindex="-1">
+			<?php if ( has_post_thumbnail() ) : $img = wp_get_attachment_image_src( get_post_thumbnail_id(), $instance['image_size'] ); ?>
+				<a href="<?php the_permalink() ?>"
+					<?php echo $link_target == 'new' ? 'target="_blank" rel="noopener noreferrer"' : '' ?>
```
gregpriday, line 11: Do we want `noopener noreferrer` to be the default here? I can understand that we use this for links inside the WordPress admin, but is it our place to block a referrer in a user's frontend? They might want/need this information to be passed.

gregpriday (commented):
Code all looks good, just a general question.

widgets/post-carousel/tpl/carousel-post-loop.php
AlexGStapleton: Replied over Slack

gregpriday (approved):

PR #1161, opened 2020-09-20
Social Network Buttons: Add dedicated icon/padding/rounding/margin fields

AlexGStapleton:
[Background to change](https://github.com/siteorigin/so-widgets-bundle/pull/1148#issuecomment-691510645)

### DIFF
--- a/widgets/social-media-buttons/social-media-buttons.php
+++ b/widgets/social-media-buttons/social-media-buttons.php
@@ -150,18 +150,31 @@ class SiteOrigin_Widget_SocialMediaButtons_Widget extends SiteOrigin_Widget {
 							'justify' => __( 'Justify', 'so-widgets-bundle' ),
 						),
 					),
-					'margin'     => array(
-						'type'    => 'select',
-						'label'   => __( 'Margin', 'so-widgets-bundle' ),
-						'default' => '0.1',
-						'options' => array(
-							'0.1' => __( 'Low', 'so-widgets-bundle' ),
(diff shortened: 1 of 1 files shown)


Misplon:
I've added a FB network without changing any default settings:

fb-develop

When I switch to this branch, the effect is:

fb-02

widgets/social-media-buttons/social-media-buttons.php
```diff
@@ -194,25 +194,28 @@ function modify_instance( $instance ) {
...
+			$instance['design']['padding']        = $instance['design']['padding'] . 'em';
 			$instance['design']['padding_unit']   = 'em';
 
 			// The margin value was previously changed based on the align setting.
-			$top = $right = $bottom = $left = $instance['design']['margin']. 'em';
+			$top = $right = $bottom = $left = $instance['design']['margin'] . 'em';
 			switch ( $instance['design']['align'] ) {
 				case 'left':
-					$left = '0';
+					$left = '0em';
```
Misplon, line 209: It isn't important but do we need `em` here?

AlexGStapleton: Yes. Without it, the value will be imported without a unit of measurement set.

AlexGStapleton:
We need to come up with some new default settings.

AlexGStapleton:
The new defaults should ideally be px based to avoid inconsistent sizing between sites. The migration code will migrate the previously selected settings and retain em so there's no need to stick with em for the new defaults.

Issue #1163, opened 2020-09-26
Tab: Add setting to disable scroll after changing/opening item


AlexGStapleton:
We should add a global widget setting for both the Tab and Accordion widget that will allow users to disable the scrolling that can occur after changing tab or opening a new accordion. This scroll occurs when parts of the panel/tab aren't completely visible and there are situations, especially if the content is very long where it may not be as desired.

PR #1165, opened 2020-09-27
Tab & Accordion: Add setting to disable scroll

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1163

### DIFF
--- a/widgets/accordion/accordion.php
+++ b/widgets/accordion/accordion.php
@@ -45,9 +45,9 @@ class SiteOrigin_Widget_Accordion_Widget extends SiteOrigin_Widget {
 		return array(
 			'scrollto_after_change' => array(
 				'type'        => 'checkbox',
-				'label'       => __( 'Scroll user to top of the panel when opening panel', 'so-widgets-bundle' ),
+				'label'       => __( 'Scroll top', 'so-widgets-bundle' ),
 				'default'     => true,
-				'description' => __( 'This setting will scroll the user to the top of any opened panels. This is done to potentially prevent the need for the user to do this ', 'so-widgets-bundle' ),
+				'description' => __( 'When opening the panel, scroll the user to the top of the panel', 'so-widgets-bundle' ),
 			)
 		);
 	}
--- a/widgets/accordion/accordion.php
+++ b/widgets/accordion/accordion.php
@@ -47,7 +47,7 @@ class SiteOrigin_Widget_Accordion_Widget extends SiteOrigin_Widget {
 				'type'        => 'checkbox',
 				'label'       => __( 'Scroll top', 'so-widgets-bundle' ),
 				'default'     => true,
-				'description' => __( 'When opening the panel, scroll the user to the top of the panel', 'so-widgets-bundle' ),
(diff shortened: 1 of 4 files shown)


widgets/accordion/accordion.php
```diff
@@ -37,8 +37,30 @@ function initialize() {
...
+
+		add_action( 'siteorigin_widgets_enqueue_frontend_scripts_sow-accordion', array( $this, 'enqueue_widget_scripts' ) );
 	}
-	
+
+	function get_settings_form() {
+		return array(
+			'scrollto_after_change' => array(
+				'type'        => 'checkbox',
+				'label'       => __( 'Scroll user to top of the panel when opening panel', 'so-widgets-bundle' ),
```
Misplon, line 48: What do you think about renaming this label to:
"Scroll to top of the panel when opening panel"

widgets/accordion/accordion.php
```diff
@@ -37,8 +37,30 @@ function initialize() {
...
+				'default'     => true,
+				'description' => __( 'This setting will scroll the user to the top of any opened panels. This is done to potentially prevent the need for the user to do this ', 'so-widgets-bundle' ),
```
Misplon, line 50: When opening the panel, scroll the user to the top of the panel.

Misplon:
If you're ok with the above changes, perhaps let's make them for both widgets.

I'm getting some unusual results testing locally. Can you perhaps send me a build and I'll test it on the demo server?

AlexGStapleton:
I've adjusted the setting name and descriptions. I also resolved a PHP compatibility issue and that may have been the error you were getting locally. Here's a build:

[so-widgets-bundle.scroll-top-settings.zip](https://github.com/siteorigin/so-widgets-bundle/files/5322398/so-widgets-bundle.scroll-top-settings.zip)

Misplon:
Thanks. I've uploaded this build to the demo server: https://demo.siteorigin.com/corp/accordion/. Both locally and on the demo server, the scroll function doesn't seem to be functioning.

AlexGStapleton:
It can be sometimes quite tricky to trigger the scroll with the Accordion widget. In Corp, the easiest way would be to disable the Sticky Header or remove the header altogether using the inspector. Then scroll to the point where the accordion panel heading is barely visible and then click it. I considered suggesting not implementing this change for the Accordion widget but I received a report from a user and their theme seemed to result in this happening on a fairly consistent basis regardless of where the viewport where positioned.

Here's a video:

[](https://drive.google.com/uc?id=1e0B63ChfVyVoowGM2nSeaFa2rHxa9Mfm)

The easiest way to replicate this with the tabs widget is to change the tab placement.

Misplon:
Thanks. Seems fine for Accordion but for me, scroll top seems to be always off for the Tabs widget. I put the tabs at the bottom of the container with a lot of text in each tab and I don't seem to be scrolled up. Do you?

Misplon:
I think for me, the Tabs setting is working in reverse. If I deactivate the Scroll top setting then I scroll to the top.

Misplon (approved):

gregpriday (approved):

Issue #1132, opened 2020-08-26
Post Carousel: Add three settings for Slides to scroll


Misplon:
Add new settings for Slides to Scroll for Mobile, Tablet, and Desktop. We should perhaps add a new Settings section. In that event, Link target and Loop posts should probably be moved into the Settings section.

AlexGStapleton:
What are your thoughts on adding these settings to a Responsive settings group instead of a general Settings group?

Do you have a preference for the setting names and descriptions? How about?

Slides to scroll on Desktop
Slides to scroll on tablet
Slides to scroll on mobile

Description: The number of slides to scroll.

The labels are a bit wordy though. Hm. Maybe the settings group could be called Slides to scroll?

Misplon:
Sure, a Responsive section sounds good.

Let's go with a settings group called "Slides to scroll". Perhaps the below will work.

**Slides to scroll**
Set the number of slides to scroll per navigation click or swipe on mobile devices.

- Desktop
- Tablet
- Mobile

PR #1171, opened 2020-10-04
Post Carousel: Add Slides to Scroll setting

AlexGStapleton:
https://github.com/siteorigin/so-widgets-bundle/issues/1132

@Misplon What are your thoughts on adding breakpoint setting fields for the respective settings? If you're keen, how do you want to handle the iPad Pro check? Dedicated Tablet Landscape and Tablet Portrait settings?

### DIFF
--- a/widgets/post-carousel/post-carousel.php
+++ b/widgets/post-carousel/post-carousel.php
@@ -100,14 +100,19 @@ class SiteOrigin_Widget_PostCarousel_Widget extends SiteOrigin_Widget {
 		);
 	}
 
-	function enqueue_widget_scripts() {
-		wp_localize_script( 'sow-carousel-basic', 'carouselBreakpoints',
+	function enqueue_widget_scripts( $instance ) {
+		wp_localize_script(
+			'sow-carousel-basic',
+			'carouselSettings',
 			apply_filters(
 				'siteorigin_widgets_post_carousel_breakpoints',
 				array(
-					'tablet_landscape' => 1366,
-					'tablet_portrait'  => 1025,
-					'mobile'           => 480,
+					'desktop_slides' => ! empty ( $instance['responsive']['desktop']['slides_to_scroll'] ) ? $instance['responsive']['desktop']['slides_to_scroll'] : 1,
+					'tablet_breakpoint_landscape' => 1366,
+					'tablet_breakpoint_portrait' => 1025,
+					'tablet_slides' => ! empty ( $instance['responsive']['tablet']['slides_to_scroll'] ) ? $instance['responsive']['tablet']['slides_to_scroll'] : 2,
+					'mobile_breakpoint' => 480,
+					'mobile_slides' => ! empty ( $instance['responsive']['mobile']['slides_to_scroll'] ) ? $instance['responsive']['mobile']['slides_to_scroll'] : 1,
 				)
 			)
 		);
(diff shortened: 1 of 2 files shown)


Misplon:
Thanks, mate. Is this PR currently working for you? I have Slides to scroll set to 4 for desktop and tablet and 1 for mobile. I can't see any change using Incognito with the build task running, same in Firefox.

Misplon:
> What are your thoughts on adding breakpoint setting fields for the respective settings? If you're keen, how do you want to handle the iPad Pro check? Dedicated Tablet Landscape and Tablet Portrait settings?

Yes, both sound good. Dedicated settings with the tablet section handled as you suggested.

AlexGStapleton:
Yes. The PR prior to the last two commits worked for me. Let me know if it still doesn't - please see if there's anything present in the console.

widgets/post-carousel/post-carousel.php
```diff
@@ -100,19 +98,6 @@ function initialize() {
 		);
 	}
 
-	function enqueue_widget_scripts() {
-		wp_localize_script( 'sow-carousel-basic', 'carouselBreakpoints',
-			apply_filters(
-				'siteorigin_widgets_post_carousel_breakpoints',
```
gregpriday, line 106: It's highly unlikely, but removing this filter might break some custom themes. Unless it's being added again somewhere else. Is there a way to have some backwards compat here? So still using this filter in `get_template_variables` and passing the results to the `$responsive_settings` variable.

gregpriday (changes requested):

widgets/post-carousel/post-carousel.php
AlexGStapleton: I could maybe add the filter here https://github.com/siteorigin/so-widgets-bundle/blob/67712369cf70ee5a7303565a7bff8381b58e86f5/widgets/post-carousel/post-carousel.php#L320 and rename any of the breakpoint settings to what they were - although that will make it harder to read the code.

Does that sound okay? If so, I'll make this change shortly.

gregpriday: I'd suggest doing something more like this:

```
$breakpoint_defaults = apply_filters(
    'siteorigin_widgets_post_carousel_breakpoints',
    array(
        'tablet_landscape' => 1366,
		'tablet_portrait'  => 1025,
		'mobile'           => 480,
));
```

Then using these values as the defaults for the new array.

`'tablet_portrait_breakpoint' => ! empty ( $instance['responsive']['tablet']['portrait']['breakpoint'] ) ? $instance['responsive']['tablet']['portrait']['breakpoint'] : breakpoint_defaults['tablet_portrait'],`

Haven't thought it through fully, but this might do the trick?

AlexGStapleton: That's a better idea. Let's do that.

AlexGStapleton:
The last commit restore `siteorigin_widgets_post_carousel_breakpoints `. Test code:

```
add_filter( 'siteorigin_widgets_post_carousel_breakpoints', function( $breakpoints ) {
	return array(
		'tablet_landscape' => 768,
		'tablet_portrait'  => 1000,
		'mobile'           => 500,
	);
} );
```

gregpriday (approved):

Issue #1174, opened 2020-10-07
Hero: Save button isn't working in Safari at Appearance > Widgets


Misplon:
The Hero Save button isn't working in Safari at Appearance > Widgets. There aren't any console errors present.

AlexGStapleton:
This is going to be tricky to fix. I'm going to have to send you a build to test. I have a few questions:

- Is the Save button grey or blue?
- If the button is blue, does the widget start saving (as in, an actual server request is made), or does nothing happen when you click the Save button?
- If the widget starts saving, is there a PHP error?
- Does this happen with the other slider widgets, or is only the Hero widget affected?
- Does this happen while using the last WB update, or does this only happen while using release and/or develop?

Misplon:
> Is the Save button grey or blue?
Blue

> If the button is blue, does the widget start saving (as in, an actual server request is made), or does nothing happen when you click the Save button?
Nothing.

> If the widget starts saving, is there a PHP error?
NA

> Does this happen with the other slider widgets, or is only the Hero widget affected?
Appears just to be the Hero.

> Does this happen while using the last WB update, or does this only happen while using release and/or develop?
Seems to persist using master or develop.

PR #1175, opened 2020-10-08
Add Number Step Option to resolve Safari bug

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1174

Safari won't allow number field inputs with a decimal amount unless the field has a decimal step amount. I'm unsure if this is a bug, or intentional.


Misplon:
@AlexGStapleton Following on from our Slack chat, I'm still seeing an issue here. You'll update the PR once you've taken another look.

AlexGStapleton:
The latest commit should resolve this issue.

Misplon (approved):

gregpriday (approved):

Issue #1172, opened 2020-10-04
 Post Carousel: Add Center Items settings


AlexGStapleton:
We should add a setting that centers all visible items if the total number of items doesn't exceed the length of the page.

**Center Items**
Center items if the total number of posts is fewer than the length of the page.

Here's some example code of how this could be done:

CSS:
```
.center-carousel .slick-track[style] {
	display: flex;
	justify-content: center;
	width: 100% !important;
	transform: none !important;
}

.center-carousel .slick-slide {
	float: none;
}
```

Javascript:
```
jQuery( function ( $ ) {
	$( window ).on( 'resize load', function() {
		$( '.sow-carousel-wrapper' ).each( function() {
			var currentCarousel = $( this ),
				$items = currentCarousel.find( '.sow-carousel-items' ),
				numVisibleItems = Math.ceil( $items.outerWidth() / $items.find( '.sow-carousel-item' ).outerWidth( true ) );

			if ( numVisibleItems >= currentCarousel.data( 'post-count' ) ) {
				currentCarousel.addClass( 'center-carousel' );
			} else if ( currentCarousel.hasClass( 'center-carousel' ) ) {
				currentCarousel.removeClass( 'center-carousel' );
			}
		} );
	} );
} );
```

nikospapageorgiou:
Alex, if it's possible to add this as an option, not default. In pages with multiple carousels in desktop mode, the centered mode maybe looks strange.
As an example, i have a page with 15 carousels for the student activities for the last 15 years. This year till now there are only 3 posts in the first carousel, aligned to left by default, and empty space after them. The centered mode will leave blank space, left and right, something strange for the visitor.
Maybe add the setting as default but only when a single carousel appears.
Thank u.

AlexGStapleton:
Hi @nikospapageorgiou,

This functionality would be optional, and behind a setting that would be disabled by default. Having it to default to off by default will prevent any undesired changes after updating.

Kind regards,
Alex

nikospapageorgiou:
That's fantastic. Thanks a lot Alex!

PR #1177, opened 2020-10-16
Post Carousel: Center Items setting

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1172

### DIFF
--- a/widgets/post-carousel/tpl/base.php
+++ b/widgets/post-carousel/tpl/base.php
@@ -34,7 +34,7 @@
 		     data-page="1"
 		     data-fetching="false"
 		     data-dir="<?php echo is_rtl() ? 'rtl' : 'ltr'; ?>"
-		     data-center-items="<?php echo esc_attr( $center_items ); ?>"
+		     data-center-posts="<?php echo esc_attr( $center_posts ); ?>"
 		>
 			<div class="sow-carousel-items">
 				<?php include plugin_dir_path( __FILE__ ) . 'carousel-post-loop.php' ?>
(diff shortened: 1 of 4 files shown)


widgets/post-carousel/tpl/base.php
```diff
@@ -34,6 +34,7 @@
 		     data-page="1"
 		     data-fetching="false"
 		     data-dir="<?php echo is_rtl() ? 'rtl' : 'ltr'; ?>"
+		     data-center-items="<?php echo esc_attr( $center_items ) ?>"
```
Misplon, line 37: Should this be `esc_attr( $center_items );` with a closing `;`?

Misplon:
The `center-carousel` isn't outputting in my test. I have Posts per page set to 3.

carousel

Misplon:
The setting above this is "Loop posts" (we aren't using the word "items"). What do you think about changing the label and description to:

**Center posts**
Applies if the combined width of the visible posts is less than the carousel container width.

widgets/post-carousel/tpl/base.php
AlexGStapleton: `;` isn't required in this situation and I left it out to match the previous line. Regardless, I'll change this (and the previous line).

AlexGStapleton:
How does that look? Previous version would only center the posts if _all_ posts were showing (as in, you couldn't load more).

gregpriday (approved):

Misplon:
@AlexGStapleton Is it expected that Posts per page isn't respected after the first load of additional posts?

Misplon (approved):

AlexGStapleton:
Yes.
For the Post Carousel widget, the Posts Per Page setting label is is misleading and was the reason for [this PR](https://github.com/siteorigin/so-widgets-bundle/pull/1176). It's likely closer to "Posts Per Load" as it's how many posts that are loaded at a time. Are you okay with that label? If so, I'll submit a PR that renames Posts Per Page to that.

Issue #1179, opened 2020-10-16
Google Fonts:  enqueue fonts directly instead of using @import


AlexGStapleton:
We're currently adding Google Font imports directly to the LESS generated CSS files using @import. We should instead enqueue these fonts isntead.

The key improvements this offers are:
- Users can remove them and rely on another plugin to handle the import.
- Users can change when the fonts are imported when using a performance plugin
- Users can be indirectly merged - rahter than four @imports of the same font there will only be the one enqueue.

PR #1185, opened 2020-10-26
Enqueue Google Fonts rather than @import

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1179

`siteorigin_widgets_google_font_fields_X` has been removed, but as it's not the main method of adding fonts I don't see any harm in removing it.

### DIFF
--- a/base/base.php
+++ b/base/base.php
@@ -144,7 +144,12 @@ function siteorigin_widget_get_font($font_value) {
 			$font_url_param .= ':' . $font_parts[1];
 		}
 		$font['url'] = 'https://fonts.googleapis.com/css?family=' . $font_url_param;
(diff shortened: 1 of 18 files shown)


AlexGStapleton:
Doc changes: https://github.com/siteorigin/so-widgets-bundle/pull/1185

Misplon (approved):

gregpriday (approved):

gregpriday:
@AlexGStapleton one question. What happens if a 3rd party plugin has a widget that was using `widget-function('import_google_font');`?

AlexGStapleton:
Nothing. The function is checked to see if it exists before it's called to avoid errors.
I removed all of our usage purely to keep things tidy, not because it would result in an error.

gregpriday (approved):
Perfect

PR #1190, opened 2020-11-05
Elementor styling changes

AlexGStapleton:
This PR does the following:

- Allows for the darkmode text color to be used outside of sections
- Vertically aligns checkboxes with rest of the label text
- Gives text fields a set height so they're more consistent with other fields.

Regarding the first point, Elementor doesn't set a class to indicate whether dark mode is active so we need to rely on elementor to set the text color outside of the sections field.

### DIFF
--- a/base/inc/fields/css/icon-field.less
+++ b/base/inc/fields/css/icon-field.less
@@ -6,6 +6,7 @@
 		display: inline-block;
 		.gradient( #f9f9f9, #f2f2f2, #f9f9f9 );
 		.box-shadow(~"0 1px 2px rgba(0,0,0,0.1)");
+		box-sizing: content-box;
 		.rounded(3px);
 		vertical-align: middle;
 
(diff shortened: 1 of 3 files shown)


Misplon:
Elementor is applying `box-sizing: border-box;` to everything. `content-box` is the default. It's also adding a border-bottom to all links. Mostly the border-box issue was causing the Add Media button to overflow in Elementor. 

We probably should add more global level rules to resolve this. For now, I've just added the declarations in the last commit.

@AlexGStapleton Can you check the Add Media button in Page Builder and the Block Editor situations and confirm it's still normal. Seems ok on my end.

AlexGStapleton:
I'm not seeing any noticeable changes with the Add Media button in either the Classic Editor or Block Editor.

Misplon:
The icon selector change was to resolve the below.

Elementor___Block_Editor

One more quick check for us and then we're good for now.

AlexGStapleton:
Looks good! 👍

Issue #1196, opened 2020-11-09
Accordion: Adding text to text mode and duplicating content results in the text not duplicated


AlexGStapleton:
Steps to replicate:

- Adding a new Accordion item
- Switching the content to text mode
- Paste a large amount of text
- Duplicate the newly added panel
- Copied text won't be present

[Example video](https://drive.google.com/uc?id=1A3mGvmYd-QnsXY288y0wo0D8I5wg9KrJ).

PR #1197, opened 2020-11-09
Repeater TinyMCE: Copy the contents of the text field

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1196

### DIFF
--- a/base/js/admin.js
+++ b/base/js/admin.js
@@ -838,8 +838,13 @@ var sowbForms = window.sowbForms || {};
 							$inputElement.parent().empty().append($inputElement);
 							$inputElement.css('display', '');
 							var curEd = tinymce.get(id);
-							if (curEd) {
-								$inputElement.val(curEd.getContent());
+							if ( curEd ) {
+								var contentVal = curEd.getContent();
+								if ( ! _.isEmpty( contentVal ) ) {
+									$inputElement.val( contentVal );
+								} else if ( contentVal.search( '<' ) !== -1 && contentVal.search( '>' ) === -1) {
+									$textarea.val( contentVal.replace( /</g, '' ) );
+								}
 							}
 						}
 						// Color field :/


Misplon (approved):

base/js/admin.js
```diff
@@ -838,8 +838,13 @@ var sowbForms = window.sowbForms || {};
...
 							$inputElement.css('display', '');
 							var curEd = tinymce.get(id);
-							if (curEd) {
-								$inputElement.val(curEd.getContent());
+							if ( curEd ) {
+								var contentVal = curEd.getContent();
+								if ( ! _.isEmpty( contentVal ) ) {
+									$inputElement.val( contentVal );
+								} else if ( contentVal.search( '<' ) !== -1 && contentVal.search( '>' ) === -1) {
+									$textarea.val( contentVal.replace( /</g, '' ) );
```
gregpriday, line 846: @AlexGStapleton what's this specific line for? Just for my own understanding. A comment might also help here. Is this some sort of tag stripping?

AlexGStapleton: This is actually a duplicate of the TinyMCE form field JS file. https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/inc/fields/js/tinymce-field.js#L121

It searches the content for that has an opening < but not a closing >. In that situation, it strips <. This is done to prevent potentially malformed HTML. In hindsight, I'm not sure how useful either situation (both in this PR and in the initial TinyMCE field) as it could very well cause issues for those who use <.

Hm. Should I remove both instances?

gregpriday: I can't actually offer any solid advice here, but I understand what this is for now. It works for me.

AlexGStapleton: Honestly, this has never come up so I think it's best to just leave it until we get a few reports as this could actually be resolving a weird edge case.

gregpriday (approved):

Issue #1222, opened 2020-12-25
PHP 8 Notices 


AlexGStapleton:
During PHP 8 testing, I experienced a few notices. The notices, and some context as to when they occurred, can be found in this file:

[php-8-notices-wb.txt](https://github.com/siteorigin/so-widgets-bundle/files/5742010/php-8-notices-wb.txt)

Not all of these notices may be directly linked to Widgets, but they all contain references to it or were triggered during a PB specific action. [Related PB issue](https://github.com/siteorigin/siteorigin-panels/issues/838)

PR #1228, opened 2021-01-05
PHP 8 changes

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1222


AlexGStapleton:
A lot of the notices related to really old prebuilt layout data. It's very unlikely any of those notices would happen without importing a prebuilt layout due to how long ago those changes were made.

gregpriday (approved):

PR #1230, opened 2021-01-06
Measurement: Add units and default_unit parameters

AlexGStapleton:
This PR will allow developers to alter the measurements list when adding the form field to their widget. It'll also allow them to change the default unit of measurement to use when either the unit of measurement isn't able to be detected or it's been removed (due to not being in the measurement list).

### DIFF
--- a/base/inc/fields/measurement.class.php
+++ b/base/inc/fields/measurement.class.php
@@ -5,6 +5,21 @@
  */
 class SiteOrigin_Widget_Field_Measurement extends SiteOrigin_Widget_Field_Text_Input_Base {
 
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
To test this, open widgets/headline/headline.php and replace:

```
'font_size' => array(
	'type' => 'measurement',
	'label' => __('Font Size', 'so-widgets-bundle')
),

```
With:

```
'font_size' => array(
	'type' => 'measurement',
	'label' => __('Font Size', 'so-widgets-bundle'),
	'units' => array(
		  'px',
		  '%',
	  	  'em',
	),
),
```

gregpriday (approved):

PR #1232, opened 2021-01-07
Google Maps: Correct Marker JS Property Names

AlexGStapleton:
This PR corrects the marker property names to use the correct camelCase format. They were previously using variable names weren't camelCase so after https://github.com/siteorigin/so-widgets-bundle/pull/1199 they broke.

This PR can be tested by giving a marker a custom icon and the icon won't show. Once the branch is enabled, the icon will show.


Misplon (approved):

Misplon:
Nicely spotted.

gregpriday (approved):

PR #1239, opened 2021-02-08
Contact: Add Number Field

AlexGStapleton:
This PR adds a number field as a Contact Form Field Type.

### DIFF
--- a/widgets/contact/fields/number.class.php
+++ b/widgets/contact/fields/number.class.php
@@ -1,0 +1,5 @@
+<?php
+
+class SiteOrigin_Widget_ContactForm_Field_Number extends SiteOrigin_Widget_ContactForm_Field_Text {
+	// This class just exists for autoloading purposes, but is the same as the text field.
+}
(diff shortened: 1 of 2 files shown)


Misplon (approved):

widgets/contact/fields/number.class.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+class SiteOrigin_Widget_ContactForm_Field_Number extends SiteOrigin_Widget_ContactForm_Field_Text {
+	// This class just exists for autoloading purposes, but is the same as the text field.
```
gregpriday, line 4: Couldn't this output a `<input type="number">` instead?

https://www.w3schools.com/tags/att_input_type_number.asp

gregpriday (changes requested):

widgets/contact/fields/number.class.php
AlexGStapleton: This field will output as a number (in other words, using the markup you mentioned). The  `SiteOrigin_Widget_ContactForm_Field_Text` class will output the input type as whatever is after `SiteOrigin_Widget_ContactForm_Field_`. So if I set it to Example, it would output with a type of `example`.

https://github.com/siteorigin/so-widgets-bundle/blob/4d3718f59f90b13a89e559fc0e1f046bc787738b/widgets/contact/fields/text.class.php#L8

This is how the [tel](https://github.com/siteorigin/so-widgets-bundle/blob/4d3718f59f90b13a89e559fc0e1f046bc787738b/widgets/contact/fields/tel.class.php) and [email](https://github.com/siteorigin/so-widgets-bundle/blob/4d3718f59f90b13a89e559fc0e1f046bc787738b/widgets/contact/fields/email.class.php) fields work.

gregpriday: Aaah, I see. Perfect!

gregpriday (approved):

PR #1243, opened 2021-02-12
Image: Add Support for Custom Image Size

### LINKED ISSUE #1241: Image Sizes: Add Custom Size Option
We should add an optional parameter for the Image Sizes Form Field that allows the user to specify a Custom Image Size. This would be to how the WordPress Image widget handles it with slight useability differences. The Image Widget handles it via an edit prompt, but for more overall control (and useability - as you don't need to open the image to see the sizes) we should introduce two visible width x height text number fields that show up after selecting the Custom Size option. If the user is still after more options, they can manage the images via the Media Library directly.

As for how the Image Widget handles custom sizes when outputting the image, it's output at exactly the specified image size, but it doesn't create an image size for the custom size. Instead, it choices the closest existing image size and uses that instead. That functionality is already in place so we should replicate that also.

AlexGStapleton:
Resolves #1241

Example screenshot

Related links:
[Demo PR](https://github.com/siteorigin/so-widgets-bundle/pull/1244) (requires this PR also - use this for testing)
[Docs](https://github.com/siteorigin/docs/pull/92)

### DIFF
--- a/base/inc/fields/css/image-size-field.less
+++ b/base/inc/fields/css/image-size-field.less
@@ -1,0 +1,11 @@
+
+.siteorigin-widget-form .siteorigin-widget-field-size {
+	
+	.custom-size-wrapper {
+		margin-top: 15px;
+
+		.siteorigin-widget-input {
+			width: 80px;
+		}
+	}
+}
--- a/base/inc/fields/css/image-size-field.less
+++ b/base/inc/fields/css/image-size-field.less
@@ -4,6 +4,11 @@
 	.custom-size-wrapper {
 		margin-top: 15px;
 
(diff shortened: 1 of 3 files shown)


Misplon:
@AlexGStapleton I think that's bold the Width and Height labels.

I take it crop is set to soft?

AlexGStapleton:
I've bolded those labels:

Screenshot_2021-03-05 Edit Page ‹ SiteOrigin — WordPress

Misplon (approved):

gregpriday (approved):
Looks good. Are we not going to use this with our image widget?

Issue #906, opened 2019-09-03
Add setting to always show slider navigation, or expand nav_always_show_mobile to drop down


AlexGStapleton:
A number of requests have come in about always showing the Slider navigation. Due to the changes made in #815, this could be done by changing [nav_always_show_mobile](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/inc/widgets/base-slider.class.php#L94) to a drop down and adding an additional conditional to [the relevant slider JS](https://github.com/siteorigin/so-widgets-bundle/blob/67907a25c3115f2de8fe993bf1bdc8345401e4fb/js/slider/jquery.slider.js#L204). Alternatively, a dedicated setting could be introduced to control the desktop version independently of the mobile setting.

Misplon:
@AlexGStapleton I'm keen to see this rolled out, I'm bumping to p2.

PR #1252, opened 2021-03-02
Sliders: Add Always show navigation on desktop setting

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/906

I went with a separate setting as implementing migration code would require modifying each individual widget (less than ideal as the central slider class was introduced to avoid that need). It would be a good idea to remove the slider dependency on [wp_is_mobile()](https://developer.wordpress.org/reference/functions/wp_is_mobile/) as that isn't cache-friendly, and it's redundant due to the slider breakpoint.

### DIFF
--- a/js/slider/jquery.slider.js
+++ b/js/slider/jquery.slider.js
@@ -186,22 +186,26 @@ jQuery( function($){
 				$p.add($n).hide();
 				if( $slides.length > 1 ) {
 					if( !$base.hasClass('sow-slider-is-mobile') ) {
-
-						var toHide = false;
-						$base
-							.on( 'mouseenter', function() {
-								$p.add($n).clearQueue().fadeIn(150);
-								toHide = false;
-							})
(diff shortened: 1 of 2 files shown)


Misplon (approved):

js/slider/jquery.slider.js
```diff
@@ -186,22 +186,26 @@ jQuery( function($){
...
-								toHide = false;
-							})
-							.on( 'mouseleave', function() {
-								toHide = true;
-								setTimeout(function(){
-									if( toHide ) {
-										$p.add($n).clearQueue().fadeOut(150);
-									}
+						if ( settings.nav_always_show_desktop && window.matchMedia( '(min-width: ' + settings.breakpoint + ')' ).matches ) {
+						$p.show();
```
gregpriday, line 190: Shouldn't these lines be indented one deeper?

Issue #866, opened 2019-05-05
Add integration for Swift Performance


Misplon:
Add integration for Swift Performance. Requested on WordPress.org
https://wordpress.org/support/topic/swift-performance-compatibility/

PR #1258, opened 2021-03-18
Improve cache plugin support

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/866

To test this please add the following PHP to [this line](https://github.com/siteorigin/so-widgets-bundle/pull/1258/files#diff-2dc7c68c2b6e69aef4ae1d74975cb590295990126bcd9a32c7f9614f0e0a6f0fR336).

`$force_delete = true;`

Then view WP Admin. View a page that has had its widget CSS cleared and new widget CSS should be generated (as well as the cache regenerated but that's harder to check). If those plugins aren't alerted (and minify isn't managed through their plugins directly) you'll end up with a bunch of 404 notices. If it's working as expected, the page display as expected, and new CSS files will be in the `wp-content/uploads/siteorigin-widgets` directory. Please note that you may need to clear your browser cache for the cache plugin to be able to regenerate the page .

To do:

- [x] Add Support for W3 Total Cache
- [x] Clear CSS Cache while viewing WP Admin (this will ensure that cache plugins are able to be reliably alerted).
- [x] Improve widget area support for new cached CSS filename.

### DIFF
--- a/compat/compat.php
+++ b/compat/compat.php
@@ -90,7 +90,6 @@ class SiteOrigin_Widgets_Bundle_Compatibility {
 			w3tc_flush_all();
 		}
 
-
 		if ( class_exists( 'Swift_Performance_Cache' ) ) {
 			Swift_Performance_Cache::clear_all_cache();
 		}
--- a/compat/compat.php
+++ b/compat/compat.php
@@ -63,10 +63,9 @@ class SiteOrigin_Widgets_Bundle_Compatibility {
 	/**
 	 * Tell cache plugins that they need to regenerate a page cache.
 	 *
-	 * @param $excluded
-	 * @param $content
+	 * @param $name The name of the file that's been deleted.
+	 * @param $instance The current instance of the related widget.
 	 *
-	 * @return string
 	 */
 	public function clear_page_cache( $name, $instance = array() ) {
 		$id = end( explode( '-', $name ) );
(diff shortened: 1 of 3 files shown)


AlexGStapleton:
@Misplon Due to https://github.com/siteorigin/so-widgets-bundle/pull/1258/commits/5fb16f14ad683cba6c4615508b70ab891289e858, widgets added to the sidebar will currently generate a new stylesheet for every single page. This, obviously, isn't ideal.

 Currently, I'm thinking of using "widgetarea" (or just widget) instead of the page id for those widgets and when those widgets expire it'll trigger a complete regenerate. With that said, I'm going to come back to this once I've had some time to think about how to avoid this as there may be a better way to do this.

Misplon:
@AlexGStapleton Sounds good, let us know what you find or the conclusion you reach.

AlexGStapleton:
@Misplon I've added instructions for how to test this PR. Let me know if you would like me to add any other plugins also (WP Super Cache, WP Rocket, etc) in this PR (rather than saving them for a future PR).

ictbeheer:
Regarding WP Rocket:
- To clean the cache of a single post you can use `rocket_clean_post( $post_id );`
- To clean the complete cache you use `rocket_clean_domain()` (Maybe followed by `rocket_clean_minify('css');` to clean the minified css files)
--

AlexGStapleton:
Hi @ictbeheer,

Awesome. Thanks mate!
I've talked to @Misplon about this and we've decided to add additional cache plugins in future PRs to keep this PR focused on the initial goal of this PR. If you're keen, consider submitting a PR to add WP Rocket support in the future. :)

compat/compat.php
```diff
@@ -54,6 +59,49 @@ function is_active( $builder ) {
...
+	 * @param $content
+	 *
+	 * @return string
+	 */
+	public function clear_all_cache() {
+		if ( function_exists( 'w3tc_flush_all' ) ) {
+			w3tc_flush_all();
+		}
+
+
```
Misplon, line 99: Looks like an extra empty line here.

gregpriday (approved):

Issue #850, opened 2019-03-26
SiteOrigin Widget: If widget has been disabled, enable it


AlexGStapleton:
Currently, when a SiteOrigin Widget block is created a widget is selected, disabling that widget will result in a missing message stating the widget has been disabled. The SIteOrigin Widget block should do what the SiteOrigin Layout Block does and enable the widget automatically rather than showing a missing widget message.

[Slack](https://siteorigin.slack.com/archives/C02SYBBH2/p1553623620046200)

PR #1259, opened 2021-03-22
Widgets Block: Activate Widgets Automatically

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/850

To test:
- Deactivate Headline widget
- Attempt to use it in a SiteOrgin WIdgets Block, and widget will be activated.
- Add some placeholder data and then save.
- Deactivate Headline widget.
- Reload Block Editor with recently added SiteOrigin Widgets Block with Headline widget.
- Widget will be activated and should preview as expected and the form should be accessiable.

### DIFF
--- a/base/inc/routes/siteorigin-widgets-resource.class.php
+++ b/base/inc/routes/siteorigin-widgets-resource.class.php
@@ -90,7 +90,6 @@ class SiteOrigin_Widgets_Resource extends WP_REST_Controller {
 		$widget = SiteOrigin_Widgets_Widget_Manager::get_widget_instance( $widget_class );
 		// Attempt to activate the widget if it's not already active.
 		if ( ! empty( $widget_class ) && empty( $widget ) ) {
-			global $wp_widget_factory;
 			$widget = SiteOrigin_Widgets_Bundle::single()->load_missing_widget( false, $widget_class );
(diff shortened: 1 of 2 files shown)


Misplon (approved):

base/inc/routes/siteorigin-widgets-resource.class.php
```diff
@@ -88,6 +88,11 @@ public function get_widget_form( $request ) {
 
 		/* @var $widget SiteOrigin_Widget */
 		$widget = SiteOrigin_Widgets_Widget_Manager::get_widget_instance( $widget_class );
+		// Attempt to activate the widget if it's not already active.
+		if ( ! empty( $widget_class ) && empty( $widget ) ) {
+			global $wp_widget_factory;
```
gregpriday, line 93: @AlexGStapleton why are you declaring this global?

gregpriday (approved):
Just a question about a global, but everything looks good.

base/inc/routes/siteorigin-widgets-resource.class.php
AlexGStapleton: Shouldn't be. Corrected.

PR #1262, opened 2021-03-28
Image Grid: Display Using Flexbox to Prevent Unintended Spacing

AlexGStapleton:
This PR prevents undesired output due to whitespace being rendered by browsers. This happens due to how browsers output `display: inline-block;` and that doesn't happen when using flexbox. This unintended spacing can make it trickier to get the desired amount of spacing because the horizontal spacing will always be more than the vertical spacing.

To test:
- Add a series of images that you can easily tell the sides from (I used [this image](https://i.imgur.com/rdUR6n7.png))
- Set the Spacing to 0 
- Inspect resulting output
- Clear wp-content/uploads/siteorigin-widgets or make a design change to the Image Grid widget (this is required to regenerate the generated CSS stylesheet).
- Inspect resulting output

Before:
Screenshot_2021-03-29 SiteOrigin

After:
Screenshot_2021-03-29 SiteOrigin(1)

### DIFF
--- a/widgets/image-grid/styles/default.less
+++ b/widgets/image-grid/styles/default.less
@@ -1,6 +1,9 @@
 @spacing: default;
 
 .sow-image-grid-wrapper {
+	display: flex;
+	flex-wrap: wrap;
+	justify-content: center;
 	padding-top: @spacing;
 	line-height: 0;
 	text-align:center;
@@ -8,10 +11,6 @@
 	.sow-image-grid-image {
 		padding: 0 @spacing @spacing @spacing;
 		display: inline-block;
-
-		> a {
-			display: inline-block;
-		}
 
 		img {
 			opacity: 0;


AlexGStapleton:
Please note that if you set a spacing value larger than one the spacing won't be completely equal [due to top spacing not being applied to images](https://github.com/siteorigin/so-widgets-bundle/blame/develop/widgets/image-grid/styles/default.less#L9).

Misplon:
@AlexGStapleton Shouldn't top padding only be zero for the images at the top of the grid? Should we sort this out? There are responsive concerns. We'll need to consider which image is at the top for each breakpoint. Separate PR?

AlexGStapleton:
Yes. I think that should be accounted for in a separate PR.

Misplon:
Thanks, mate. Please, log and self-assign.

Misplon (approved):

Issue #1111, opened 2020-07-29
New Widget: Anything Carousel


Misplon:
**Item Repeater**

- Title
- Content

**Carousel Settings**

- Previous/Next navigation arrows
- Navigation dots
- Loop items. Automatically return to the first item after the last item.
- Autoplay
- Autoplay pause on hover
- Animation speed
- Timeout

**Design**

- Navigation color
- Item padding
- Item margin

**Responsive**

The same group of settings being used on the Post Carousel widget but with `Slides to show` added to each section.

AlexGStapleton:
Do you want to add a dedicated title field for the widget (like the post carousel widget)?

Misplon:
Yes, I think we should.

AlexGStapleton:
We're currently using a filter for the breakpoints in the post carousel widget. Now that there's going to be two widgets that use it, I think we should set up dedicated fields global widget settings fields for them. Thoughts?

Misplon:
Sounds good, thanks.

PR #1265, opened 2021-04-06
Base Carousel Widget

AlexGStapleton:
This is the foundation of https://github.com/siteorigin/so-widgets-bundle/issues/1111

- Styling of navigation is left up to the widget. With that said, it's likely to be expanded in future commits/PR.

In hindsight, the Post Carousel Commit could have been spun off into a separate PR as it does exceed the scope of the title of the PR. Having it in this PR does however make it much easier to test. The Anything Carousel will definitely be in a separate PR.


AlexGStapleton:
@Misplon Regarding the responsive settings, it makes sense to add a copy of all of the global widget settings. What are your thoughts on doing that? If you're keen, how do you want to handle the widget specific responsive settings? We could handle the fields like page builder and have an empty field mean the default value is used.

Misplon:
Sounds good. If we could add in placeholder support though, that would be ideal vs blank fields when the default is being used.

AlexGStapleton:
The last commit adds global responsive fields. It uses placeholders when the user is editing widgets to prevent the global settings from being saved.

`get_breakpoints()` is present to allow for continued support of the `siteorigin_widgets_post_carousel_breakpoints` filter. You can use the following code to confirm that filter still works as expected:

add_filter( 'siteorigin_widgets_post_carousel_breakpoints', function( $breakpoints ) {
	return array(
		'tablet_landscape' => 333,
		'tablet_portrait'  => 222,
		'mobile'           => 111,
	);
} );

A global `siteorigin_widgets_carousel_breakpoints` filter doesn't have a use case due to the responsive settings now being editable via widget settings. @Misplon If you're keen though, I can add that filter.

Misplon:
Thanks, mate. We can add that filter at some point if we get a request.

Misplon (approved):

Misplon:
@AlexGStapleton Can you take a look at the above conflict? Thanks, mate.

AlexGStapleton:
@Misplon The last two commits could result in the Post Carousel styling not working as expected so re-testing will be required.

base/inc/widgets/base-carousel.class.php
```diff
@@ -0,0 +1,260 @@
...
+					plugin_dir_url( SOW_BUNDLE_BASE_FILE ) . 'css/lib/slick.css',
+					array(),
+					'1.8.1'
+				),
+			)
+		);
+
+	}
+
+	// Allow widgets to override settings.
```
gregpriday, line 40: Same here, just need to stick to using PHPDoc format.

gregpriday (changes requested):

gregpriday (approved):

Misplon:
@AlexGStapleton Just a small conflict here, thanks mate.

Issue #1268, opened 2021-04-11
Icons: Alter filter options to be same height


AlexGStapleton:
<img width="914" alt="Edit_Page_‹_Puro_—_WordPress" src="https://user-images.githubusercontent.com/17275120/114306708-0d570e80-9b20-11eb-9d4f-09f00ab79fd6.png">

PR #1270, opened 2021-04-12
Icon: Ensure Filter options are the same height

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1268

Vertical Align adjustment is to prevent a slight offset of the selectors. A build is required for this PR to be tested.

### DIFF
--- a/base/inc/fields/css/icon-field.less
+++ b/base/inc/fields/css/icon-field.less
@@ -81,10 +81,11 @@
 		.box-shadow(~"0 1px 2px rgba(0,0,0,0.0275)");
 		margin-top: 10px;
 
-		select.select.siteorigin-widget-icon-family,
+		select.siteorigin-widget-icon-family,
 		select.siteorigin-widget-icon-family-styles,
 		.siteorigin-widget-icon-search {
 			height: 30px;
+			line-height: 30px;
 			vertical-align: top;
 		}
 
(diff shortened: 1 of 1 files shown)


Misplon:
Thanks, mate. Search icons placeholder looks a little low down in the input.

Block Editor:
Screenshot 2021-05-01 at 13 36 25

What have we done elsewhere with regards to line-height? The core is setting line-height to 2. If we set it to 30px, the issue is resolved.

AlexGStapleton:
How does that look? The last commit should resolve this, and should also correct the sizing of the icon family in the Block Editor.

Misplon:
The issue remains:

Screenshot_2021-05-03_at_12_54_14_png___500___Layer_1__RGB_8____

Line height needs to be set to 30px.

AlexGStapleton:
How about that? Oddly, I was getting the invese (the search field text was 1px higher).

Misplon:
@AlexGStapleton Doing it this way seems to get everything aligned nicely for me. Can you spot any issues with my change?

AlexGStapleton:
That's caused a very slight misalignment for me in the Block Editor. Here's a video of the difference:

https://drive.google.com/uc?id=1_dp7U53lX_FftlCnE_QzO9Ctr71vyfr1

Were you having issues with the previous location of the line-height?

Misplon:
It wasn't right without the line height for me. On my display it now looks correct in both editors:

Block
block

Classic
classic

The alignment in your video after the shift when the select placeholders shift up, that alignment is correct. Here is it measured from the video:

video

Misplon (approved):

PR #1271, opened 2021-04-12
Slider: Add "Loop slide background videos" setting

AlexGStapleton:
This PR adds a Slider Controls setting that allows the user to prevent the looping of background videos. The setting name is based on the setting just above this setting ('Show slide background videos on mobile').

### DIFF
--- a/base/inc/widgets/base-slider.class.php
+++ b/base/inc/widgets/base-slider.class.php
@@ -332,7 +332,7 @@ abstract class SiteOrigin_Widget_Base_Slider extends SiteOrigin_Widget {
 					$classes[] = 'sow-mobile-video_enabled';
 				}
 
-				$this->video_code( $background['videos'], $classes );
+				$this->video_code( $background['videos'], $classes, $controls );
 			}
 
 			if( $background['opacity'] < 1 && !empty($background['image']) ) {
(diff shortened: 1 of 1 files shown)


Misplon:
Thanks, mate. If Autoplay is provided as an individual setting available per video, shouldn't looping also be a per video setting?

AlexGStapleton:
I've run a few tests and it's not possible to loop by source. In other words, the user can't add two different video sources to the same slide and only have one of them loop. We can however set up a slide specific loop. If that's okay, where would you like to add that setting? Just below the Background Videos repeater?

Misplon:
* Thanks. Let's leave the setting where it is.

* When I switch to the branch the following video stops looping even though I have looping enabled.

https://user-images.githubusercontent.com/789159/118098124-d05a9200-b3d3-11eb-9938-e98247f42a90.mp4

* Which video types does this setting control?

AlexGStapleton:
> When I switch to the branch the following video stops looping even though I have looping enabled.

Fixed.

> Which video types does this setting control?

As of the last commit, both. Requires https://github.com/siteorigin/so-widgets-bundle/pull/1298

Misplon:
@AlexGStapleton Are you sure default true is the right way to go for this setting?

Misplon (approved):

gregpriday (approved):

PR #1276, opened 2021-04-16
Font: Allow italic variant with no font weight

AlexGStapleton:
This is a follow up to https://github.com/siteorigin/so-widgets-bundle/pull/1237. That PR relied on a number always being present before italic, but that doesn't always happen. This PR will allow for all possible variant formats. For example:

Alegreya:italic
Alegreya:500
Alegreya:500italic


Misplon (approved):

Misplon:
I wasn't able to select Alegreya:italic using the master branch, it defaults to Alegreya:regular. PR seems to fix the issue.

gregpriday (approved):

PR #1277, opened 2021-04-16
Post Carousel: Prevent loading more items if complete

AlexGStapleton:
Due to https://github.com/siteorigin/so-widgets-bundle/pull/1220, there's a situation where the number of items may be higher after initial load (this is corrected after loading more posts). This PR accounts for this and will prevent more posts from being loaded if the number of items exceeds the total number of posts.


AlexGStapleton:
To test this, use the filter introduced in #1220 and set a value that's lower than the number of posts initially output and not equal to the number of items visible at a time. So, for example, if 12 posts were set to output and there are visible at once, use 7, 11, etc.

Misplon (approved):

gregpriday (approved):

PR #1281, opened 2021-04-21
Anything Carousel

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1111
Requires https://github.com/siteorigin/so-widgets-bundle/pull/1265


Misplon:
Thanks, mate. Further feedback to follow.

Are you keen to perhaps set the repeater titles to match the text inserted in the Title field?

AlexGStapleton:
That was intended, but it wasn't working as expected due to a typo. The last commit corrects that issue.

Misplon:
* Let's enable Navigation dots and Looping by default.
* Let's go with 3 slides on desktop, 3 on tablet landscape, 2 on portrait, and 1 on mobile.

Are the slides to show settings working for you?

AlexGStapleton:
>Let's enable Looping by default.

Resolved by https://github.com/siteorigin/so-widgets-bundle/pull/1265/commits/139691220443987a13356d1634b7cfb4335b34bc

> Let's go with 3 slides on desktop, 3 on tablet landscape, 2 on portrait, and 1 on mobile.

Resolved by https://github.com/siteorigin/so-widgets-bundle/pull/1265/commits/1e3c9f645d9e2efcf1d630c70d8ffe991ce662e7.

Please note that if you've saved the global settings the responsive values won't be updated. Pre-existing widgets also won't be updated.

> Are the slides to show settings working for you?

Yes.

Misplon:
@AlexGStapleton Here is our first theme from Ric.

Anything Carousel@2x

Here are the left/right buttons:

[nav-arrows.zip](https://github.com/siteorigin/so-widgets-bundle/files/6662173/nav-arrows.zip)

Alternatively, here are the arrows only:

[arrows-only.zip](https://github.com/siteorigin/so-widgets-bundle/files/6662269/arrows-only.zip)

Ric's suggestion is to try making the nav arrow and circle `#000` on hover. If we use the arrows only we'd need to create the circle using CSS whichever you think is best.

* Nav arrow color: `#626262`.
* Nav arrow circle color: `#bebebe`.

The carousel spacing:

* 68px between each item.
* 144px bottom margin below the items/above the nav dots.

The arrow styling is as follows:

* 64px x 64px.
* 25px left/right margin from the closest carousel item.

The nav dots are as follows:

* 24px by 24px.
* 16px apart.
* Default color: `#bebebe`.
* Selected and hover color: `#f14e4e`.

Misplon:
To handle the above theme requirements what do you think about removing `Navigation color` and adding:

Navigation Arrows Color
Navigation Arrows Hover Color
Navigation Dots Color
Navigation Dots Selected and Hover Color

* Should we keep Item padding? Is it necessary? 

* How do you want to handle the Navigation dots margin? Will it just be handled by the item bottom margin?

AlexGStapleton:
To confirm, do themes work like the Button widget, or like the Accordion/Tab presets? Based on the outline it sounds like the latter (specifically due to the setting values), but I had thought it was going to be closer to the former so I want to double-check before proceeding

 I should note that if the intention is for it to be a preset rather than a theme, the preset selected isn't stored so it may make it trickier for us to implement dedicated settings. We may need to alter how that field works to ensure the preset is stored.

> To handle the above theme requirements what do you think about removing Navigation color and adding:

Sounds good.

> Should we keep Item padding? Is it necessary?

I think it's nice to have, but I don't think it's a must-have. If you would prefer to keep the settings on the lighter side, we definitely can remove it.

> How do you want to handle the Navigation dots margin? Will it just be handled by the item bottom margin?

Yes. While we could add dedicated spacing, I think it'll only make it harder for people to get the desired spacing. Relying on the one setting will also the avoid potential "I set this to 0 but there's still spacing present" question.

Misplon:
Super, thanks.

Just noting our Slack chat here, it sounds like we're more along the lines of the Tabs themes with conditional settings thrown in there.

Let's remove item padding for this base theme.

Agreed on the navigation dots margin.

AlexGStapleton:
You mentioned base theme, do you mean that the user _has_ to select a theme? Not selecting a theme could allow the user to adjust any (well, within reason) design settings and it would effectively look how it currently is. If you want them to use a theme, that's completely fine, I'll just need to make some adjustments to the presets field to require selection - it might be worth duplicating the presets form field and calling the duplicated version to themes/skin.

Are those arrows specific to this theme (well, specific at the moment anyway), or are the new general arrows for the carousel?

Regarding the circle, we could make this an optional checkbox setting - Arrow Circle Background.

> Ric's suggestion is to try making the nav arrow and circle #000 on hover

Wouldn't that make the arrow impossible to see, or is that the intention?

Misplon:
**Are those arrows specific to this theme (well, specific at the moment anyway), or are the new general arrows for the carousel?**

Yes, the arrows, the navigation dots, the default spacing, it all makes up this theme and is specific to this theme.

**Regarding the circle, we could make this an optional checkbox setting - Arrow Circle Background.**
 
Let's come back to this one, perhaps, if it comes up as a request.

**Ric's suggestion is to try making the nav arrow and circle `#000` on hover.**

Ric is referring to the color of the navigation circle, the border, and the color of the navigation arrow. To confirm, we're not talking about filling in the circle on hover, just changing the color of the circle border and the color of the arrow.

Misplon:
Regarding the nav arrow hover styling, we can always tweak that once we're a bit further along based on how it turns out.

AlexGStapleton:
The base "theme" was added in the last commit. You'll need to pull the latest version of https://drive.google.com/uc?id=1bpLeHlaE4UxDdKRdhelk02Mn5-54UGIB as that has also received a couple of commits.

>The arrow styling is as follows: 64px x 64px.

Is that right? That looks way larger than the mockup. I ballparked the font-size of the arrow so you may need to adjust it to get the correct sizing. That CSS is here.

https://github.com/siteorigin/so-widgets-bundle/blob/97d2eaf29c5de6336471886a0f517e41a7b23b89/widgets/anything-carousel/styles/base.less#L32-L36

I removed the Item Padding. We should consider re-adding it in the future, but as the current styling doesn't use it, it didn't make much sense to include it.

---

Please note that due to the setting changes, you may see notices related to missing fields. That'll be resolved once saving and it won't occur for users as they won't have used the older version of this widget. The item margin default has been changed but for any pre-existing Anything Carousels, this change won't be reflected so you won't have the new item margin.

[Here's a test layout](https://drive.google.com/uc?id=1bpLeHlaE4UxDdKRdhelk02Mn5-54UGIB) based on the most recent version of this widget.

Misplon:
My bad, brother. Looks like I'm looking at double-size dimensions here. Let me confirm with Ric and come back to you.

Misplon:
@AlexGStapleton Hopefully, I have the wire theme sorted now. Let me know if you spot any issues. Here is the design we're working with: https://www.figma.com/file/YLHlNwnOJvMYVNzRPpZ9ks/Anything?node-id=0%3A31.

Misplon:
@AlexGStapleton What do you think about adding new font settings and re-organizing our settings as follows:

# Design

## Title
- HTML tag
- Font
- Size
- Color

## Item
- Font
- Size
- Color
- Margin

## Navigation
- Navigation arrows color
- Navigation arrows color hover
- Navigation dots color
- Navigation dots selected and hover color

I'm not sure if the nav settings need the "Navigation" prefix once under their own heading.

AlexGStapleton:
@Misplon I don't think the Navigation settings should include Navigation in their names as they're all in the same section so there's little room for potential confusion.

Are the titles going to be managed by the theme? (asking so I know where to add a `default` or not) Do you have defaults/values in mind?

Misplon:
I agree regarding the navigation labels.

Yeah, let's leave the title as we do with the Features widget.

AlexGStapleton:
@Misplon Regarding the item values:

> Item
    Font
    Size
    Color

- Is Color for the text or the item background? If it's the text I think we should call it Text color as users may think it's the background (I don't think it is, but I'm double-checking to be sure)
- I think we should rename the Size setting to Font Size as it's not clear whether Size means the size of the item or the font size.
- Are these general item styling, or are they specific to the titles? If general item styling, should we introduce Title settings for the Item Titles?

AlexGStapleton:
The last commit required a change to the base carousel so please ensure the latest version of https://github.com/siteorigin/so-widgets-bundle/pull/1265 is being used.

It applies the Item styling as general styling rather than specific to the title - this can be easily changed.

AlexGStapleton:
@Misplon Please note that the last commit will briefly result in notices as the Design settings drastically changed. It'll only occur for us as we previously used the old settings, normal users won't experience these notices. Simply save the widget once and they'll be gone.

Misplon:
> Is Color for the text or the item background? If it's the text I think we should call it Text color as users may think it's the background (I don't think it is, but I'm double-checking to be sure)

Cool, let's change to Text color.

> I think we should rename the Size setting to Font Size as it's not clear whether Size means the size of the item or the font size.

Cool.

> Are these general item styling, or are they specific to the titles? If general item styling, should we introduce Title settings for the Item Titles?

Not quite following here. The titles have their own section, below the item title is the item, it has its own design section.

AlexGStapleton:
@Misplon We discussed the title settings over Slack and decided to rename the Title settings to Item title.

Misplon (approved):

Misplon:
@AlexGStapleton Can you add a prompt to upgrade for the Carousel Addon?

AlexGStapleton:
Done. I based the wording on the Accordion teaser, adjust it as needed.

gregpriday (approved):

PR #1283, opened 2021-04-23
Slider: Only apply foreground height if background_image is set

AlexGStapleton:
[Reported here](https://siteorigin.com/thread/siteorigin-widgets-bundle-1-18-1/). Will submit for review once user replies.

### DIFF
--- a/widgets/slider/slider.php
+++ b/widgets/slider/slider.php
@@ -191,7 +191,7 @@ class SiteOrigin_Widget_Slider_Widget extends SiteOrigin_Widget_Base_Slider {
 							<?php endif; ?>
 						<?php endforeach; ?>>
 					<?php endif; ?>
-					<div class="sow-slider-image-foreground-wrapper" style="<?php if ( ! empty( $frame['background_image'] ) && ! empty( $foreground_src[2] ) ) echo 'height: ' . intval( $foreground_src[2] ) . 'px;'; ?>">
+					<div class="sow-slider-image-foreground-wrapper" style="<?php if ( ! empty( $frame['custom_height'] ) && ! empty( $foreground_src[2] ) ) echo 'height: ' . intval( $foreground_src[2] ) . 'px;'; ?>">
 						<?php
 						echo siteorigin_widgets_get_attachment_image(
 							$frame['foreground_image'],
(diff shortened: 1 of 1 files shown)


Misplon (approved):

AlexGStapleton:
I've discovered a potential issue that can still arise if a user has a background image and a foreground image. If the foreground image is larger than the background, the slider will be increased in height. This didn't happen previously so it's an unintended change. The last commit will fix this.

To test that commit:

- Add a background image and then view the slider. Note the height of the slider.
- Edit the slider again and set a foreground image that's larger than the Background Image (ideally, the foreground image should be significantly larger so it's obvious).
- Save, and view the slider. You'll notice that prior to that commit the slider would be larger than it was previously. Now it's the same size as it would have been pre-1.18.x. 
- lease set a Height to confirm that setting is still working as expected.

Misplon:
Once I set a height, the last step, the slider changes, and the foreground image is no longer visible.

AlexGStapleton:
I'm having trouble replicating that. Can you please provide me with a copy of the images you're using, and clarify what Height you've set?

--

A user has reported over email that the last commit resolved the spacing issue they were facing, and the larger slider after update (issues were on different sites).

gregpriday (approved):

Misplon:
For the foreground try:



For the background:

ave-calvar-ltuLbi-Q1hM-unsplash

Check the slider before setting a height value. Then set the height to 500.

Misplon:
For the foreground try:

ruud-luijten-4MHr9v1jVEA-unsplash

For the background:

ave-calvar-ltuLbi-Q1hM-unsplash

Check the slider before setting a height value. Then set the height to 500.

AlexGStapleton:
Thanks, I've tried those values and the only thing I see with those settings is the foreground image with the background image being completely covered by the foreground image. Are you able to replicate this on another install?

Misplon:
Testing on the demo server now...

Misplon:
Is this an expected result? https://demo.siteorigin.com/corp/slider-sandbox/

AlexGStapleton:
Okay, I see what's going on - I wasn't testing with a sidebar so the image was displaying at a normal size for me. The last commit will apply the height directly to the image instead of the foreground container. How does that look?

Misplon (approved):

PR #1284, opened 2021-04-25
Widgets Block: Ensure Icons are enqueued

AlexGStapleton:
[Reported here](https://wordpress.org/support/topic/symbol-on-buttons-not-visible/)

This PR resolves an issue where users who use the Widgets Block not displaying icons correctly.

This issue was caused by https://github.com/siteorigin/so-widgets-bundle/pull/1234. Unlike fonts, icons are loaded in the template itself so they won't be enqueued by [the standard widget style enqueue](https://github.com/siteorigin/so-widgets-bundle/blob/1.18.1/compat/block-editor/widget-block.php#L133). That's not something we can change at this point as it's very likely other developers are doing it in the same manner as us, so instead this PR ensures icons are enqueued by storing the icons in an attribute and outputting them when rendering.

Only those who made a change to the SiteOrigin Widgets Block on 1.18.x, and used a widget that added an icon are impacted. Those users will need to resave the layout after updating for the newer data to be saved and rendered. Those who haven't made any changes since 1.18.x will be using the old PHP render so will be unaffected by this issue. 

To test this PR please add a SiteOrigin Widgets Block and select a widget that uses icons (eg. Button, Icons, Accordion, etc). Save and view the widget on the front end. The icons won't be present. Switch to this branch and then reload the editor (to allow for the updated JS). Save, and then view the page again.

### DIFF
--- a/compat/block-editor/widget-block.php
+++ b/compat/block-editor/widget-block.php
@@ -155,6 +155,17 @@ class SiteOrigin_Widgets_Bundle_Widget_Block {
 			} else {
 				$widget->generate_and_enqueue_instance_styles( $instance );
 				$widget->enqueue_frontend_scripts( $instance );
+
+				// Check if this widget uses any icons that need to be enqueued.
+				if ( ! empty( $attributes['widgetIcons'] ) ) {
+					$widget_icon_families = apply_filters('siteorigin_widgets_icon_families', array() );
+					foreach ( $attributes['widgetIcons'] as $icon_font ) {
+						if ( ! wp_style_is( $icon_font ) ) {
+							$font_family = explode( 'siteorigin-widget-icon-font-', $icon_font )[1];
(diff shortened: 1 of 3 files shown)


compat/block-editor/widget-block.php
```diff
@@ -155,6 +155,17 @@ public function render_widget_block( $attributes ) {
 			} else {
 				$widget->generate_and_enqueue_instance_styles( $instance );
 				$widget->enqueue_frontend_scripts( $instance );
+
+				// Check if this widget uses any icons that need to be enqueued.
+				if ( ! empty( $attributes['widgetIcons'] ) ) {
+					$widget_icon_families = apply_filters('siteorigin_widgets_icon_families', array() );
```
Misplon, line 161: Small spacing fix here `apply_filters('siteorigin_`

Misplon (approved):

gregpriday (approved):

PR #1303, opened 2021-05-16
Video Player: Allow External Videos to Loop

AlexGStapleton:
Requires https://github.com/siteorigin/so-widgets-bundle/pull/1298

Prior to this PR, the Loop setting would disappear if an External Video was selected.

### DIFF
--- a/widgets/video/tpl/default.php
+++ b/widgets/video/tpl/default.php
@@ -58,7 +58,7 @@ do_action( 'siteorigin_widgets_sow-video_before_video', $instance );
 		<?php endforeach; ?>
 	</video>
 	<?php else : ?>
-	<?php echo $so_video->get_video_oembed( $src, $autoplay, $related_videos ); ?>
+	<?php echo $so_video->get_video_oembed( $src, $autoplay, $related_videos, $loop ); ?>
 	<?php endif; ?>
 </div>
 <?php do_action( 'siteorigin_widgets_sow-video_after_video', $instance ); ?>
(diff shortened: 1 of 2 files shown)


Misplon:
@AlexGStapleton Under which conditions can I get a YT video to autoplay and loop?

AlexGStapleton:
@Misplon Tick Autoplay and Loop. As mentioned over Slack, certain browser features will prevent the autoplay. We can enforce this by setting the video muted by default - do you want to do that?

Misplon:
Without mute being added if I'm not seeing autoplay in Chrome macOS, where is it happening?

AlexGStapleton:
It's autoplaying without issue for me in Chrome and Firefox.

Are you okay with Mute being added as an optional setting? The description could be:
> Depending upon browser permissions, the video may not autoplay unless the video is also muted. For example, all mobile browsers require the video to be muted.

This makes it clear that if the user wants it to always autoplay they'll need to mute it.

Misplon:
@AlexGStapleton Thanks, mate, I'll get back to you regarding the above.

I've taken this branch, merged develop, and created a build. Autoplay is working for me here https://demo.siteorigin.com/corp/video-test/ but looping isn't. Is looping working for you on that test page?

AlexGStapleton:
Oh! The Video Widget template changes weren't merged, sorry about that. 🤦 
Please try again.

Misplon (approved):

gregpriday (approved):

PR #1305, opened 2021-05-19
Improve widget block saving

AlexGStapleton:
This PR brings the SiteOrigin Widgets Block in line with the SiteOrigin Layout Block. Previously, the data of the block would be saved in an inconsistent (to the user) manner, now it will happen after every change (ie. `.change()`). This will allow for more reliable `widgetHtml` generation.

This PR also implements the same update presentation method used by the SIteOrgin Layouts Block to prevent saving before `widgetHtml` is still being generated.

To test this PR, add a block and make some changes. Save the page and then open the page in a different tab. Make a change, save, and refresh the page in the other tab. Has the page updated correctly? Prior to this PR there's a chance that wouldn't be the case.

### DIFF
--- a/compat/block-editor/widget-block.js
+++ b/compat/block-editor/widget-block.js
@@ -94,7 +94,7 @@
 				wp.data.dispatch( 'core/editor' ).lockPostSaving();
 				jQuery.post( {
 					url: sowbBlockEditorAdmin.restUrl + 'sowb/v1/widgets/previews',
-					beforeSend: function ( xhr ) {
+					beforeSend: function( xhr ) {
 						xhr.setRequestHeader( 'X-WP-Nonce', sowbBlockEditorAdmin.nonce );
 					},
 					data: {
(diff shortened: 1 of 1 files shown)


compat/block-editor/widget-block.js
```diff
@@ -79,6 +90,33 @@
 				}
 			}
 
+			function generateWidgetPreview( widgetData = false) {
+				wp.data.dispatch( 'core/editor' ).lockPostSaving();
+				jQuery.post( {
+					url: sowbBlockEditorAdmin.restUrl + 'sowb/v1/widgets/previews',
+					beforeSend: function ( xhr ) {
```
Misplon, line 97: Should this perhaps be `function(` ?

Misplon (approved):
Just a small spacing change requested otherwise this is a nice usability improvement.

Misplon (approved):

gregpriday (approved):

PR #1310, opened 2021-05-20
Contact: Allow whitelisted HTML in Checkboxes options

AlexGStapleton:
This PR will allow users to add HTML to checkboxes options. A use case for this is for GDPR type validations ("here's a link to our privacy policy").

### DIFF
--- a/widgets/contact/fields/checkboxes.class.php
+++ b/widgets/contact/fields/checkboxes.class.php
@@ -13,7 +13,7 @@ class SiteOrigin_Widget_ContactForm_Field_Checkboxes extends SiteOrigin_Widget_C
 					<li>
 						<label>
 							<input type="checkbox" value="<?php echo esc_attr( $option['value'] ) ?>" name="<?php echo esc_attr( $options['field_name'] ) ?>[]" id="<?php echo esc_attr( $options['field_id'] ) . '-' . $i ?>"<?php echo checked( in_array( $option['value'], $options['value'] ), true, false ) ?>/>
-							<?php echo esc_html( $option['value'] ); ?>
+							<?php echo wp_kses_post( $option['value'] ); ?>
 						</label>
 					</li>
 				<?php endforeach; ?>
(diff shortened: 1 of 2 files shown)


Misplon:
Spaces surrounding an `a` tag appear to be stripped, is that happening for you?

AlexGStapleton:
Ah. That's happening due to applying flex to the checkbox label. Do you have any issues with removing the `display: flex` from the checkbox label and removing the `float` from the checkbox?

Misplon:
That doesn't result in the same alignment as before for me.

How about something like this. The following is for testing:

```
.sow-form-field-checkboxes label *:not(input) {
	display: contents;
}
```

AlexGStapleton:
There are accessibility concerns with `display: contents` that have only _just_ been addressed in Chrome and haven't been addressed in Safari.

https://hiddedevries.nl/en/blog/2018-04-21-more-accessible-markup-with-display-contents

What are your thoughts on moving the input outside of the label, and connecting it to the label using a `for` attribute. it's not as simple, but it'll avoid this without potentially introducing other issues and it should look the same.

Misplon:
Let's do that. We can apply flex to the list item.

AlexGStapleton:
How does that look?

Misplon:
@AlexGStapleton Can you confirm this is working for you? It isn't for me. Here is a live demo of this branch https://demo.siteorigin.com/corp/contact/.

AlexGStapleton:
The widget on that page had the outdated widget LESS - you can tell by inspecting the checkbox label as it no longer has `display: flex;` applied and it doesn't have the same selector as the radio field.



I've made a design setting change to the contact form on that page (I changed the error notice background) and saved the page and it's now working as expected. This won't be required for normal users as widget CSS is regenerated after a plugin update.

Misplon:
Thanks. I did do a plugin update though. It was a manual update but I bumped the version number.

AlexGStapleton:
That's not the same as an automatic update. Having to manually update the widget to trigger a widget CSS refresh, or manually clearing the `siteorigin-widgets` CSS cache is expected during development, and not required during a proper release.

Misplon:
Got you.

Misplon (approved):

gregpriday (approved):

PR #1311, opened 2021-05-20
Fix #1309 : Default icon in field Icon with custom font families (wit…

### LINKED ISSUE #1309: Default icon in field Icon with custom font families (without that icon) prevent init
I've installed the premium widgets and accordeon have default value for headings.

My setup have the font awesome pro installed, and all the other remove.

That make the field not init for the close and open icon because the accordeon have default icon that is in the fontawesome normal.

To fix this I add that check in the icon field (Line 26, in `render_field` function)
```php
if ( ! empty( $value ) && array_key_exists( $value_parts['family'], $icon_families_styles ) ) {
```
I can make a PR .

mamarmite:
…hout that icon) prevent init.

To replicate : 
```php
//remove fontawesome from the font families list.
//use the icon field in a form with a default font-awesome icon set on that field.
function set_current_project_fonts_siteorigin( $families ) {
    // base icons families are set in so-widgets-bundle\icons\icons.php
    $families = array();//clear the base icon, since we need user ot use only the one that are loaded.
    $families["fontawesomepro"] = Icons\FontAwesomePro::get_family_data();
    return $families;
}
add_filter('siteorigin_widgets_icon_families', __NAMESPACE__.'\\set_current_project_fonts_siteorigin', 11);
```

If my logic is right, the bug would occur is fontawesome isn't there for the default icon. So I would suggest just to remove fontawesome, or just keep one font that isn't fontawesome.

Let me know, if there is anything I can do :)

### DIFF
--- a/base/inc/fields/icon.class.php
+++ b/base/inc/fields/icon.class.php
@@ -23,7 +23,7 @@ class SiteOrigin_Widget_Field_Icon extends SiteOrigin_Widget_Field_Base {
 
 		$value_parts = self::get_value_parts( $value, $icon_families_styles );
 
-		if ( ! empty( $value ) ) {
+		if ( ! empty( $value ) && array_key_exists( $value_parts['family'], $icon_families_styles ) ) {
 			$value_family = $value_parts['family'];
 			$value_style = empty( $value_parts['style'] ) ? '' : ( '-' . $value_parts['style'] );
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
I've run a few tests and this will work but it won't clear the default value so the icon will still be stored regardless of whether it's valid. I've submitted https://github.com/siteorigin/so-widgets-bundle/pull/1313 which will clear the default.

Issue #1266, opened 2021-04-06
Image Grid: Only Remove Padding From Items As Needed


AlexGStapleton:
[This change was suggested here](https://github.com/siteorigin/so-widgets-bundle/pull/1262#issuecomment-814122907).

We're currently removing the top padding for every item rather than just the items that need it to be removed. We're not removing the bottom padding on the last line. We should only remove the top padding on the first line and remove the bottom padding on the last line.

AlexGStapleton:
@Misplon There isn't a simple method of doing this as flexbox doesn't provide any targeting options for just the first/last line (can't use `::first-line` for example) and `nth-child` is out of the question due to the wrap being automated. Here are a couple of options:

1. Do this using JavaScript (will cause a CLS so it's less than ideal)
2. Vertically split the item padding between the top and bottom (instead of `0` top padding and `10px` bottom padding, it'll be `5px` for both the top and bottom). This will result in items with a row before and after will have the expected amount of padding but rows without a row before or after will have half the amount of padding.
3. Apply equal spacing to all items, and attempt to remove it from the top/bottom of the container using negative margin.
4. Apply equal spacing to all items and don't worry about removing the spacing.

Which route would you like to take?

Misplon:
Thanks, mate. Let's perhaps try:

Apply equal spacing to all items, and attempt to remove it from the top/bottom of the container using negative margin.

AlexGStapleton:
I missed [this line](https://github.com/siteorigin/so-widgets-bundle/blob/develop/widgets/image-grid/styles/default.less#L8) when I initially wrote the above. With that in mind, the way to do this would be to do a hybrid of 2 and what's currently being used. I've submitted that here: https://github.com/siteorigin/so-widgets-bundle/pull/1314

Misplon:
Resolved via https://github.com/siteorigin/so-widgets-bundle/pull/1315.

PR #1314, opened 2021-05-25
Image Grid: Apply Vertical Padding evenly

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1266

This PR will ensure padding is evenly vertically split between the items and ensure the offset outlined in https://github.com/siteorigin/so-widgets-bundle/issues/1266 doesn't happen. To test this set a few different spacing values, and add enough image grid items to result in at least four rows.

Please note that during testing you'll need to clear the contents of `wp-content/plugins/siteorigin-widgets` to see a difference.

### DIFF
--- a/widgets/image-grid/styles/default.less
+++ b/widgets/image-grid/styles/default.less
@@ -5,12 +5,12 @@
 	flex-wrap: wrap;
 	justify-content: center;
 	align-items: baseline;
-	padding-top: @spacing;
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
I'm closing this PR in favour of a rework of the spacing setting which will be handled in a different PR.

PR #1323, opened 2021-06-01
Update Instance size to height

KevinTaron:
Hey that should be $instance['size_height'] right?
If not, deleted my request.


AlexGStapleton (approved):
Nice catch. Thanks mate! :)

PR #1329, opened 2021-06-07
Multiple media: Improved Image Support & `image_dimensions` parameter 

AlexGStapleton:
This PR allows for non-Image to be seen in the widget after saving the page (previously you could add them, but couldn't select/view them when editing the widget after loading), and adds the `image_dimensions` parameter to allow for developers to set specific dimensions.

[Test widget](https://drive.google.com/uc?id=14vytZ86-OK8xNPvzE6MRUD8I9JlXrwlp)

This PR also reduces the thumbnail dimensions from 75x75 to 64x64. This is to reduce stretching with WP Media Type Icons.

### DIFF
--- a/base/inc/fields/multiple-media.class.php
+++ b/base/inc/fields/multiple-media.class.php
@@ -62,7 +62,7 @@ class SiteOrigin_Widget_Field_Multiple_Media extends SiteOrigin_Widget_Field_Bas
 				<?php
 				if ( is_array( $attachments ) ) {
 					foreach ( $attachments as $attachment ) {
-						$post = get_post( $attachment );
+						$title = get_the_title( $attachment );
 						$src = wp_get_attachment_image_src( $attachment, 'thumbnail' );
 
 						if ( empty( $src ) ) {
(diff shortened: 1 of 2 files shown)


AlexGStapleton:
@Misplon What are your thoughts on adding a setting to allow for the title to be visible above/below the thumbnail? While it's possible to see the thumbnail by hovering the thumbnail, it's not possible to tell at a glance what an item is unless it's an image.

Misplon:
Sounds good. Perhaps default enabled and then possible to disable with a filter? What were you thinking?

Misplon (approved):

base/inc/fields/multiple-media.class.php
```diff
@@ -45,6 +57,17 @@ protected function render_field( $attachments, $instance ) {
...
 
+		// Ensure thumbnail_dimensions are valid. 
+		if (
+			empty( $this->thumbnail_dimensions ) ||
+			empty( $this->thumbnail_dimensions[0] ) ||
+			empty( $this->thumbnail_dimensions[1] ) ||
+			! is_numeric( $this->thumbnail_dimensions[0] ) ||
+			! is_numeric( $this->thumbnail_dimensions[1] )
+		) {
+			$this->thumbnail_dimensions = array( 64, 64 );
```
gregpriday, line 68: Maybe having this `array( 64, 64 )` as a class constant or static class variable would be better? It's used in multiple places.

gregpriday (approved):
One suggestion, but this is good to go. It can be merged if @AlexGStapleton is happy.

Issue #1343, opened 2021-06-20
Headline: Add a Mobile Alignment setting


Misplon:
@AlexGStapleton If you're in agreement, let's add a Mobile Alignment setting below each of the Alignment settings in the Headline widget.

PR #1344, opened 2021-06-20
Headline: Mobile Alignment Settings

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1343

This PR adds mobile alignment settings for the headline, sub-headline, and divider. It also adds a global widget breakpoint setting that allows users to control the mobile alignment CSS is used.

Pre-existing widgets will have Mobile alignment set to the same value as the related Alignment setting. Please test to ensure the migration works as expected by setting up a few headlines prior to switching to the PR branch.

### DIFF
--- a/widgets/headline/headline.php
+++ b/widgets/headline/headline.php
@@ -37,7 +37,7 @@ class SiteOrigin_Widget_Headline_Widget extends SiteOrigin_Widget {
 				'type' => 'measurement',
 				'label' => __( 'Responsive Breakpoint', 'so-widgets-bundle' ),
 				'default' => '780px',
-				'description' => __( 'This setting controls when the Mobile alignment settings will be used. The default value is 780px.', 'so-widgets-bundle' ),
+				'description' => __( 'The pixel resolution when the mobile alignment settings will be applied.', 'so-widgets-bundle' ),
 			),
 		);
 	}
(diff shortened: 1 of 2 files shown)


widgets/headline/headline.php
```diff
@@ -31,6 +31,17 @@ function initialize(){
 		add_filter( 'siteorigin_widgets_wrapper_data_' . $this->id_base, array( $this, 'wrapper_data_filter' ), 10, 2 );
 	}
 
+	function get_settings_form() {
+		return array(
+			'responsive_breakpoint' => array(
+				'type' => 'measurement',
+				'label' => __( 'Responsive Breakpoint', 'so-widgets-bundle' ),
+				'default' => '780px',
+				'description' => __( 'This setting controls when the Mobile alignment settings will be used. The default value is 780px.', 'so-widgets-bundle' ),
```
Misplon, line 40: I know this is the standard description for this setting but it isn't the best. Let's perhaps change this to:

`The pixel resolution when the mobile alignment settings will be used.`

Misplon:
**Headline:**
Try:

Alignment: center. 
Mobile alignment: right. 

Does that produce the expected result for you?

Misplon:
Testing error on my end, please, ignore the above. Thanks.

Misplon (changes requested):
Looks good, just a minor global setting description change requested.

widgets/headline/headline.php
AlexGStapleton: Sounds good. Do you want to retain the "The default value is 780px" text?

Misplon: Super, thanks. I'd like to remove the `The default value is 780px.` text.

Misplon (approved):

gregpriday (approved):

PR #1356, opened 2021-06-27
Preset Form Field Adjustments

AlexGStapleton:
This commit includes three preset specific changes:

**Preset Field: Store Selected Preset**
This commit allows for the selected preset to be saved.

**Preset: Add `default_preset` to allow default Preset** 
This commit allows the preset field to support a default preset. This change was complicated due to how state emitters work.
The `default_preset ` parameter was used over `default` due to how `default` being a valid form field option and it not allowing for reliable default detection in the preset JS.

**SiteOrigin Widget: dynamic_preset_state_handler** 

This commit allows a method of dynamically setting the state_handler of an array of `form_fields` based on the values used by the preset field. This basically means, we can easily show/hide fields based on the preset values. This is going to be used heavily by some of our upcoming widgets. 

---

You can test the above changes by using this [test plugin](https://siteorigin.com/wp-content/uploads/2021/06/siteorigin-preset-field-demo.zip). Once installed, navigate to Plugins > SiteOrigin Widgets and activate the `SiteOrigin Preset Field` widget. Add it to any page and you'll notice:

- That the default_preset "test-2" is loaded by default.
- Upon changing preset, certain fields will change and even display altogether.
- Upon saving, the selected preset is saved. 

Documentation: https://github.com/siteorigin/docs/pull/120.

### DIFF
--- a/base/siteorigin-widget.class.php
+++ b/base/siteorigin-widget.class.php
@@ -1350,4 +1350,35 @@ abstract class SiteOrigin_Widget extends WP_Widget {
 
 		return $values;
 	}
+
+	// Add state_handler for fields based on how they're adjusted by preset data.
+	protected function dynamic_preset_state_handler( $state_name, $preset_data, $fields ) {
+		$adjusted_fields = array();
+		// Build an array of all the adjusted fields by the preset data, and note which presets adjust them.
+		foreach ( $preset_data as $preset_id => $preset ) {
+			$extracted_fields = array_keys( $preset['values'] );
+			foreach ( $extracted_fields as $field ) {
+				$adjusted_fields[ $field ][] = $preset_id;
+			}
+		}
+
+		// Add state handlers for any of the $adjusted_fields. 
+		foreach ( $adjusted_fields as $field => $used_by ) {
+			// Skip field if it's not adjusted by of the presets, or if the field has a state_handler already.
+			if (
+				! isset( $fields[ $field ] ) ||
+				isset( $fields[ $field ]['state_handler'] )
+			) {
+				continue;
+			}
+
+			$used_by = implode( ',', $used_by ); 
+			$fields[ $field ]['state_handler'] = array(
+				$state_name . '[' . $used_by . ']' => array( 'show' ),
+				'_else[' . $state_name . ']' => array( 'hide' ),
+			);
+		}
+
+		return $fields;
+	}
 }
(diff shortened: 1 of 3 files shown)


AlexGStapleton:
@Misplon Where do you want to put the "SiteOrigin Widget: dynamic_preset_state_handler" documentation? Adding it to the Form Field, or State Emitters page is a bit much as it's quite specific. Maybe a new page under the Advanced Concepts?

Misplon:
@AlexGStapleton Thanks, mate. Sure, let's go with Advanced Concepts. Is it a concept though? Advanced Functionality?

AlexGStapleton:
I don't think we should introduce a new section called Advanced Functionality, but I would be okay with renaming Advanced Concepts to Advanced Functionality.

Misplon:
Ahh my bad, I misread that. Got you. Please, proceed as discussed.

Misplon:
@AlexGStapleton Thanks, mate. The only item that isn't working for me is:

* That the default_preset "test-2" is loaded by default.

AlexGStapleton:
It appears there was an issue with the initial PR test plugin. Please try [the preset docs version of that plugin](https://siteorigin.com/wp-content/uploads/2021/06/siteorigin-preset-field-demo.zip) instead. I've tested and confirmed that works for me. I've updated the installation instructions in the initial issue.

Misplon:
Awesome. Super cool.

Misplon (approved):

AlexGStapleton:
The last commit is a pretty significant rework to allow for better handling of section fields.

To test it please use the test plugin to test existing functionality and https://github.com/siteorigin/so-widgets-bundle/pull/1359/commits/044089a0a2fd94accd38d12eeda1dae1154e3a6f to test section support - the column count field in the settings section will change and disappear based on the selected template.

No documentation changes are required.

Misplon (approved):

Misplon:
Awesome, nicely done mate.

base/siteorigin-widget.class.php
```diff
@@ -1350,4 +1350,71 @@ function save_global_settings( $values ){
 
 		return $values;
 	}
+
+	// Add state_handler for fields based on how they're adjusted by preset data.
```
gregpriday, line 1354: For these comments, can we go with a PHPDoc format rather.

gregpriday (changes requested):
Just a change to comments.

gregpriday (approved):

Issue #1066, opened 2020-06-09
New Widget: SiteOrigin Blog Widget


Misplon:
## Template: 

- [Standard](https://demo.siteorigin.com/corp/standard-blog-layout/)
- [Grid](https://demo.siteorigin.com/corp/grid-blog-layout/)
- [Offset](https://demo.siteorigin.com/corp/offset-blog-layout/)
- [Alternate](https://demo.siteorigin.com/corp/alternate-blog-layout/)
- [Masonry](https://demo.siteorigin.com/corp/masonry-blog-layout/)
- [Portfolio](https://demo.siteorigin.com/corp/portfolio/)

## Settings:

- Column count
- Featured image (Checkbox)
- Post content (Select: Post Excerpt | Full Post Content): Choose how to display your post content. Select Full Post Content if using the "more" quicktag.
- Excerpt length (Number): If a manual post excerpt isn't added, one will be generated. Choose how many words it should be.
- Post excerpt read more link (Checkbox): Display the Read More link below the post excerpt.
- Post date (Checkbox)
- Post author (Checkbox)
- Post categories (Checkbox)
- Post tags (Checkbox)
- Post comment count (Checkbox)

PR #1359, opened 2021-07-04
Blog Widget

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1066
Requires https://github.com/siteorigin/so-widgets-bundle/pull/1356


AlexGStapleton:
Defaults haven't been set so if you have any preferences let me know. The Column Count setting doesn't do anything at the moment as the Standard template would presumably not use that setting.

AlexGStapleton:
@Misplon Oh whoops, didn't tag you.

AlexGStapleton:
@Misplon Regarding the pagination, do you want to use this?

https://github.com/siteorigin/siteorigin-panels/pull/781

This is currently "locked" to the post loop widget but I can foresee some users wanting to have multiple blog widgets on the same page. If we use this, I'll need to duplicate the code (with modifications) in WB directly to allow for it to work without PB.

Should we add a general Title field for the widget? This would be in line with most widgets.

Misplon:
> Defaults haven't been set so if you have any preferences let me know. The Column Count setting doesn't do anything at the moment as the Standard template would presumably not use that setting.

Thanks. The column count would be one for the standard layout.

> @Misplon Regarding the pagination, do you want to use this?

siteorigin/siteorigin-panels#781

This is currently "locked" to the post loop widget but I can foresee some users wanting to have multiple blog widgets on the same page. If we use this, I'll need to duplicate the code (with modifications) in WB directly to allow for it to work without PB.

We'll almost certainly run into challenges with multiple widgets on a page so yeah, let's do it.

> Should we add a general Title field for the widget? This would be in line with most widgets.

Sure, sounds fine.

AlexGStapleton:
How do you want to handle a situation where we don't find any posts? We could output nothing, but we could handle it like themes typically do and show an optional message set by the user.

Misplon:
I'm happy to copy the default WordPress message for that. I don't think it needs to be user editable at this stage. If we get requests we can take another look.

Misplon:
@AlexGStapleton Let's perhaps include this:
* Support of ‘Co-Authors Plus’ plugin to showcase multiple post authors.
https://docs.wpvip.com/technical-references/plugins/incorporate-co-authors-plus-template-tags-into-your-theme/

AlexGStapleton:
> I'm happy to copy the default WordPress message for that. I don't think it needs to be user editable at this stage. If we get requests we can take another look.

By default, WordPress doesn't output _anything_ when no posts are found. That's currently how the widget works. Most themes however will output a message instead when no posts are found. We could base the message on one of our how we handle it in Corp but the default isn't the greatest:

https://github.com/siteorigin/siteorigin-corp/blob/develop/template-parts/content-none.php

AlexGStapleton:
@Misplon How 1:1 are we going to keep the styling of the templates with Corp? Just a general guideline for the moment, or as close as possible?

Most of the templates collapse in Corp at 480px. Should we use a higher collapse point for the Blog widget? Typically, we use 780px but it's not like we're locked to that.

Misplon:
780px is fine.

Templates are going to take some work to figure out how much styling to add to override theme styling but yeah, let's keep it as close as possible I reckon.

AlexGStapleton:
@Misplon Currently all of the template colors/background are stored in style/TEMPLATENAME.less. They're separated as they could very easily be made settings.

How do you want to handle featured image size? We could introduce template specific image sizes, or we could introduce an Image Size setting.

AlexGStapleton:
I've finished adding the rest of the templates. I'm going to hit these shortly:

- Support of ‘Co-Authors Plus’ plugin 
- Implementing a version of siteorigin/siteorigin-panels#781

The portfolio template will attempt to display using the Jetpack Portfolio tags. If it can't, it fallback to `category`. The user can change the "fallback" term by using the `siteorigin_widgets_blog_portfolio_fallback_term` filter.

Misplon:
@AlexGStapleton Are we able to display the default for the Post Content dropdown?

Edit_Page_‹_SiteOrigin_—_WordPress_🔊

AlexGStapleton:
@Misplon https://github.com/siteorigin/so-widgets-bundle/pull/1359/commits/364f013349e2c9486a992ecb3adca6a9d8fbfb5f implements https://github.com/siteorigin/siteorigin-panels/pull/781.

AlexGStapleton:
@Misplon 
> Are we able to display the default for the Post Content dropdown?

I'm having trouble replicating this. [Is the default_preset](https://github.com/siteorigin/so-widgets-bundle/blob/develop/base/inc/fields/presets.class.php#L43-L45) being set correctly for you?

Misplon:
Edit_Page_‹_Puro_—_WordPress

Above is what I'm seeing mate.

Misplon:
Ahh my bad, pulled the most recent changes, issue resolved.

AlexGStapleton:
@Misplon No further structural changes are expected so I've marked this as Ready for Review.

AlexGStapleton:
> Let's perhaps include this:
> 
>  Support of ‘Co-Authors Plus’ plugin to showcase multiple post authors.
>  https://docs.wpvip.com/technical-references/plugins/incorporate-co-authors-plus-template-tags-into-your-theme/

Oh! I forgot about this. I'll do this shortly.

AlexGStapleton:
@Misplon Last commit adds [Co-Authors Plus](https://wordpress.org/plugins/co-authors-plus/) Integration.

Misplon:
@AlexGStapleton Do you want to add the Corp arrows for pagination and read more?

arrows

I'm keen if you are.

Misplon:
The following looks to be printing with all meta settings disabled:

```
Warning: Undefined array key "categories" in /Users/amisplon/Sites/siteorigin/wp-content/plugins/so-widgets-bundle/widgets/blog/blog.php on line 169

Warning: Undefined array key "author" in /Users/amisplon/Sites/siteorigin/wp-content/plugins/so-widgets-bundle/widgets/blog/blog.php on line 170
```

Misplon:
**CLS**
If possible, let's try to address the CLS issue caused by our blog styles loading after the theme styles. Thanks, mate.

Misplon:
@AlexGStapleton What do you think about the following new settings for the Standard template?

**Standard**

- Post Border Color
- Post Background Color

- Post Title Color
- Post Title Hover Color
- Post Meta Color
- Post Meta Hover Color

- Post Content Color
- Post Link color
- Post Link Hover Color

- Pagination Link Color
- Pagination Link Hover Color

Misplon:
### Grid Template

### Design

- Post Border Color
- Post Background Color

- Post Title Color
- Post Title Hover Color
- Post Title Font Size
- Post Meta Color
- Post Meta Hover Color
- Post Meta Font Size

- Post Content Color
- Post Link color
- Post Link Hover Color
- Post Content Font Size

- Pagination Link Color
- Pagination Link Hover Color
- Pagination Font Size

----

* Should we put the pagination styling outside of the template-specific design settings?
* Let's perhaps make the excerpt read more setting conditional on the excerpt content setting.

* In Corp we're not equalizing the column height when full content is selected. I think we should do that here. If you agree let's put a class name at the `.sow-blog` level when full content is selected.

[Corp Reference](https://github.com/siteorigin/siteorigin-corp/blob/develop/sass/site/primary/_posts-and-pages.scss#L239-L242).

Misplon:
## Offset Template

### Design

- Post Border Color
- Post Background Color

[]()

- Post Title Color
- Post Title Hover Color
- Post Title Font Size
- Post Meta Color
- Post Meta Hover Color
- Post Meta Font Size
- Offset Post Meta Color
- Offset Post Meta Link Color
- Offset Post Meta Hover Color
- Offset Post Meta Font Size
- Offset Post Meta Link Font Size

[]()

- Post Content Color
- Post Link Color
- Post Link Hover Color
- Post Content Font Size

[]()

- Pagination Link Color
- Pagination Link Hover Color
- Pagination Font Size

[]()

----

Misplon:
## Alternate Template

### Design

- Post Border Color
- Post Background Color

[]()

- Post Title Color
- Post Title Hover Color
- Post Title Font Size
- Post Meta Color
- Post Meta Hover Color
- Post Meta Font Size

[]()

- Post Content Color
- Post Link Color
- Post Link Hover Color
- Post Content Font Size

[]()

- Pagination Link Color
- Pagination Link Hover Color
- Pagination Font Size

[]()

----

Post Content
I think we should default Post Content to Excerpt.

Misplon:
# Portfolio

## Design

### Filter Categories

- Enable
- Color
- Hover Color
- Font Size
- Text Transform
- Current Border Color
- Current Border Thickness

### Featured Image

- Border Color
- Hover Background
- Hover Background Opacity
- Post Title Color
- Post Title Font Size
- Divider Border Color
- Divider Border Thickness
- Divider Border Margin
- Post Meta Color
- Post Meta Font Size

### Pagination

- Top Margin
- Link Margin
- Border Color
- Background
- Hover Background
- Border Radius
- Text Color
- Hover Text Color

[]()

----

**Featured Image Sizing and Spacing*
Can we try get the featured image sizing and spacing the same as the original design?

AlexGStapleton:
I've had a look over the settings and the only differences I can see are:

- The Offset theme has Offset Post Meta
- The Overlay theme has Filter Categories and Featured Image settings.

For everything else, all templates have access to all other settings. There were a few instances of settings being reordered. Is that intentional?

> * Let's perhaps make the excerpt read more setting conditional on the excerpt content setting.

Is this for all templates, or just the Standard and Grid templates? If it's just for those two settings I may need to introduce a different read more setting for the other templates.

Misplon:
Thanks, mate.

# Standard

If I roll back to `642569c2 Blog: Use Image Size for Portfolio Images` the Standard post meta looks as follows:

intended

Using the latest commit in this branch, the post meta looks as follows:

standard-meta

Misplon:
@AlexGStapleton My comments above have been edited. Thanks again.

AlexGStapleton:
>  Corp has specificity over the Design > Post settings. The settings print but don't work due to specificity.

I can't seem to replicate this. To confirm, you mean the Post settings right? Is it both settings, or just one of them?

Everything else mentioned has been resolved - a build will be required for the preset change.

Misplon:
> Corp has specificity over the Design > Post settings. The settings print but don't work due to specificity.

I'm unsure why but I can't replicate the issue anymore. Thanks for the changes. I'll get back to you once I've done further testing.

Misplon:
I've uploaded a build to the demo server. 

https://demo.siteorigin.com/corp/sandbox/

- It looks like an issue if we don't set the border color.
- I see your meta changes but I'm not seeing them take effect.
- I'm battling with performance when saving the widget.

Misplon:
Locally, I'm seeing the following settings overridden:

Design > Post > Border Color 
Design > Post > Background Color

Page_Builder_–_SiteOrigin

Misplon:
@AlexGStapleton I'll wait for you, mate. I've tried a build without merging develop and the pagination changes aren't showing for me so it looks like I'm missing something. Pagination background is targeting the dots for example.

Misplon:
Thanks, mate. I've updated the string. Please, try Polestar as your theme. Does the Standard template load as follows for you?

Puro_–_Just_another_WordPress_site

AlexGStapleton:
> Post Border Color:

I'm not able to replicate this result, but I've changed it regardless.

Added the defaults, and corrected meta colors.

Misplon:
Thanks, mate.

Here is what I'm seeing for pagination:

pagination

Can you perhaps look at getting the original styling back?

so-widgets-bundle

Changes are also required for the dots selector.

Misplon:
- Read more link.
If possible, apply the Post Content Font Size setting to the read more link.

AlexGStapleton:
>  If possible, apply the Post Content Font Size setting to the read more link.
This was already the case but it appears it can be overridden due to theme styling. I've applied font styling specifically to the more link.

Misplon:
Thanks, mate.

For the pagination arrows, let's try:

```
padding-left: 14px;
padding-right: 14px;
```

If I clear the Post they don't return to the default, they are cleared. Is that expected? Post > Border Color and Post > Background Color.

AlexGStapleton:
>  If I clear the Post they don't return to the default, they are cleared. Is that expected? Post > Border Color and Post > Background Color.

Expected, yes. Unfortunate, yes. The Presets Field doesn't update the default, it only sets the value. Updating the default, while very complicated, should be possible. Another option would be to disable the Default button.

> Should we perhaps remove the padding: 11px 17px; and instead offer a height/width setting for the pagination buttons? 

I like this idea. Is there a specific reason you want to go with width/height over just padding?

Misplon:
@AlexGStapleton Could we update a few defaults, if possible:

Standard:

Post Meta
Color: #929292
Hover Color: #f14e4e

Post Content
Color: #626262
Link Hover Color: #626262

Pagination
Border Color: #626262
Link Color: #626262

Apologies for the changes. I'll get through the rest of the templates this week. Tempted to launch it with just Standard, but yeah, nearly there.

Misplon:
@AlexGStapleton Are the pagination width and height settings available when you test? Perhaps you can send me a build to confirm?

I'm getting unexpected results testing at the moment. For example, the below is what the Offset presets produce:

offset

AlexGStapleton:
> Are the settings presets loading for you when changing templates if settings have been edited?

Fixed.

> Add icon-long-arrow-right for Continue reading the same as Corp.

That's using an icon font, which is pretty heavy for one icon (it's easy to justify for Corp as there are several icons present). Are you okay with using the pagination arrow used by Corp instead?

Screenshot 2022-07-18 at 13-51-49 Standard Blog Layout - Corp

> Is styling working in the Live Editor for you?

Yes. It's possible this was unintentionally fixed by the merge branch (which will ensure this PR is able to be tested on the demo site without fear of breaking unrelated things like Anchor ids). If it's still not working for you, can you please outline a specific setting you're checking? For reference, I tested this using **Design > Post** and **Design > Pagination**.

AlexGStapleton:
I'm having trouble replicating this freeze. I remember you mentioning it previously and I was only able to replicate it while using a Mac. I'll gain access to a Mac and report back shortly.

These are ready for testing:
> Add pagination for Continue reading the same as Corp.
> Post Title CLS: Most noticeable with the post titles. Let's perhaps check out what can be done for post titles when able.
> CLS is also quite noticeable with the Portfolio template because of the button tags used to filter categories.

AlexGStapleton:
> The freezing issue came back right after you fixed the template settings issue.

I figured that was the case. Thanks for confirming.

Test results:
- Unable to replicate on 2020 Mac Mini.
- Able to replicate on Macbook Air late 2020.

I suspect it's likely performance related rather than being a Mac specific issue.

AlexGStapleton:
The last commit should help. It'll cut down on the number of fields being processed from 100ish to 18 without losing any functionality for our usage. Let me know how that looks?

AlexGStapleton:
I'm not, no. I suspect that error is related to that display error you highlighted over Slack. Can you send me a copy of the layout you're using? That should be all that's required for me to replicate it and that'll allow me to identify what specifically is causing that.

AlexGStapleton:
I've been able to replicate this issue. This is a PHP 8 specific error and it's caused by adding a [hyphen prior to a less variable](https://github.com/siteorigin/so-widgets-bundle/commit/015c05c069b70c057882d8a98596872e9b52241b#diff-d1c4407e47dbe534e14059c9429873189a544cc06068cedc7a381b05c6d7bb0eR217).  Resolved by [the last commit.](https://github.com/siteorigin/so-widgets-bundle/pull/1359/commits/b92b399be1241760c1b039c7f63bce40e4e9629f)

AlexGStapleton:
Resolved by https://github.com/siteorigin/so-widgets-bundle/pull/1359/commits/27e6d05c2dc0ba107d30d8f3bfd4becdf342ea82 (that should also help with performance in certain situations).

Misplon:
@AlexGStapleton Well done mate.

Misplon:
@AlexGStapleton, I'm getting there. I'll keep going with minor tweaks tomorrow. Can you perhaps add support for the read more link so we can format like the excerpt more link?

Changing to the Portfolio template is the only triggering a short period grey screen for me. Not sure if anything can be done there.

AlexGStapleton:
> Changing to the Portfolio template is the only triggering a short period grey screen for me

Hm. Maybe try increasing [this value](https://github.com/siteorigin/so-widgets-bundle/pull/1359/files#diff-c391e1b0d6c9bdcb8bbee04fd91793c9feeb8549b2cb70b33e4aa9fc0ce1bbffR1365) to 40 to see if that helps.

I'm not able to replicate this (or any of the other template-related lag) anymore while using a Macbook Air so I assume what I was previously getting was unrelated to what you were getting. I suspect this issue will be reduced over time as WebKit becomes more optimised for m1.

AlexGStapleton:
> I'll keep going with minor tweaks tomorrow. Can you perhaps add support for the read more link so we can format like the excerpt more link?

Done. Requires the `Post Excerpt Read More Link` setting to be enabled. This was done like this to allow users to use the theme read more over ours and to add a workaround for compatibility issues related to us overriding it.

AlexGStapleton:
[siteorigin_widgets_blog_full_content_read_more](https://github.com/siteorigin/so-widgets-bundle/commit/aae612b5c2dc28b00215c7c21f47490dde9b67d5) can be tested by using the following PHP:

add_filter( 'siteorigin_widgets_blog_full_content_read_more', '__return_false' )

That'll prevent the read more from being handled by the Blog widget.

Misplon:
@AlexGStapleton 

For the Portfolio layout, let's add a new setting for Filter Categories and place it above Post Categories.
Portfolio Column Count doesn't seem to be working for me now.
The Portfolio Template struggles in the admin on my M1.

Misplon:
> Hm. Maybe try increasing [this value](https://github.com/siteorigin/so-widgets-bundle/pull/1359/files#diff-c391e1b0d6c9bdcb8bbee04fd91793c9feeb8549b2cb70b33e4aa9fc0ce1bbffR1365) to 40 to see if that helps.

Marginally better but not much.

AlexGStapleton:
> For the Portfolio layout, let's add a new setting for Filter Categories and place it above Post Categories.

Added.

> Marginally better but not much.

Thanks for trying. How is it when compared to the other layout? Just a rough estimate of how long it takes vs a different layout should be fine.

Misplon:
> For the Portfolio layout, let's add a new setting for Filter Categories and place it above Post Categories.

Not sure I'm seeing that on my end.

> Thanks for trying. How is it when compared to the other layout? Just a rough estimate of how long it takes vs a different layout should be fine.

It's unplayable. I'll send you a screencast.

Misplon (approved):

Issue #1358, opened 2021-06-30
Image Grid widget: Add Support for Custom Image Size 


AlexGStapleton:
Related https://github.com/siteorigin/so-widgets-bundle/pull/1244

We should introduce custom image sizes for the Image Grid widget.

PR #1371, opened 2021-07-19
Image Grid: Custom Size

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1358

This PR allows adds support for custom sizes.

The Custom Size attachment size setting does not affect external images. This is due to the Image Grid not support attachment sizes for external images (this could be worthwhile for a future update). The maximum width/height settings are hidden when the Attachment Size is set to Custom Size. This is to prevent potential sizing issues.


AlexGStapleton:
@Misplon You'll notice styling issues with the Image Size field while using this PR. That's resolved by https://github.com/siteorigin/so-widgets-bundle/pull/1372

Misplon (approved):

Issue #1365, opened 2021-07-13
Testimonials: Adjust Flexbox to column for mobile


Misplon:
At the assigned mobile resolution, adjust Flexbox to column. We'll need to make width and padding adjustments too.

As an example, here is what I sent to one of our premium users:

```
@media (max-width: 600px) {
	.sow-testimonials .sow-testimonial-wrapper.sow-layout-side .sow-testimonial {
		flex-direction: column;
	}

	.sow-testimonials .sow-testimonial-wrapper.sow-layout-side .sow-testimonial .sow-testimonial-user {
		padding: 0 0 15px !important;
		text-align: center;
		width: 100% !important;
	}

	.sow-testimonials .sow-testimonial-wrapper.sow-layout-side .sow-testimonial .sow-testimonial-name {
		font-size: 15px;
	}

	.sow-testimonials .sow-testimonial-wrapper.sow-layout-side .sow-testimonial .sow-testimonial-location {
		font-size: 15px;
	}

	.sow-testimonials .sow-testimonial-wrapper.sow-layout-side .sow-testimonial .sow-testimonial-text {
		font-size: 16px;
		margin-bottom: 35px
	}
}
```

PR #1379, opened 2021-07-21
Testimonials: Collapse to Column On Mobile

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1365


AlexGStapleton:
@Misplon This PR doesn't use the font sizes in the linked issue. If you want to keep them, I think we should switch to em/rem.

Issue #125, opened 2016-02-29
Add the ability to set relationship in image widget


AlexGStapleton:
[Issue reported in this thread](https://wordpress.org/support/topic/page-builder-doesnt-generate-for-responsive-lightbox-by-dfactory?replies=3#post-8092413).

rel is typically used to allow images to detected by lightboxes and the like. We should allow the user to input a relationship the same way they would specify a class.

Misplon:
Still relevant.

PR #1385, opened 2021-08-10
Image Widget: rel field

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/125

Wasn't sure of the placement so I put it towards the bottom.


AlexGStapleton:
Let me know if you would like to do the same for the Image Grid (and the desired placement).

Misplon (approved):

PR #1395, opened 2021-09-09
Button: Ensure $instance isn't empty before modifying it

AlexGStapleton:
[Reported here](https://siteorigin.com/thread/so-widgets-bundle-generating-an-error/)

Resolves the following error:

> Fatal error: Uncaught TypeError: Cannot access offset of type string on string in /home/qwerty/public_html/wp-content/plugins/so-widgets-bundle/widgets/button/button.php:433

This isn't simple to test as the error requires a very specific setup so just confirm the Button widget displays as expected.

### DIFF
--- a/widgets/button/button.php
+++ b/widgets/button/button.php
@@ -405,6 +405,10 @@ class SiteOrigin_Widget_Button_Widget extends SiteOrigin_Widget {
 	 * @return mixed
 	 */
 	function modify_instance( $instance ) {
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
The user has reported this has helped. Waiting for final confirmation before marking as Ready for Review.

AlexGStapleton:
The user reported success.

Misplon (approved):

PR #1425, opened 2021-11-12
Updated Premium CTA's

Misplon:


AlexGStapleton (approved):
Looks good.

We should consider moving the premium check directly to the `SiteOrigin_Widget` rather than each individual teaser as it seems odd to do it for each and every teaser - this isn't worth delaying change.

PR #1436, opened 2022-01-05
Post Carousel: Theme Related Changes

AlexGStapleton:
This PR:
- Introduces Asset Registering https://github.com/siteorigin/so-widgets-bundle/commit/4b4355f93bf122ee558b3e0787980e61b2977965
- Removes the Loop Posts setting in favour of the global setting (the migration process will need to be directly tested).
- Introduces the `siteorigin_widgets_post_carousel_settings_form` filter that allows for filtering of the carousel settings added to the widgets form.
- Introduces the `siteorigin_widgets_post_carousel_settings_frontend` filter which allows filtering of the carousel settings passed to the Carousel JS _before_ they're json encoded.
- Introduces a navigation layout that allows for inline dot and arrow navigation.


Misplon:
Let's add `box-shadow: none;` to `.sow-carousel-wrapper .slick-dots li button` to avoid a shadow added by some themes.

AlexGStapleton:
Done via https://github.com/siteorigin/siteorigin-premium/pull/587/commits/4e7da516d002fca4fd354318346d308a70c63afd as this PR doesn't directly add any dot CSS - it's added by the theme itself.

Misplon (approved):

PR #1438, opened 2022-01-05
Slider: Disable `loading` override when AMP plugin is active

AlexGStapleton:
This PR resolves an AMP compatibility issue with the `loading` attribute `eager` value. If the [AMP plugin](https://wordpress.org/plugins/amp/) is active and enabled, this PR will prevent the Slider widgets `loading` attribute override for images. 

`siteorigin_widgets_slider_attr` filter to do this. This was done in response to [this thread](https://wordpress.org/support/topic/disallowed-attribute-value/) and you can test this issue by checking the attribute of the SiteOrigin Slider Foreground and Background Image to ensure it's set to loading and `skip-lazy` isn't a set class for the image.


Misplon:
Thanks, mate. With only a SO Slider background image added I seem to getting:

`<div class="sow-slider-base" style="display: none;">`

Misplon:
Unsure why but I can't replicate the issue. Thanks.

Misplon (approved):

PR #1442, opened 2022-01-25
Widgets Block: Prevent Potential Undefined Error Outside of WP Admin

AlexGStapleton:
This PR will resolve the following Updraft Central error: I'm unsure when this error occurs as the logs don't provide further context but based on the context it's likely outside of WP Admin.

` The remote site has encountered a fatal error while executing the requested command. Remote response follows: PHP Fatal error (Error) has occurred during UpdraftCentral command execution. Error Message: Call to undefined function get_current_screen() (Code: 0, line 27 in /htdocs/wordpress/wp-content/plugins/so-widgets-bundle/compat/block-editor/widget-block.php) `

### DIFF
--- a/compat/block-editor/widget-block.php
+++ b/compat/block-editor/widget-block.php
@@ -24,13 +24,13 @@ class SiteOrigin_Widgets_Bundle_Widget_Block {
 	}
 
 	public function enqueue_widget_block_editor_assets() {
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
User reported success.

Misplon (approved):

PR #1443, opened 2022-01-27
Social Media Buttons: Hover Background and Icon Color

AlexGStapleton:
This PR will introduce `Icon color hover` and `Background color hover` settings. These settings are visible after enabling the `Use hover effects`. If these settings aren't set, the previous button styling will be used.


Misplon:
Thanks, mate. Are you ok with reversing the label word order?

Icon color hover
Background color hover

to:

Icon hover color
Background hover color

Misplon (approved):

Issue #1452, opened 2022-02-28
Post Carouse: Fix Default Thumbnail External URL field


Misplon:
Check the Default Thumbnail External URL field. Resolve functionality. Thanks!

PR #1453, opened 2022-03-02
Post Carousel: Add Support for Fallback External URL

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1452

Please test that post featured images, external images, and the external URL work as expected. The external URL field will only be used if the external image isn't set.

This PR also ensures that the assigned image size is used for the fallback image.

### DIFF
--- a/base/inc/widgets/tpl/carousel.php
+++ b/base/inc/widgets/tpl/carousel.php
@@ -20,6 +20,7 @@
 	?>
 	<div class="sow-carousel-wrapper"
 		 data-dir="<?php echo is_rtl() ? 'rtl' : 'ltr'; ?>"
+		 style="opacity: 0;"
 		 <?php
 		 foreach ( $settings['attributes'] as $n => $v ) {
 		 	if ( ! empty( $n ) ) {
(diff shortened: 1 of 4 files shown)


AlexGStapleton:
Working build:

[so-widgets-bundle.carousel-cls-improvements.zip](https://github.com/siteorigin/so-widgets-bundle/files/8179586/so-widgets-bundle.carousel-cls-improvements.zip)

Active on demo site. https://github.com/siteorigin/so-widgets-bundle/pull/1453/commits/3e8872e41a0745ac7a85004ce1a135e388ed778c affects all carousels and should work alongside all previous CLS improvements to prevent any noticeable CLS. Please briefly all carousels (add widget, view page and confirm no CLS is recorded).

Misplon:
Very minor but it looks like we aren't doing the same proportional sizing with default thumb from the setting vs external URL.

**Default Thumbnail:**
default-thumb

**Default Thumbnail External URL:**
default-thumb-external

I'm going to merge shortly. Let's perhaps solve this one in another PR.

Misplon (approved):

PR #1459, opened 2022-03-08
Widget Block Third Party Widget Changes

AlexGStapleton:
This PR will:
- Fix an issue that prevented ajax errors from showing these were present when trying to use a deactivated third party widget. This error could occur even when the widget was active and that would make the widget unusable.
- Prevent a third party widget data unintentionally having the wrong author.
- List the author in widget names for all third party widgets. This was previously the case for only certain widgets. `Widget Name by Author`.

You can test this PR using the [Livemesh SiteOrgin Widgets](https://wordpress.org/plugins/livemesh-siteorigin-widgets/) plugin.

### DIFF
--- a/compat/block-editor/widget-block.js
+++ b/compat/block-editor/widget-block.js
@@ -11,15 +11,14 @@
 	var Spinner  = components.Spinner;
 	var __ = i18n.__;
 
-	var ajaxErrorHandler = function( response ) {
+	var getAjaxErrorMsg = function( response ) {
 		var errorMessage = '';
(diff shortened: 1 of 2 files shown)


Misplon:
Thanks, mate. Are we expecting a widget to display in the list if not activated?

Edit_Page_“Block_Editor”_‹_SiteOrigin_—_WordPress

SiteOrigin_Widgets_‹_SiteOrigin_—_WordPress

AlexGStapleton:
Yes. I can change that if you would prefer.

Misplon (approved):

PR #1461, opened 2022-03-12
Widgets Block: Resolve notice when rendering if widget isn't selected

AlexGStapleton:
Resolve `( ! ) Notice: Undefined index: widgetData in app\public\wp-content\plugins\so-widgets-bundle\compat\block-editor\widget-block.php on line 149`


Misplon:
@AlexGStapleton Thanks, mate. How are you recreating and testing this issue?

AlexGStapleton:
You can trigger this error by adding a SiteOrigin Widgets Block, saving the page and then reloading the editor.

Misplon (approved):

PR #1463, opened 2022-03-13
Post Carousel: Add Autoplay Continuous Scroll

AlexGStapleton:
This PR is a follow up to https://github.com/siteorigin/so-widgets-bundle/pull/1462. It changes how the looping of the post carousel is handled when you reach the end of the post. Instead of transitioning back to the first item it "scrolls" to the first items in a more continuous fashion. Due to how visually different this is compared to the base method, it's behind a setting to prevent an unexpected change.

You can enable continuous scroll for the Post Carousel widget by enabling looping, autoplay, and the **Autoplay Continuous Scroll** setting.


Misplon:
Thanks, mate. Are we able to make the setting dependent on Loop Items?

Misplon (approved):

Misplon (approved):

PR #1482, opened 2022-04-06
Features: Center Align Left/Right Icon On Mobile

AlexGStapleton:
This PR will center the feature icon on mobile when the container position is set to right or left. This will result in the same alignment as the top/bottom alignments.

Screenshot 2022-04-06 at 23-29-41 Features CLS – SO


AlexGStapleton:
The updated branch with develop and the last commit resolves that for me. Demo page setup at https://demo.siteorigin.com/corp/1482-2/

Misplon (approved):

PR #1488, opened 2022-04-13
Accordion/Tab - Add Support for Custom Anchor Id

AlexGStapleton:


Misplon:
Thanks, mate. Please, give the attached layout a try with https://github.com/siteorigin/siteorigin-premium/pull/598.

[home.json.zip](https://github.com/siteorigin/so-widgets-bundle/files/8547671/home.json.zip)

The anchor issue doesn't seem to be present in the Accordion, only the Tabs Widget.

AlexGStapleton:
I've had a look at the provided layout and it's working as expected. The Anchor ID addon relies on the title field to use as the actual id and if the id (read: title) isn't unique, only the first instance of that id will be used - this is the case for all widgets supported by the Anchor ID. Try changing one of the titles to something else and then trying to link to the tab with the adjusted title you altered.

We could maybe add a field to each repeater that allowed for a custom anchor id. I don't think there will be many real world instances of a number of tabs with the same title present though.

Misplon (approved):

PR #1491, opened 2022-04-16
Social Media Buttons: Hover Adjustments

AlexGStapleton:
- Clean up selectors to prevent CSS output that's never used.
- Flat: Correctly override Icon color hover.
- Remove :visited and :active in favour of :hover and :focus
- Ensure hover styling isn't output when hover styles are disabled.


Misplon:
I'm seeing a hover color without the hover setting enabled. Are we good leaving this rule in place seen below?

Puro_–_Just_another_WordPress_site

Misplon:
Perhaps for another PR, but can you perhaps check out the below layout in Twenty Fifteen. There is a bit of a hover jump we could look to eliminate. It looks like a border-color initial rule is causing the non-hover icon color to not appear.

[home.json.zip](https://github.com/siteorigin/so-widgets-bundle/files/8501854/home.json.zip)

PR #1492, opened 2022-04-18
Image Widgets: Use `wp_get_loading_attr_default`

AlexGStapleton:
This PR alters the Image and Image Grid widget to use [wp_get_loading_attr_default](https://developer.wordpress.org/reference/functions/wp_get_loading_attr_default/). This will prevent a situation where the image can cause a large contentful paint due to being set to lazy load above the fold. For users who would like to fine-tune how many images are excluded from lazy loading they can use the `wp_omit_loading_attr_threshold` filter to control that.

To test this, please add two valid SiteOrigin Image widgets. Check the markup of the first image and confirm no loading, and confirm it's present in the second one. Repeat this process for the Image grid widget. Please note that if a featured image is set, that'll be excluded from the lazy load isntead.


Misplon:
Thanks, mate. Looks like the help link was removed from the Image Grid.

Misplon (approved):

Misplon (approved):

PR #1493, opened 2022-04-18
Social Media BTN: Account for Twenty Fifteen Link `border-bottom` Override

AlexGStapleton:
[Reported here](https://github.com/siteorigin/so-widgets-bundle/pull/1491#issuecomment-1100895003)


Misplon:
Thanks, mate. Looks good for all but Atom which is still showing a hover jump with hover styles enabled.

Misplon (approved):

PR #1512, opened 2022-05-10
Contact: reCAPTCHA v3

AlexGStapleton:
This PR adds [reCAPTCHA v3](https://www.google.com/recaptcha/admin) support to the contact form. 
Please test:
- upgrading an existing contact form with v2 enabled.
- v2.
- v3.

There's unfortunately no reliable way for us to detect the reCAPTCHA version type based on the provided key so if the user inputs an invalid one, there's not much we can do about it.

It might be a good idea to introduce a global setting to the Contact Form that allows for the reCAPTCHA to be loaded sitewide. This will allow for "improved verification" as reCAPTCHA will be able to monitor behaviour throughout the site.

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -234,7 +234,7 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 						'fields' => array(
 							'use_captcha' => array(
 								'type'    => 'radio',
-								'label'   => __( 'reCAPTCHA version', 'so-widgets-bundle' ),
+								'label'   => __( 'reCAPTCHA', 'so-widgets-bundle' ),
 								'default' => false,
 								'options' => array(
 									''   => __( 'Disabled', 'so-widgets' ),
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -253,18 +253,34 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 							),
 							'site_key'    => array(
 								'type'  => 'text',
-								'label' => __( 'reCAPTCHA Site Key', 'so-widgets-bundle' ),
+								'label' => __( 'reCAPTCHA v2 Site Key', 'so-widgets-bundle' ),
 								'state_handler' => array(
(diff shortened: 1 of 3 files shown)


Misplon:
Thanks, mate. 

Let's consider changing the label from "reCAPTCHA version" to "reCAPTCHA" so that the "Disabled" option makes a bit more sense.

Misplon:
@AlexGStapleton Is the form submitting normally to you when you test with recaptcha disabled? Test: https://demo.purothemes.com/polestar/sandbox/

Misplon:
With V3 keys inserted the following was seen in the console:

```
jquery.min.js?ver=3.6.0:2 Uncaught Error: Syntax error, unrecognized expression: #contact-form-8968,
    at Function.se.error (jquery.min.js?ver=3.6.0:2:13639)
    at se.tokenize (jquery.min.js?ver=3.6.0:2:21650)
    at se.select (jquery.min.js?ver=3.6.0:2:22477)
    at Function.se (jquery.min.js?ver=3.6.0:2:7116)
    at Function.s.find (jquery-migrate.min.js?ver=3.3.2:2:2230)
    at S.fn.init.find (jquery.min.js?ver=3.6.0:2:25047)
    at s.fn.init.S.fn.init (jquery.min.js?ver=3.6.0:2:25536)
    at new s.fn.init (jquery-migrate.min.js?ver=3.3.2:2:1827)
    at S (jquery.min.js?ver=3.6.0:2:1051)
    at jquery.theme.min.js?ver=1.8.1:1:5896
```

AlexGStapleton:
> Let's consider changing the label from "reCAPTCHA version" to "reCAPTCHA" so that the "Disabled" option makes a bit more sense.

Agreed. Changed.

> Is the form submitted normally to you when you test with recaptcha disabled? 

Fixed.

> With V3 keys inserted the following was seen in the console:

I can't seem to replicate this. It seems as though the error is related to the theme and, presumably, page jump. Can you please confirm which theme you're using? I tried Corp, Unwind and North but I can't seem to get theme to trigger this notice.

Misplon:
Would it be possible to store the keys for the V2 and V3 radio fields separately?

Misplon:
Looking great!

Misplon (approved):

PR #1514, opened 2022-05-18
Editor: fix bug with spaces in wrapper that caused incorrect displaying of editor widget

alex-one-v:


AlexGStapleton:
Hi Alex,

I've run a few tests and I'm unable to identify the issue you're referencing in your message. Can you please elaborate on the issue this PR is resolving?

Kind regards,
Alex

alex-one-v:
Hi Alex,

After more deeper investigation I've found out that the issue was because of some buggy 3d-party plugin. So you can decline the PR in fact. Sorry for taking up your time.

PR #1516, opened 2022-05-26
Sliders: Wait for Modern Parallax to Be Setup

AlexGStapleton:
This PR will help prevent a visual "jump" for sliders using the modern parallax. Prior to this PR, this jump is as a result of the image not parallaxing on load and that jump is the image correctly parallaxing.

To test this PR, please add a slider to the top of the page and to the second row. Add a slide or two and enable parallax for those slies. Save and view the page. If you see a jump shortly after load, try the PR. Otherwise, try simulating a slower connection to see if that allows you to replicate the jump.

### DIFF
--- a/js/slider/jquery.slider.js
+++ b/js/slider/jquery.slider.js
@@ -168,7 +168,7 @@ jQuery( function($){
 				if ( slidesWithModernParallax.length ) {
 					// Wait for the parallax to finish setting up before
 					// setting up the slider itself.
-					if ( ! slidesWithParallax.find( '.simpleParallax' ).length ) {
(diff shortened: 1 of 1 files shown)


Misplon:
Thanks.

I'm testing at https://demo.siteorigin.com/corp/sandbox/.
I've uploaded this branch.
Parallax only seems to work once I've opened the browser console.

https://user-images.githubusercontent.com/789159/171041993-198fd610-2776-44e9-a1fe-57a354bda455.mov

Misplon:
@AlexGStapleton A build from this branch will have to be re-uploaded when attending.

Misplon (approved):

PR #1521, opened 2022-05-30
Hero: Better Top Padding Handling

AlexGStapleton:
This PR ensures that the top padding unit of measurement actually matters. Before this PR the Top and Bottom Padding and Extra Top Padding are simply added together and the unit of measurement isn't really factored in. After this PR the unit of measurement does matter.

To test:
- Add A Hero widget with a frame.
- Give the Hero some height (be it frame content or actual Height).
- Set Padding and Extra Top Padding using two different unit of measurements.
- Do the same for the Mobile version of those settings.
- Save and view the Hero.
- Inspect the Hero's `.sow-slider-image-wrapper` for the padding.

### DIFF
--- a/widgets/hero/hero.php
+++ b/widgets/hero/hero.php
@@ -488,13 +488,12 @@ class SiteOrigin_Widget_Hero_Widget extends SiteOrigin_Widget_Base_Slider {
 				$settings = $instance['layout']['mobile'];
 
 				$meas_options['slide_height_responsive'] = ! empty( $settings['height_responsive'] ) ? $settings['height_responsive'] : '';
-				$meas_options['slide_padding_responsive'] = ! empty( $settings['padding'] ) ? $settings['padding'] : '';
-				$meas_options['slide_padding_sides_responsive'] = ! empty( $settings['padding_sides'] ) ? $settings['padding_sides'] : '';
+				if ( ! empty( $settings['padding'] ) || ! empty( $settings['extra_top_padding'] ) ) {
+					$meas_options['slide_padding_responsive'] = ! empty( $settings['padding'] ) ? $settings['padding'] : 0;
 
-				if ( ! empty( $settings['extra_top_padding'] ) ) {
-					// Add extra padding to top padidng.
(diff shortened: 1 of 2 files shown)


Misplon:
@AlexGStapleton This PR could have a significant impact on existing Hero instances. If for example, px was used for the first field and percentage for the second, say 50 and 100, the result of this PR will be very noticeable. 150px will change to 50px + 100px. Do you want to consider a migration?

AlexGStapleton:
That's true. The last commit adds a migration method to account for a situation where the user has and hasn't set Padding. Please test with a combination of with and without

Misplon:
I don't seem to be winning. Do I need a build? My test layout is below:
[block-editor.json.zip](https://github.com/siteorigin/so-widgets-bundle/files/8800326/block-editor.json.zip)

Ignore the name.

AlexGStapleton:
Thanks. The provided layout helped identify an unexpected situation of unintended repeated migration when not actually possible and once that was fixed the provided layout worked for me. Let me know how you go.

Misplon:
Thanks, mate. The issue appears to persist.

`Warning: Undefined array key "extra_top_padding" in /Users/amisplon/Sites/siteorigin/wp-content/plugins/so-widgets-bundle/widgets/hero/hero.php on line 494`

AlexGStapleton:
> I've adjusted Top and bottom padding from 50 to 51. Extra top padding is still at 100%. At what point does the new behaviour take over?

Before you initially saved. After saving, the updated `padding_top_extra` setting will be in place so the migration won't be triggered. To be able to trigger it now you need to get the old setting. To do this:

- Switch to develop.
- Open Hero and set `Extra Top Padding`.
- Switch back to this branch.
- Open the Hero widget you set up and check the `Extra Top Padding`.

This is very much a dev build specific situation and won't occur for normal users. Also, you don't need to follow the initial test instructions as you've already done that.

As a reminder,  the migration code will only be triggered if `extra_top_padding` is set (only possible before saving) and not set to 0px (the default).

Misplon (approved):

PR #1522, opened 2022-06-01
Contact: Add Really Simple CAPTCHA Integration

AlexGStapleton:
This PR adds integration for the [Really Simple CAPTCHA](https://wordpress.org/plugins/really-simple-captcha/) plugin. This allows users to add a DSGVO complaint captcha that's widely supported and used (600k installs).

To test:

- Install Really Simple CAPTCHA
- Open a contact form or add a new one.
- Open the Spam Protection Settings group and open the Really Simple CAPTCHA settings, and tick Use Really Simple CAPTCHA.
- Fill Out the form incorrectly and confirm the rejection.
- Fill it out correctly and confirm the submission.

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -346,6 +346,23 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 							),
 						)
 					),
+
+					'simple' => array(
+						'type'   => 'section',
+						'label'  => __( 'Really Simple CAPTCHA', 'so-widgets-bundle' ),
+						'fields' => array(
+							'enabled' => array(
+								'type'    => 'checkbox',
+								'label'   => __( 'Add Really Simple CAPTCHA', 'so-widgets-bundle' ),
+								'description' => sprintf(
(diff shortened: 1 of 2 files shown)


Misplon:
@AlexGStapleton Should we try to put the Captcha error message above the field as we do with other fields if they are required and you don't fill them out?

AlexGStapleton:
> Thanks, mate. Are we expecting the Field Labels > Position setting to be functional? For example, if Below is selected.

Currently no, but we could add support for some of them. Hm. How do you want to handle the Inside field position? Default to top (read: default)? Show no label? Another option would be to completely remove the label altogether.

> Should we try to put the Captcha error message above the field as we do with other fields if they are required and you don't fill them out?

Done.

Misplon:
For label positions we can't support, I reckon we default to Top.

AlexGStapleton:
Sounds good. The last PR allows for that. Here's a test layout that includes all of the label positions.

https://drive.google.com/uc?id=1_0WMv7Q0NAAHDLgNMIkhHn5qM0mTRCY1

Misplon (approved):

PR #1532, opened 2022-06-09
Upgrade Font Awesome 5.15.1 to 6.1.1

AlexGStapleton:
Build requires https://github.com/siteorigin/plugin-build/pull/32

This PR allows for Font Awesome 6.1.1. Some migration code was required due to the extensive number of name updates by Font Awesome but that should be completely accounted for. Please try


Misplon:
@AlexGStapleton This seemed to fail for me. I used the adjust icon for the first feature here:

https://demo.siteorigin.com/corp/sandbox-two/

After upgrading, the icon is missing.

AlexGStapleton:
Ah. I've noticed an issue with the migration - it almost never runs. I've changed it so it always run (there's so many replacements at this point it just makes sense) and improved the performance for the more modern format.

AlexGStapleton:
I'm not seeing that.  I've run a test using two different computers using [this build](https://drive.google.com/uc?id=17kupZpNYZhEsErjGdYw-AcjpVW5_1pQw) and can't cause this. I've also uploaded it to the demo site:

https://demo.siteorigin.com/corp/sandbox-2/


Hm. Maybe it's caching related? Try clearing your browser cache to see if that helps. Unsure why it would be that but if it is, I'll look into versioning the font URLs to prevent this. For reference, here's what I'm seeing:

PR #1533, opened 2022-06-10
Ensure We're Able to Detect Widget Before Enqueuing Widget Script

AlexGStapleton:
[Reported here](https://wordpress.org/support/topic/php-8-0-8-1-version/)

This PR resolves the following warning:

`Warning: Undefined array key “1” in /web/htdocs/www.vendicompraffitta.it/home/wp-content/plugins/so-widgets-bundle/so-widgets-bundle.php on line 843`

There's no reliable way for us to test this at this time as it's dependent on an unknown third party widget. Sending a build to the user for confirmation.

### DIFF
--- a/so-widgets-bundle.php
+++ b/so-widgets-bundle.php
@@ -840,10 +840,13 @@ class SiteOrigin_Widgets_Bundle {
 							/* @var $widget SiteOrigin_Widget */
 							$opt_wid = get_option( 'widget_' . $widget->id_base );
 							preg_match( '/-([0-9]+$)/', $id, $num_match );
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
I've sent a build to the user to test.

Misplon:
Thanks, we'll wait for the test results.

AlexGStapleton:
While it didn't help the user, the change itself is a good idea and worthwhile - it's always a good idea to ensure something exists before using it and we previously weren't doing that. I'm going to mark this as ready to review. To test it, simply a third-party widget (like the Archives widget) and one of our widgets display as expected.

PR #1540, opened 2022-06-15
Image: Add External Image Size Setting

AlexGStapleton:
This PR introduces the External Image Size setting to the Image widget. To do this [State Emitter Detection For External Fields](https://github.com/siteorigin/so-widgets-bundle/commit/edf50817453af51c5148ae7dee58b5ce04d343e1)
was required. An adjustment to the Slider widget was also required to maintain compatibility for the Slider widget. To test change, simply set a Foreground image and confirm the Height setting appears.

### DIFF
--- a/base/inc/fields/js/media-field.js
+++ b/base/inc/fields/js/media-field.js
@@ -6,6 +6,7 @@
 		var $field = $( this );
 		var $media = $field.find('> .media-field-wrapper');
 		var $inputField = $field.find( '.siteorigin-widget-input' ).not('.media-fallback-external');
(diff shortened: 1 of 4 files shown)


AlexGStapleton:
To confirm, are you using the Image Size field or the External Image Size field? To control the size of the external image you'll need to adjust the external Image Size (which has two image sizes, Default and Custom Size). The Image Size cannot affect the External Image because we can't size an external image exactly like WordPress does.

If you're not seeing an external image size (it's only visible after adding an External Image), try doing a build.

AlexGStapleton:
How does that look?

Issue #1523, opened 2022-06-01
Social Media Buttons: Add Patreon


Misplon:

AlexGStapleton:
Currently working on this but as this is only available in Font Awesome 6+ it's taking a little longer than normal. I've been able to get the icons to import automatically, added migration code to account for any renames and get it to display. I've run into an issue with certain icons not rendering in the admin, but once that's resolved I should be good to proceed.

Unfortunately, in 5.15.4 Tripadvisor requested their icon to be removed so we don't have an icon for it anymore. How do you want to handle the SiteOrigin Social Media Button?

https://github.com/FortAwesome/Font-Awesome/blob/master/UPGRADING.md
Is there any other icons you would like added at the same time? https://fontawesome.com/search?s=brands

Misplon:
Super, thanks for the update.

How did we handle the discontinuation of Google Plus in the past? Do you want to remove the TA icon with a notice in the changelog?

Happy to add any you'd like to or to wait for requests.

AlexGStapleton:
We switched the Google Plus icon over to the standard Google icon. https://github.com/siteorigin/so-widgets-bundle/pull/1418
I'm not too sure what we could use for the TA icon but if you have an idea you think could work we could do that.

Misplon:
If the icon is in use, let's use the plane or the suitcase icon. Which one you keen for?

PR #1552, opened 2022-06-22
Social media Button Changes

AlexGStapleton:
Resolves https://github.com/siteorigin/so-widgets-bundle/issues/1523

This PR:

- Adds compatibility for https://github.com/siteorigin/so-widgets-bundle/pull/1532. Brand icons wouldn't show correctly unless the specific icon font library was set. To test this, confirm any Brand network works and that the Tripadvisor network works.
- Changes the Tripadvisor Icon to Suitcase due to the removal of that icon.
- Adds Patreon.


AlexGStapleton:
Fixed.

Issue #1554, opened 2022-06-23
Contact Form: Account for long checkbox labels on mobile


Misplon:
Account for long checkbox labels on mobile. I think this is only happening on iOS.

To test, set Sandbox to public and view https://demo.siteorigin.com/corp/sandbox/ on mobile. Remember to set the page back to private when done. 

![Photo_-_Google_Photos](https://user-images.githubusercontent.com/789159/175379162-af6acfde-b1c3-4278-bf5d-791ee9f9a5c7.jpg)

We can probably resolve this by setting the input to flex: 1 and the label to flex: 2. For the moment, I've set the flexbox to column for the reporting user.

Ref: https://stackoverflow.com/questions/34220553/flexbox-resizes-a-checkbox-too-small

PR #1558, opened 2022-06-30
Contact: Maintain Radio/checkbox Input Size on iOS

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1554

This issue does not happen on Android so you'll need to test this PR using an iOS device. [Test layout](https://drive.google.com/uc?id=1gwV4O_FNLlXyoKUEUTHdgeCaHXuZi2xf).

### DIFF
--- a/widgets/contact/styles/default.less
+++ b/widgets/contact/styles/default.less
@@ -144,6 +144,7 @@
 				float: left;
 				margin-right: 0.5em;
 				height: auto;
+				flex: none;
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
> Can we perhaps work on a follow-up to ensure the radio or checkboxes aren't slightly cut off?

I'm having trouble replicating that result. Are you getting that while using the release build? I tried using an iPhone and an iPad - both using the latest iOS.

Issue #1537, opened 2022-06-14
Post Carousel: Attempt to detect if a post is partially displayed


Misplon:
If post navigation arrows aren't displayed and if a post is only partially visible, display the navigation arrows.

Update when we know more. [HelpScout](https://secure.helpscout.net/conversation/1916726734/60951?folderId=472482).

PR #1559, opened 2022-07-02
Carousel: Show Nav Item is Partially Visible

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1537

This PR will result in the navigation showing if the last post is partially visible when the navigation would previously not be shown.


AlexGStapleton:
This PR is intended to allow for that after clicking the navigation. Is that not happening for you? What's happening for you instead?

AlexGStapleton:
I've been able to replicate this issue when there is 2 posts. I was testing with a category of 4. This is likely going to be quite complicated to account for. Hm.

AlexGStapleton:
Oddly, I can't replicate that locally. :(
Regardless, I should be able to account for both with the same fix.

AlexGStapleton:
How does that look?

PR #1560, opened 2022-07-05
Anything Carousel: Adaptive Height

AlexGStapleton:
This PR adds a setting to the Anything Carousel Widget which allows items to be dynamically sized based on the currently visible items rather than all items. This means in a situation where a set of items has a large amount of text, but another set (not visible at the same time) doesn't, it would previously be sized to account for the first set. Now it'll be sized on the visible items.

This setting is disabled for pre-existing Anything Carousel widgets and enabled for new instances. This setting can be enabled in the Settings.

### DIFF
--- a/js/carousel.js
+++ b/js/carousel.js
@@ -48,6 +48,7 @@ jQuery( function ( $ ) {
 					),
 				variableWidth: $$.data( 'variable_width' ),
 				accessibility: false,
+				adaptiveHeight: carouselSettings.adaptive_height,
 				cssEase: carouselSettings.animation,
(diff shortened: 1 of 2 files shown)


AlexGStapleton:
Oh! There was an unexpected requirement for `slidesToScroll` to be set to 1. I've tested removing this requirement and I'm not noticing an issue so I've committed it to our fork. I've tested the provided layout and it works after removing this requirement.

AlexGStapleton:
Fixed.

> I think we should default this setting to disabled.

Done.

PR #1564, opened 2022-07-10
Allow Other Plugins To Filter Global Setting Defaults

AlexGStapleton:
This PR allows for developers to filter the global setting defaults. It uses the pre-existing `siteorigin_widgets_settings_form` filter so existing code will start to filter the defaults without adjustments.


AlexGStapleton:
Documented here https://github.com/siteorigin/docs/pull/134

PR #1567, opened 2022-07-15
Contact: Avoid Notice If Missing Certain fields

AlexGStapleton:
This PR will prevent a notice when not including a name or message field that that occurs after submitting the form.

```
Warning: Undefined array key "message" in /bitnami/wordpress/wp-content/plugins/so-widgets-bundle/widgets/contact/contact.php on line 1470

Warning: foreach() argument must be of type array|object, null given in /bitnami/wordpress/wp-content/plugins/so-widgets-bundle/widgets/contact/contact.php on line 1470
```


AlexGStapleton:
This notice will only after submitting the form. I've updated the initial message to mention this.

PR #1574, opened 2022-08-04
HTML Validation Improvements

bbceg:
Don't display empty inline style tags.
When displaying inline style tags, check if the theme supports html5.


Misplon (approved):

Misplon:
Thanks, @bbceg.

PR #1575, opened 2022-08-07
Release/1.39.0

Misplon:
@AlexGStapleton 

- Try setting a pagination top margin of 10 in the Standard template. Save. Switch to the Grid template. The top margin value doesn't appear to reset. Is that expected?

- We're going to need to re-look at the way the grid works. If you have fewer posts than columns, `space-between` doesn't help us. If you have for example, two posts and four columns you'll end up with a post at flex start and a post at flex end.


Misplon:
The above points aren't necessarily release blockers, we could proceed.

Misplon:
@AlexGStapleton Portfolio has the same issue as Grid if the column count exceeds the available posts. An easy example being 2 posts with four columns. No issues with Masonry.

Misplon:
@AlexGStapleton With Portfolio selected, try to disable Filter Categories. The section header remains but the section can't be opened.

AlexGStapleton:
> Try setting a pagination top margin of 10 in the Standard template. Save. Switch to the Grid template. The top margin value doesn't appear to reset. Is that expected?

I can't replicate this so, no, this isn't expected. Are you able to replicate this with a build?
I tried with not only Grid but every template and couldn't replicate it. Also tried new and existing versions of the blog widget.

> We're going to need to re-look at the way the grid works ... Portfolio has the same issue as Grid if the column count exceeds the available posts

Fixed.

> With Portfolio selected, try to disable Filter Categories. The section header remains but the section can't be opened.

By section header, do you mean the actual filter categories section above the posts or categories on post hover?

If it's the filter categories, I can't replicate this.
If it's the categories you see when hovering items, that's controlled by the categories section.
If neither, can you please elaborate on what you mean by section head?

Misplon:
@AlexGStapleton I'll send a screencast in Slack showing both.

Issue #1535, opened 2022-06-10
Features: Add a setting for Read More link equal height


Misplon:
Perhaps the setting label could be:

`Bottom align the More link text`

PR #1577, opened 2022-08-13
Features: Add `Bottom align the More link text` setting

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1535

Please test using all icon container positions.

### DIFF
--- a/widgets/features/features.php
+++ b/widgets/features/features.php
@@ -250,6 +250,12 @@ class SiteOrigin_Widget_Features_Widget extends SiteOrigin_Widget {
 				'default' => true,
 			),
 
+			'more_text_bottom_align' => array(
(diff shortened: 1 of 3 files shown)


Misplon:
Thanks, mate.

- I think we should go with default disabled for this setting.
- Please find screenshots and a la
you below. The first screenshot is with the setting disabled. The second screenshot shows the effect of the setting enabled. To recreate, enable and disable the More bottom align setting and check the results.





[page-builder.json.zip](https://github.com/siteorigin/so-widgets-bundle/files/9331758/page-builder.json.zip)

AlexGStapleton:
How does that look?

Misplon (approved):

Issue #1576, opened 2022-08-07
Blog: Add a conditional Excerpt Length setting


Misplon:
Add a conditional Excerpt Length setting. The setting will be especially helpful for templates like Alternate if the available space is limited.

![Sandbox_-_Vantage](https://user-images.githubusercontent.com/789159/183291339-df78c673-8d8e-4b4d-8b7e-67c24669f064.jpg)

Issue #1254, opened 2021-03-04
State Emitter: Add multi-conditional support


AlexGStapleton:
We should introduce the ability to have multiple settings to affect a conditional. Like standard PHP, we should be able to set `AND` or `OR` operators to allow for the conditional to have specific criteria dependent on needs.

AlexGStapleton:
This is actually already possible using [the conditional emitter](https://siteorigin.com/docs/widgets-bundle/form-building/state-emitters/#heading-the-conditional-callback). [Here's an example](https://github.com/siteorigin/so-widgets-bundle/blob/4a10395a0f7f8d09e8315e79e809b9ae8f396018/widgets/slider/slider.php#L108-L109) of where we do this in a widget. It could maybe be expanded to allow for different options, but I think that's best handled by a custom emitter and potentially https://github.com/siteorigin/so-widgets-bundle/issues/1569

PR #1578, opened 2022-08-13
Blog: Add Excerpt Length

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1576

Please test with and without excerpt length adjustment. Please also confirm `siteorigin_widgets_blog_excerpt_length` still works as intended. Test snippet:

```
add_filter( 'siteorigin_widgets_blog_excerpt_length', function( $length ) {
  return 10;
} );
```


Misplon:
Thanks, mate.

- If you set a count over 55, is it working?
- Let's make the setting conditional on selecting Excerpt but not on enabling excerpt read more.
- I'm keen to make the excerpt dots the same as Corp ... vs. the default of [...]

Misplon:
Super; thanks. When you have a moment, can you perhaps check out the fatal in template/base.php. Cheers!

Misplon:
Thanks, mate. I'm seeing settings remain when switching to Portfolio. Is that happening for you or just a M1 issue?

https://user-images.githubusercontent.com/789159/184901149-413022f8-c526-45c4-aa81-37a0c86fb596.mov

AlexGStapleton:
Ah. That's an oversight. Due to multi-conditional state emitters not currently being possible (https://github.com/siteorigin/so-widgets-bundle/issues/1254), we have to choose whether we want to show Excerpt Length when Post Content is set to Excerpt or when a template that supports excerpts is enabled.

I suspect the latter will be your preference but I'll wait for your reply before making any changes.

Misplon:
We can't do both? Both would be ideal.

Misplon:
If Excerpt Length is displayed when Post Content is set to Excerpt would Excerpt Length be hidden for Portfolio as there isn't a Post Content setting in the Portfolio template.

AlexGStapleton:
> We can't do both?

We cannot. https://github.com/siteorigin/so-widgets-bundle/issues/1254 would be required for that - it's quite a large change to make just for this.

> If Excerpt Length is displayed when Post Content is set to Excerpt would Excerpt Length be hidden for Portfolio as there isn't a Post Content setting in the Portfolio template.

Hm. By default, no (that's just conditionals, which aren't supported), however, if I juggle a few things around with preset data I can do this. Okay. Done. The last PR accounts for this.

Misplon (approved):

PR #1581, opened 2022-08-16
Blog: Prevent Potential Pagination Disaplay Issue After Disabling Addon

AlexGStapleton:
This will prevent a situation where the pagination can incorrectly render if the addon was previously active and it no longer is.


Misplon:
Thanks, mate. This PR seems to remove the pagination. I've uploaded a build to the demo server.

https://demo.siteorigin.com/corp/sandbox-two/

AlexGStapleton:
Oh! Sorry about that!

Misplon (approved):

PR #1584, opened 2022-08-24
Contact Select: Add `Allow multiple selections` Setting

AlexGStapleton:
This PR will allow users to set the Select field to allow multiple selections. Selections will be listed in the email `with each, selected, item, separated, by a coma`.

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -1501,6 +1501,12 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 			}
 			$body .= "\n\n";
 		}
+		foreach ( $email_fields['message'] as $m ) {
+			$body .= '<strong>' . $m['label'] . ':</strong>';
+			$body .= "\n";
+			$body .= htmlspecialchars( $m['value'] );
+			$body .= "\n\n";
+		}
 		$body = wpautop( trim( $body ) );
 
 		if ( $this->is_dev_email($instance['settings']['to']) || empty( $instance['settings']['to'] ) ) {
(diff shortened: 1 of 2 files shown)


widgets/contact/contact.php
```diff
@@ -192,6 +192,15 @@ function get_widget_form() {
 						)
 					),
 
+					'multiple_select' => array(
+						'type'        => 'checkbox',
+						'label'       => __( 'Allow multiple selections', 'so-widgets-bundle' ),
+						'state_handler' => array(
+							'field_type_{$repeater}[select]' => array( 'show' ),
+							'_else[field_type_{$repeater}]'      => array( 'hide' ),
```
Misplon, line 200: @AlexGStapleton Is this spacing correct? Thanks.

Misplon:
@AlexGStapleton Can you confirm that the notification email to admin arrives normally when selecting all options. Mine seems to be missing the Message and the select data.

AlexGStapleton:
Ah. There was an issue that resulted in forms after the selection not processing correctly. That's now fixed.

Misplon:
Thanks, mate. Please try the layout below and select some of the select items. The email contents aren't being delivered for me.

[so-widgets-bundle.2508-contact-select.zip](https://github.com/siteorigin/so-widgets-bundle/files/9427857/so-widgets-bundle.2508-contact-select.zip)

AlexGStapleton:
It appears you've provided a build rather than a layout. That's not intentional, right? I ask as the build looks to be named in reference to this PR. Regardless, I've spotted that the messages aren't being directly output so I've set them to output. How does that look?

Misplon:
Thanks, mate. Sorry for the wrong upload for the layout.

Are you seeing the semi-colon in the notification email?

Testing1_-_amisplon_gmail_com_-_Gmail

AlexGStapleton:
Is that fields label empty? If I remove the label that's how it appears for me:
Screenshot 2022-08-28 at 01-18-48 MailHog

Misplon:
Quite right!

Misplon (approved):

PR #1589, opened 2022-09-07
Blog: More Specific Article Reset

AlexGStapleton:
This PR will ensure some of our adjustments using the article are more specific. This is important as it appears certain themes (such as Vantage) apply styles to the `article` element containing the post using `article.post`. In the case of Vantage, that results in unexpected padding while using the Alternate template.

Screenshot 2022-09-08 at 02-06-09 SO

### DIFF
--- a/widgets/blog/css/style.less
+++ b/widgets/blog/css/style.less
@@ -8,7 +8,7 @@
 	}
 
 	// Global Styles.
-	article {
+	.sow-blog-posts article {
 		border: none;
 		box-sizing: border-box;
 		margin: 0;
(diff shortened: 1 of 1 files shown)


Misplon:
Thanks, mate. Further changes will be needed. For example, the following has lower specificity than the updated global rule in this PR:

`.so-widget-sow-blog-alternate-be65dae67c95-26891 .sow-blog-layout-alternate article`

AlexGStapleton:
The last commit should avoid that while still accomplishing the goals of this PR.

Misplon (approved):

PR #1590, opened 2022-09-07
Blog Alternate: Left Align Odd Images

AlexGStapleton:
This PR will left align odd images in the Alternate Template. Currently, they're centered. It also cleans up the CSS for the featured image in this layout.

Screenshot 2022-09-08 at 02-15-46 SO

### DIFF
--- a/widgets/blog/styles/alternate.less
+++ b/widgets/blog/styles/alternate.less
@@ -58,34 +58,26 @@
 		margin-top: 0;
 		width: 100%;
 
+		.sow-entry-thumbnail {
+			width: 100%;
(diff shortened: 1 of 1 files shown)


Misplon:
The centering inherited from Corp was intentional. Without centering there is a big gap between the image and text.

Page_Builder_–_SiteOrigin

Misplon:
The gap depends on the page width. I guess maintaining the alternate positioning is more inline with the layout name. Centering would be best for users using this layout in a wide container though.

Misplon (approved):

PR #1592, opened 2022-09-08
Blog: Add `siteorigin_widgets_blog_excerpt_trim`

AlexGStapleton:
This PR adds a filter that controls the trim indicator when the excerpt is generated. Filter example:

```
add_filter( 'siteorigin_widgets_blog_excerpt_trim', function( $trim ) {
	return '[...]';
} );
```

You can also completely remove the excerpt (leaving only the continue reading link) using the following PHP:

```
add_filter( 'siteorigin_widgets_blog_excerpt_length', function( $length ) {
	return 0;
} );

add_filter( 'siteorigin_widgets_blog_excerpt_trim', function( $trim ) {
	return '';
} );
```

### DIFF
--- a/widgets/blog/blog.php
+++ b/widgets/blog/blog.php
@@ -1036,7 +1036,11 @@ class SiteOrigin_Widget_Blog_Widget extends SiteOrigin_Widget {
 		$excerpt = get_the_excerpt();
 		$excerpt_add_read_more = str_word_count( $excerpt, 0, '0..9' ) >= $length;
 		if ( ! has_excerpt() ) {
-			$excerpt = wp_trim_words( $excerpt, $length, '...' );
+			$excerpt = wp_trim_words(
+				$excerpt,
+				$length,
+				apply_filters( 'siteorigin_widgets_blog_excerpt_trim', '...' )
+			);
 		}
 
 		if ( $settings['read_more'] && ( has_excerpt() || $excerpt_add_read_more ) ) {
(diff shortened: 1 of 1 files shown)


Misplon:
@AlexGStapleton Thanks, mate. Is this PR working for you? I'm not able to adjust the excerpt trim indicator. If I directly edit `alter_excerpt_more_indicator` I see a change.

AlexGStapleton:
I can confirm this is working for me. I also sent two builds of this PR (one before and after the latest release) and the user confirmed it was working both times. I added the provided snippet using Code Snippets so there shouldn't be any specific timing requirements. I'm not too sure why it wouldn't be working for you. Can you send me a copy of your layout so I can run some tests using it?

Misplon:
Thanks, mate. I'm now running this branch on the demo server. Please, check https://demo.siteorigin.com/corp/blog-widget-grid-template/.

[so-widgets-bundle.blog-excerpt-trim-1009-1959.zip](https://github.com/siteorigin/so-widgets-bundle/files/9541124/so-widgets-bundle.blog-excerpt-trim-1009-1959.zip)

AlexGStapleton:
Ah. I worked out what was going on. There's an additional area we were applying the dots and I had thought that was purely for full content but I was mistaken. It'll now work as expected even if the `excerpt_length` isn't very low.

Misplon (approved):

Issue #1593, opened 2022-09-09
Contact Form: Add a Honeypot spam option


Misplon:
Ref: https://docs.gravityforms.com/spam/#h-honeypot
Feedback to provided at https://wordpress.org/support/topic/sugestion-12/.

PR #1595, opened 2022-09-12
Contact: Add `Honeypot` and `Browser Check` Anti-spam Options

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1593

This PR implements two new anti-spam options the honeypot and a browser check. The browser check is stronger overall, but it requires JavaScript to be enabled so not all users will be okay with that.

To test the honeypot option: find the (CSS) hidden field in the markup and set it to show. Give it a value and submit it. If it failed after inputting data into the field, it worked.

To test the browser check, disable JavaScript.

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -236,7 +236,16 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 				'label'  => __( 'Spam Protection', 'so-widgets-bundle' ),
 				'hide'   => true,
 				'fields' => array(
-
+					'honeypot' => array(
+						'type'        => 'checkbox',
+						'label'       => __( 'Honeypot', 'so-widgets-bundle' ),
+						'description' => __( 'Add an invisible field to your form that if filled out, the form will be rejected.', 'so-widgets-bundle' ),
+					),
+					'browser_check' => array(
+						'type'        => 'checkbox',
+						'label'       => __( 'Browser Check', 'so-widgets-bundle' ),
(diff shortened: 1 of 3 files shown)


Misplon:
Thanks, mate.

Since we gave Really Simple its own section, should we add new sections for Honeypot and Browser Check to keep the Spam section consistent?

AlexGStapleton:
I think it makes sense to separate the Really Simple Cache into a section as it requires an additional plugin and has related settings (visible after enabled). Honeypot and Browser check will never have related settings and don't require an additional plugin or third-party service (like Askimet, reCAPTCHA does). While I can give them their own section, they don't function in the same way so I think it would be more consistent (no need to think about a plugin or third-party service) to keep them separate. Thoughts?

Misplon:
Sounds good.

widgets/contact/contact.php
```diff
@@ -236,7 +236,16 @@ function get_widget_form() {
 				'label'  => __( 'Spam Protection', 'so-widgets-bundle' ),
 				'hide'   => true,
 				'fields' => array(
-
+					'honeypot' => array(
+						'type'        => 'checkbox',
+						'label'       => __( 'Honeypot', 'so-widgets-bundle' ),
+						'description' => __( 'Add an invisible field to your form that if filled out, the form will be rejected.', 'so-widgets-bundle' ),
```
Misplon, line 242: @AlexGStapleton You ok with the following?

`Adds a hidden form field that only bots can see. The form will reject the submission if the hidden field is populated.`

AlexGStapleton: Yes, that's much better.

Misplon (approved):

PR #1600, opened 2022-09-15
Blog: Add Support for `Featured Image Size` Setting

AlexGStapleton:


Misplon:
@AlexGStapleton Thanks, mate.

The default option isn't loading for me when I change to Offset or any of the templates using the default featured image. The default seems to load if I use:

`"featured_image_size": "full"`

Are you ok with using the above instead of `"featured_image_size": "post-thumbnail"` ?

Misplon:
@AlexGStapleton You keen to move Featured Image Size under the Featured Image checkbox?

AlexGStapleton:
> The default option isn't loading for me when I change to Offset or any of the templates using the default featured image.

Hm. Oh. Okay. Currently, we're using the default image size which is `post-thumbnail` which is dependent on the theme to register. If they don't, the full size is used - that's why it's not working as expected for you. This means that on certain sites the blog image may be slightly different to what we would otherwise expect.

I can change it to full but in that situation, I should completely drop `post-thumbnail` for consistency. Are you okay with that? It may result in changes to certain sites but overall it's much more consistent. Another option would be to leave the size field blank for templates that don't include pre-defined image sizes.

Misplon:
Thanks, mate. Do you have a preference here?

> I can change it to full but in that situation, I should completely drop post-thumbnail for consistency. Are you okay with that? It may result in changes to certain sites but overall it's much more consistent. Another option would be to leave the size field blank for templates that don't include pre-defined image sizes.

AlexGStapleton:
Dropping post-thumbnail for consistency feels like the way to go.

Misplon:
Sounds good; let's proceed. Thanks.

Misplon:
Thanks, mate. Can you check the Portfolio layout for us?

Misplon:
@AlexGStapleton I'm not sure what changed, but the issue seems to be resolved https://demo.siteorigin.com/corp/blog-widget-portfolio-template/. the last time I uploaded a build, there was no spacing between portfolio items. If you can confirm that Portfolio template is looking good on your end, we can proceed.

Misplon:
The issue I'm seeing isn't related to this PR. I can replicate the issue (sometimes) with the release version. 

Classic_Editor_–_Demo

AlexGStapleton:
> The issue I'm seeing isn't related to this PR. I can replicate the issue (sometimes) with the release version.
> 
> Classic_Editor_–_Demo

Hm. Oh. Are there posts that don't have a featured image set? I just ran some tests and noticed that I had a post that didn't have one set and it resulted in an empty space. How do you want to handle that? Skip the post? Maybe we could add a placeholder.

Misplon:
Thanks, mate. 

Is that the case here? https://demo.siteorigin.com/corp/sandbox/. The issue seems present without a post missing a featured image.

We could try a placeholder that has a #f2f2f2 background for now. Perhaps we can do it with only CSS?

PR #1601, opened 2022-09-15
Carousel: Only Allow Positive Slides to scroll and Slides to show

AlexGStapleton:
This PR resolves an issue that occurs if you set negative Slides to scroll or Slides to show value. Basically, the browser hangs. This PR ensures the values can only be positive. For example, `-2` is treated as `2`.


Misplon:
@AlexGStapleton Can we perhaps take this a bit further and convert the settings value to a positive after save?

Misplon (approved):

PR #1615, opened 2022-09-27
Google Fonts GDPR: Add Support For `siteorigin_web_font_url `

AlexGStapleton:
This PR will allow users to replace Google Fonts with GDPR-friendly mirrors such as [Bunny Fonts](https://fonts.bunny.net/).

Test snippet:
```
add_filter( 'siteorigin_web_font_url ', function( $url ) {
	return 'https://fonts.bunny.net/css';
} );
```

To test this snippet please add a SiteOrigin Headline widget, add some text and then set a font that's from Google Fonts.

### DIFF
--- a/base/base.php
+++ b/base/base.php
@@ -152,7 +152,7 @@ function siteorigin_widget_get_font($font_value) {
 			$font['weight_raw'] = filter_var( $font['weight'], FILTER_SANITIZE_NUMBER_INT );
 			$font['style'] = ! is_numeric( $font['weight'] ) || $font['weight'] == 'italic' ? 'italic' : '';
 		}
-		$font['url'] = 'https://fonts.googleapis.com/css?family=' . $font_url_param;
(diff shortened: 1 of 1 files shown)


Misplon:
@AlexGStapleton Do you perhaps want to rename this and the premium filter to something more generic like siteorigin_web_font_url ?

Misplon:
@AlexGStapleton Thanks, mate. Do you want to get a bit more detailed with the ID?

`sow-google-font-inter-css`

AlexGStapleton:
I've thought about this, and I think it's better if we don't. While this PR allows for a global method of disabling the Google Font import, users previously had to rely on manually checking for stylesheets and depending on how we alter the style name we could break their code. If this wasn't GDPR-related it might be worth that risk, but as it is I don't think it is as it's entirely possible for it to go unnoticed until it's too late.

Misplon (approved):

Misplon:
Sounds good. Thanks.

PR #1621, opened 2022-10-06
Masonry: Add Title & Related Settings

AlexGStapleton:
This PR introduces the Display Image Title setting which allows users to display the image's title directly above or below the image. Please note that due to how the Masonry works, the title may result in some items that don't have a 1:1 aspect ratio if those other items don't have a title or similar length title.


Misplon:
Thanks, mate. 

We've got a bit of a Settings layout situation on our hands now. Do you perhaps want to add a new parent section called Layout?

**Layout:**
Layout origin
Desktop Layout
Tablet Layout
Mobile Layout

Let's rename Title to Image Title.

For the final order, let's try the following:

Preloader
Image Title
Layout (parent heading)

Misplon:
The title doesn't display for me using below because it's outside of the image container. The following uses overflow: hidden:

`.so-widget-sow-simple-masonry-default-e31791ed3871-26891 .sow-masonry-grid-item`

Screenshot 2022-10-11 at 21 20 43

Above Image works for one out of my three test images:

Page_Builder_–_SiteOrigin

What has the interest been in adding this setting to the Simple Masonry Widget?

AlexGStapleton:
> The title doesn't display for me using below because it's outside of the image container. The following uses overflow: hidden:

This should have been resolved by [the JavaScript changes](https://github.com/siteorigin/so-widgets-bundle/pull/1621/files#diff-caba28d547d27e8d80fd7987ef70bb23399f68e9b644baee484ea096e0b6e10f). Hm. Does using a build help here?

> What has the interest been in adding this setting to the Simple Masonry Widget?

The ability to add text that displays before/after the Simple Masonry has been a long-requested feature of the Masonry widget but the requests aren't frequent. The lack of a title would easily be the main reason people would use the Image Grid, or even the Image widget, over the Simple Masonry.

Misplon (approved):

PR #1623, opened 2022-10-07
Contact: Prevent Notice Before Edit

AlexGStapleton:
This PR prevents the following notice:

`( ! ) Notice: Undefined index: _sow_form_id in C:\Users\alex\Local Sites\siteorigin\app\public\wp-content\plugins\so-widgets-bundle\widgets\contact\contact.php on line 1273`

This notice occurs if you add a contact form and then save the page without opening the contact. View the page and you'll get the above error message. Opening the widget will generate `_sow_form_id` and that's required to allow for multiple contact forms on the same page without submitting them all at once. We assume this exists because it almost always does outside of this one specific instance. This is more so to account for situations where I user adds a contact form and previews it prior to editing it - likely situation for that would be while using the Live Editor.

### DIFF
--- a/widgets/contact/contact.php
+++ b/widgets/contact/contact.php
@@ -1251,7 +1251,8 @@ class SiteOrigin_Widgets_ContactForm_Widget extends SiteOrigin_Widget {
 			if ( empty( $field['type'] ) ) {
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


Misplon:
I'm unable to replicate the issue. The change looks good. Happy to proceed.

Misplon (approved):

PR #1630, opened 2022-10-12
Introduce Widget Validation: `sowbForms.validateFields` & `sow_validate_widget_data` Event

AlexGStapleton:
This PR allows for widget validation. It also introduces the `sow_validate_widget_data` event to allow other developers to validate things. It also introduces `required` support. You can set any field to required by setting `required` to `true`. Currently missing required fields are only communicated using a red border of the field but we could potentially introduce a notification similar to the data mismatch prompt we have for widgets.

Screenshot 2022-10-12 at 16-51-46 Edit Page ‹ SO — WordPress

You can test this PR by adding the following [on the line after this](https://github.com/siteorigin/so-widgets-bundle/blob/1.42.2/widgets/button/button.php#L57):

`'required' => true,
`
Please test the required functionality in the following contexts:

- SiteOrigin Page Builder (the Classic Editor, Layout Builder and SiteOrigin Layouts Block. (requires https://github.com/siteorigin/siteorigin-panels/pull/1007)
- Legacy Widgets Page & new Widgets page.
- SiteOrigin Widgets Block.
- Customizer.
- SiteOrigin Widgets Settings.

You can test the `sow_validate_widget_data` event using the following snippet:

```
jQuery( document ).on('sow_validate_widget_data', function( e, valid, form, id ) {
	if ( id == 'sow-button' ) {
		var values = sowbForms.getWidgetFormValues( form );
		if ( ! values.url.length ) {
			form.find( '[name*="url"]' ).val( 'https://siteorigin.com/' );
		}
	}

	return valid;
} );
```
This JavaScript will:
- Confirm the widget is the SiteOrigin Button widget.
- If a URL isn't set, set it to https://siteorigin.com.

### DIFF
--- a/admin/admin.js
+++ b/admin/admin.js
@@ -188,11 +188,17 @@ jQuery( function( $ ){
 
 	dialog.find( '.so-save' ).on( 'click', function( e ) {
 		e.preventDefault();
+		var $form = dialog.find( 'form' );
+
+		validSave = sowbForms.setupRequiredFields( $form )
+		if ( typeof validSave == 'boolean' && ! validSave ) {
+			return false;
+		}
 
 		var $$ = $( this );
 		$$.prop( 'disabled', true );
 
-		dialog.find( 'form' ).on( 'submit', function() {
+		$form.on( 'submit', function() {
 			$$.prop( 'disabled', false );
 			dialog.hide();
 		} ).trigger( 'submit' );
--- a/base/css/admin.less
+++ b/base/css/admin.less
@@ -103,6 +103,14 @@ div.siteorigin-widget-form {
 				color: #F02311;
 				font-style: italic;
 			}
+
+			.siteorigin-widget-input {
+				transition: 300ms ease-in-out border-color;
+
+				&.sow-required-error {
+					border-color: #F02311;
+				}
+			}
 		}
 
 		input[type="text"] {
--- a/base/js/admin.js
+++ b/base/js/admin.js
@@ -1516,6 +1516,72 @@ var sowbForms = window.sowbForms || {};
 		}
 	};
 
+	/**
+	 * Look for and valid any fields that are required.
+	 */
+	sowbForms.validateFields = function( form ) {
+		var valid = $( document ).triggerHandler(
+			'sow_validate_widget_data',
+			[
+				valid,
+				form,
+				// Widget ID.
+				form.find( '.siteorigin-widget-form' ).data( 'id-base' )
+			]
+		);
+
+		if ( typeof valid != 'boolean' || valid ) {
+			form.find( '.siteorigin-widget-field-is-required' ).each( function() {
+				var $$ = $( this );
+				var $field = $$.find( '.siteorigin-widget-input' );
+
+				if (
+					! $field.val() ||
+					(
+						$$.hasClass( 'siteorigin-widget-field-type-checkboxes' ) &&
+						! $field.prop( 'checked' )
+					)
+				) {
+					valid = false;
+					$field.addClass( 'sow-required-error' );
+				}
+
+				$( 'sow-required-error' ).one( 'change', function() {
+					$field.removeClass( 'sow-required-error' );
+				} )
+			} );
+		}
+
+		return valid;
+	}
+
+	// Validate widget added using Page Builder.
+	if ( typeof panelsOptions == 'object' ) {
+		$( document ).on( 'close_dialog_validation', function( e, values, widget, id, instance ) {
+			return sowbForms.validateFields( $( instance.el ) );
+		} );
+	}
+
+	// Validate widget added using Classic Widgets & Customizer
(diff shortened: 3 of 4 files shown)


Misplon:
Thanks, mate.

- If you don't enter a Button Title, are you able to exit the widget? I don't seem to be able to exit/close the widget.
- Should we change (Required) to an asterisk?

Misplon:
@AlexGStapleton 

Please check:
- SO Widget Block in Block Editor: `ReferenceError: validationSetup is not defined`
-  Button Widget in Block Widget Interface. Check if working, doesn't seem to for me.
- Widget Block in Block Widget Interface. Check if working, doesn't seem to for me.
- Customizer (insert Button Widget): `Uncaught TypeError: i.find is not a function`

Thanks!

AlexGStapleton:
I've added an additional snippet for `sow_validate_widget_data` testing.

Misplon:
> That's intentional. With this change, and a field being set to required, the user must enter Button Text. If they were able to close the widget, the field isn't directly required - it's just labeled as such. With that said, there may be better ways to communicate this - maybe click update, display notice, and then allow update regardless of value.

I'm not sure I'm following this functionality. What if I don't want to enter a title right now, but I'd like to go back to editing my page? The user can't escape.

Misplon:
> With that said, there may be better ways to communicate this - maybe click update, display notice, and then allow update regardless of value.

Sounds good.

> You can download it by [clicking here](https://github.com/siteorigin/so-widgets-bundle/pull/1630/commits/31bd5d65cf418632e1fb7dc79a1a5c7c44f38aa2). The only change made in build not present in this PR was setting Button Text as Required for testing purposes.

The link doesn't appear to be correct. Thanks, mate.

Misplon:
I'm trying to think of another way to handle exiting without saving with a required field not filled in. I think your suggestion is the way to go. Thanks!

AlexGStapleton:
Build link: https://drive.google.com/uc?id=1hPQziVvJMokZkLnxGW8_We1PSzz-Xq_P

> I'm not sure I'm following this functionality. What if I don't want to enter a title right now, but I'd like to go back to editing my page? The user can't escape.

I agree it's not ideal, but there really are not many options we have for actually enforcing the required nature of the field. I think what I suggested in my last comment is the way to go but I'm open to ideas. That suggest isn't exactly ideal either as it can still result in a situation where the data is missing, but at the very least, we communicated the required nature and required the user to proceed. If nothing else comes to mind, I'll get started on this.

> I'm trying to think of another way to handle exiting without saving with a required field not filled in

Existing widgets? Hm, yeah, that's tricky as enforcing required fields after the widget has been introduced is a lot. I guess I could modify the required parameter to support three options:

- True (full).
- Warn/soft (warn the user about the required nature but don't prompt them).
- False (off - default).

Thoughts?

Misplon:
I originally commented in [Slack](https://siteorigin.slack.com/archives/D0MGRNNTA/p1665995771453879). I'll comment here to clean up our Slack chat.

We could offer full validation if there were a way of exiting a widget and leaving it in draft form until the validation has been satisfied. Or if there were a way of deleting a widget and exiting if you didn't want to complete the required field. Given how things are currently set up, I don't think we can reasonably offer full validation. (What if you insert a widget but no longer want to use it? Unless you refresh the page, you must fill in the field to continue editing.)

Soft and off sound good. Let me know how you want to proceed.

AlexGStapleton:
https://github.com/siteorigin/so-widgets-bundle/pull/1630/commits/2ea12482e7b10d5a6c6dd3228ade173927bdd8b8 allows for optional error messages for required fields. This is done by setting the error message to the required value rather than true. For example:

Screenshot 2022-10-27 at 04-53-10 Premium Addons ‹ SO — WordPress

Misplon:
Thanks, mate.

- Are you encountering an error when inserting a SiteOrigin Widget Block with this branch as a build?
This block has encountered an error and cannot be previewed.
`ReferenceError: validationSetup is not defined`

Misplon:
**- Block Widget Area:**
- Legacy Widget: The SiteOrigin Button Widget doesn't have a Save button in this area.
- SiteOrigin Widget Block: Same error as the Block Editor.

AlexGStapleton:
I've updated the PR description to reference changes made after it was posted.

AlexGStapleton:
> Legacy Widget: The SiteOrigin Button Widget doesn't have a Save button in this area.

As of https://github.com/siteorigin/so-widgets-bundle/pull/1630/commits/4464afba7c7514599a7c217b3c8ce394df6cfdfa this should be resolved as it's treated the same as the Block Editor (editor specific save button rather than widget level saving).

AlexGStapleton:
Resolved. The last will require a build to test.

AlexGStapleton:
The last commit should resolve the issues with the new widget area, and the other notice.

PR #1635, opened 2022-10-14
TinyMCE Field: Prevent PHP 8 Uncaught TypeError TinyMCE Plugins

AlexGStapleton:
This PR will resolve an issue that can occur with PHP 8 if TInyMCE Plugins have been filtered and something other than array is returned.

`Fatal error: Uncaught TypeError: array_unique(): Argument #1 ($array) must be of type array, null given in /wp-content/plugins/so-widgets-bundle/base/inc/fields/tinymce.class.php:363`

### DIFF
--- a/base/inc/fields/tinymce.class.php
+++ b/base/inc/fields/tinymce.class.php
@@ -347,20 +347,23 @@ class SiteOrigin_Widget_Field_TinyMCE extends SiteOrigin_Widget_Field_Text_Input
 		);
 		
 		if ( $user_can_richedit ) {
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
@outtheboxthemes I've prepared a build for you to try. You can download it by [clicking here](https://drive.google.com/uc?id=1Kw1_06qzSwpjlhKVTc0VafnL2jlSHDKg). Once downloaded, navigate to **Plugins > Installed Plugins** and deactivate the SiteOrigin Widgets Bundle, and then delete it. Scroll to the top of the page and click **Add New**, **Upload Plugin** and upload so-widgets-bundle.1.43.1-beta.zip

Issue #1637, opened 2022-10-17
CTA: Introduce New Button Align Settings


Misplon:
@AlexGStapleton What do you think about adding two new options to the CTA Button Align setting:

Center Top
Center Bottom

Issue #457, opened 2017-04-18
Features Widget: Add responsive options for container position


AlexGStapleton:
In certain instances, the features container position may not work on mobile devices and the user may prefer having it appear on the top or bottom rather than to the left or to the right.

It might be a good idea to add the ability to specify tablet and mobile positions rather than a single option.

[Slack Reference](https://siteorigin.slack.com/archives/C02SYBBH2/p1492501796070837)

PR #1643, opened 2022-10-23
CTA: Add Center Top/Button Button Alignments

AlexGStapleton:
Resolve https://github.com/siteorigin/so-widgets-bundle/issues/1637

### DIFF
--- a/widgets/cta/cta.php
+++ b/widgets/cta/cta.php
@@ -96,7 +96,9 @@ class SiteOrigin_Widget_Cta_Widget extends SiteOrigin_Widget {
 						'label' => __( 'Button align', 'so-widgets-bundle' ),
 						'default' => 'right',
 						'options' => array(
+							'top' => __( 'Center Top', 'so-widgets-bundle' ),
 							'left' => __( 'Left', 'so-widgets-bundle' ),
+							'bottom' => __( 'Center Bottom', 'so-widgets-bundle' ),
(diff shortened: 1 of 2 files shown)


AlexGStapleton:
@Misplon What are your thoughts on adding an optional (due to the pre-existing nature of the widget) responsive collapse for the left and right alignments? We could call it... say, Mobile button layout (could also introduce a Tablet) and provide the following options:

- Standard/desktop Alignment.
- Above text.
- Below text.

I think it wouldn't be a bad idea to introduce it in this PR (unless it was to be merged really quickly) as it'll benefit from the top alignment and the LESS cleanup.

Misplon:
Thanks, mate; yeah, agreed. 

Let's re-think this slightly. Can we perhaps rename the setting to something like Alignment? How do you want to do this? Within a Responsive section?

- For Center Top and Center let's perhaps center the text.
- Do you want to remove the Mobile Align setting from the Button section? I'm not sure it does anything.

AlexGStapleton:
> Can we perhaps rename the setting to something like Alignment

We can do this, but we'll need to rename the center top/bottom settings as they don't really make sense when the setting is called Alignment.

> Within a Responsive section?

I opted to add a single setting below the alignment setting. If we were to introduce other responsive settings a dedicated section may be worth it but with only a single one it's hard to justify.

AlexGStapleton:
This PR could resolve https://github.com/siteorigin/so-widgets-bundle/issues/457 in its current state.

PR #1649, opened 2022-10-31
Blog: Allow Excerpt Length to Be Set to 0, and Hide Indicator If It Is

AlexGStapleton:
Prior to this PR it wasn't possible to set the Excerpt Length to 0 without using a filter and even if you did, you would have to remove the Indicator using an additional filter. This PR allows Excerpt Length to be set to 0 and removes the Indicator.


Misplon (approved):

Misplon:
Looks great, thanks.

PR #1652, opened 2022-11-11
Simple Image Masonry: Use Pure CSS (`object-fit`) for Sizing Images & Resolve Image Title Sizing Issue

AlexGStapleton:
This PR moves away from using JavaScript to size the image dimensions and uses pure CSS instead. [object-fit](https://caniuse.com/#search=object-fit) is widely supported at this point with 98.63% global support so it's very safe to use. This change should reduce the potential for future CLS flagging due to measurement changes, make the widget lighter weight overall (fewer repaints), and it also resolves a sizing issue when titles are used.

### DIFF
--- a/widgets/simple-masonry/js/simple-masonry.js
+++ b/widgets/simple-masonry/js/simple-masonry.js
@@ -53,25 +53,7 @@ jQuery( function ( $ ) {
 							}
 						}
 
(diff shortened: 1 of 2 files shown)


Misplon:
Nicely done, mate.

Misplon (approved):

PR #1660, opened 2022-11-30
Add min/max for number input

ictbeheer:


base/inc/fields/number.class.php
```diff
@@ -44,10 +60,22 @@ protected function get_input_classes() {
...
+
+		if ( ! empty( $this->min ) ) {
+			$value = max( $value, $this->min );
+		}
+
+		if ( ! empty( $this->max ) ) {
+			$value = min( $value, $this->max );
+		}
+
+		if ( ! empty( $this->abs ) ) {
```
ictbeheer, line 76: Maybe we can remove the abs property? It is not documented on https://siteorigin.com/docs/widgets-bundle/form-building/form-fields/ and to ensure the number is positive you can now use min = 0

AlexGStapleton: Due to backwards compatibility reasons and expectations, this isn't possible - it has to be retained.

base/inc/fields/number.class.php
```diff
@@ -28,13 +45,12 @@ protected function get_default_options() {
 	}
 
 	protected function get_input_attributes() {
-		if ( empty( $this->step ) ) {
-			return array();
-		}
-	
-		return array(
+		$input_attributes = [
```
AlexGStapleton, line 48: Can you please switch this to a "long" array? For newly introduced code we try to maintain WordPress PHP Coding Standards.

https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#declaring-arrays

AlexGStapleton (changes requested):
Thank you for submitting another PR! :)
Just a formatting change and this PR should be good to merge.

AlexGStapleton (approved):

PR #1688, opened 2023-01-18
Fix fatal error mce_external_plugins missing editor_id

joelataccelm:
if site origin hooks to mce_external_plugins before a plugin written to expect the second param $editor_id, it will cause a fatal error.


AlexGStapleton (approved):
Hi @joelataccelm,

Thank you for the PR. This is a great catch and fix, and we really appreciate the PR. :)

Kind regards,
Alex

PR #1841, opened 2023-11-22
Beaver Builder: Switch Color Fields to a Color Input

AlexGStapleton:
This PR will switch WB Color Form Fields to the [color input type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color). This will allow for those fields to be more useable overall.

### DIFF
--- a/base/js/admin.js
+++ b/base/js/admin.js
@@ -264,9 +264,18 @@ var sowbForms = window.sowbForms || {};
 			// For any repeater items currently in existence
 			$el.find('.siteorigin-widget-field-repeater-item').sowSetupRepeaterItems();
 
-			// Set up any color fields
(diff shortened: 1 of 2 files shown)


Misplon:
Superseded by #2353, which reimplements this fresh on current `develop` (this branch had gone stale/conflicting). The new version also guards the conversion so it only applies to full `#rrggbb` values — empty, shorthand and `rgba()` alpha values stay as text inputs so Beaver Builder can't overwrite them on save. Thanks Alex.

Misplon:
Merged as #2353. Reimplemented fresh on develop: in Beaver Builder (where wpColorPicker isn't loaded) SiteOrigin color fields now fall back to a native color input. It's guarded to full #rrggbb values, so empty, shorthand and rgba() alpha values stay as text inputs and aren't coerced to #000000 or overwritten on save. Thanks for the original idea, Alex.

Issue #1975, opened 2024-05-14
PHP Fatal error:  Uncaught TypeError: stripslashes(): Argument #1 ($string) must be of type string, array given in /base/inc/post-selector.php:71


JoelLisenby:
I noticed the following error, when running PHP 8.x on the following line: https://github.com/siteorigin/so-widgets-bundle/blob/a795ebc454593be7169a54e7e8420fb337208554/base/inc/post-selector.php#L71

`PHP Fatal error:  Uncaught TypeError: stripslashes(): Argument #1 ($string) must be of type string, array given in /wp-content/plugins/so-widgets-bundle/base/inc/post-selector.php:71`

You can see on the following line 'date_query_relative' an array, not a string.

https://github.com/siteorigin/so-widgets-bundle/blob/a795ebc454593be7169a54e7e8420fb337208554/base/inc/fields/posts.class.php#L88

JoelLisenby:
@AlexGStapleton thank you for the quick response. I noticed the error was still occurring and made a comment here with a small adjustment to the fix. https://github.com/siteorigin/so-widgets-bundle/pull/1976#issuecomment-2111124603

PR #1976, opened 2024-05-14
Post Selector: Prevent Potental Fatal

AlexGStapleton:
Resolves https://github.com/siteorigin/so-widgets-bundle/issues/1975

`date_query_relative` can be a (json encoded) `string`, or an `array`. This PR allows for both.


JoelLisenby:
Hi @AlexGStapleton, just wanted to mention on line 75 now I think you may have meant to go with this

`if ( ! is_array( $query['date_query_relative'] ) ) {`

instead of this

`if ( ! is_array( stripslashes( $query['date_query_relative'] ) ) ) {`

PR #2264, opened 2025-10-17
Fix bottom-align spacing in Features widget

Misplon:
## Summary
- keep icon/title/text anchored to the top while only `.sow-more-text` drops to the bottom when Bottom Align is enabled
- restore spacing/flex behaviour for right/left icon layouts so icons retain margin and headings stay aligned
- CSS-only changes in `widgets/features/styles/default.less`

## Testing
- manual: visit `http://localhost/siteorigin/features-block-test/`. find the file included below.
  - verify rows: **Top / bottom-align on / more yes**, **Right / bottom-align off / more yes**, **Right / bottom-align on / more yes**, **Right / bottom-align on / more no**, **Bottom / bottom-align on / more yes**
  - confirm only the More paragraph sits at the bottom, icons have consistent spacing, and taller columns stay top-aligned

[features-block-test.html](https://github.com/user-attachments/files/22967963/features-block-test.html)

### DIFF
--- a/widgets/features/styles/default.less
+++ b/widgets/features/styles/default.less
@@ -184,10 +184,6 @@
 			width: @container_size;
 			flex: 0 0 @container_size;
 
-			& when ( @more_text_bottom_align = true ) {
-				/* handled per-position to avoid collapsing horizontal spacing */
-			}
-
 			&:not(.sow-container-none) [class^="sow-icon-"],
 			.sow-icon-image {
 				align-items: center;


widgets/features/styles/default.less
```diff
@@ -134,6 +166,18 @@
 			}
 		}
 
+		.sow-features-feature-right-left-container {
+			display: flex;
+			flex-direction: inherit;
+			gap: 12px;
```
AlexGStapleton: Is this gap value intentional? This will result in extra spacing. Compare "Top / bottom-align off / more yes" before and after.

widgets/features/styles/default.less
```diff
@@ -142,7 +186,7 @@
 			flex: 0 0 @container_size;
 
 			& when ( @more_text_bottom_align = true ) {
-				margin: auto;
+				/* handled per-position to avoid collapsing horizontal spacing */
 			}
```
AlexGStapleton: I think this should be removed.

AlexGStapleton (changes requested):
There's an overflow present in the "Right / bottom-align on / more yes" row. You can verify this by checking `.sow-features-feature` box modal using your browser's inspector and you'll notice the text overflow.

Misplon:
> There's an overflow present in the "Right / bottom-align on / more yes" row. You can verify this by checking .sow-features-feature box modal using your browser's inspector and you'll notice the text overflow.

I'm not seeing an overflow:
Screenshot 2025-10-26 at 11 37 26

widgets/features/styles/default.less
Misplon: Removing.

AlexGStapleton:
Just tried again, and it's not happening for me now with a new build but the same page. Unclear what the difference is. I'm happy to proceed with a merge if you are.

PR #2265, opened 2025-10-20
Backups: Ensure timestamp is set when saving widgets.

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/1294

This PR ensures the timestamp is added when updating a widget.

I'm unsure how the timestamp was previously being set. I had a look at a version from three years ago and couldn't find it being set anywhere.

### DIFF
--- a/base/siteorigin-widget.class.php
+++ b/base/siteorigin-widget.class.php
@@ -840,17 +840,7 @@ abstract class SiteOrigin_Widget extends WP_Widget {
 		}
 
 		if ( $new_instance !== $old_instance ) {
-			$current_timestamp = isset( $new_instance['_sow_form_timestamp'] ) ? $new_instance['_sow_form_timestamp'] : null;
-			$previous_timestamp = isset( $old_instance['_sow_form_timestamp'] ) ? $old_instance['_sow_form_timestamp'] : null;
-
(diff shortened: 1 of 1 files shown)


Misplon:
Thanks. Below is feedback from Codex:

base/siteorigin-widget.class.php:842 casts the millisecond timestamp from microtime(true) to int. On 32-bit PHP builds this overflows (values > 2147483647), so the stored _sow_form_timestamp collapses to that max value. That breaks the “newer version” detection in base/js/admin.js because the saved instance can never have a greater timestamp than the session backup, leaving stale backups in place. Consider keeping the value as a string, or dropping the cast and relying on PHP’s automatic type handling so 32-bit installs continue to see monotonically increasing timestamps.

Misplon:
@AlexGStapleton Please check the above. Codex:

> I noticed that the branch’s earlier change always overwrote _sow_form_timestamp with a server-side microtime, which meant the saved timestamp could   easily be lower than the value stored in the editor’s session. The field-backup JavaScript treats a lower saved timestamp as “there’s a newer version to restore,” so after every save the restore notice would reappear. My fix keeps the client-provided timestamp when it moves forward and only generates a server timestamp when the submitted value is missing or older than what we already had. This preserves the current backup workflow while still guarding against stale or invalid timestamps.

Issue #2270, opened 2025-11-05
Missing 2nd paramter in "tiny_mce_plugins" filter


sanzeeb3:
"tiny_mce_plugins" is a WP Core filter which accepts 2 parameters: https://developer.wordpress.org/reference/hooks/tiny_mce_plugins/

But the plugin overrides the filter with only one parameter

https://github.com/siteorigin/so-widgets-bundle/blob/cdd2de0e3242926993809459f28fb46eff5857ac/base/inc/fields/tinymce.class.php#L361

This causes fatal error with any plugins using the "tiny_mce_plugins" with 2 parameters.

PR #2271, opened 2025-11-06
TinyMCE: Resolve Fatal related to `tiny_mce_plugins` Filter

AlexGStapleton:
Resolved https://github.com/siteorigin/so-widgets-bundle/issues/2270

We now pass the element id to identify the usage.


Misplon:
Thanks, please can you check that you can select the content area in the Editor Block.

Misplon:
The issue now seems to be resolved on my end, but perhaps do a quick check.

Misplon (approved):

PR #2274, opened 2025-12-07
Fix Hero repeater media field after duplicate

Misplon:
## Summary
- ensure cloned repeater items re-run field setup by clearing data-initialized flags
- fixes media field in duplicated Hero frames not opening background picker

## Testing
- not run (admin JS only)


Misplon:
To replicate the issue, add a Hero frame with a background image. Duplicate the frame and then try open the Background image in the duplicated frame.

PR #2276, opened 2025-12-16
Blocks: Resolve apiVersion browser warning

AlexGStapleton:
This PR resolves the following warning that'll be output for all of all our blocks.

The block "sowb/widget-block" is registered with API version 2 or lower. This means that the post editor may work as a non-iframe editor. Since all editors are planned to work as iframes in the future, set the `apiVersion` field to 3 and test the block inside the iframe editor. See: https://developer.wordpress.org/block-editor/reference-guides/block-api/block-api-versions/#version-3-wordpress-6-3

### DIFF
--- a/compat/block-editor/widget-block.js
+++ b/compat/block-editor/widget-block.js
@@ -27,9 +27,6 @@
 		return errorMessage;
 	};
 
-	// Track the active AJAX request for generating widget previews.
-	let sowbGenerateWidgetPreviewActiveRequest = null;
-
 	// Certain widgets are excluded from the content check as
 	// they don't contain "standard" content indicators.
 	const widgetsExcludedFromContentCheck = [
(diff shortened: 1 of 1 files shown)


Misplon:
Thanks. The SiteOrigin Block toolbar seems to be missing when this branch is in use.

AlexGStapleton:
That issue should now be fixed.

This fix was a little more complicated than the outline suggested, as you can't make use of useBlockProps with how the Block was previously set up. When testing this issue, please ensure the preview and widget form both update as expected.

Misplon:
Thanks.

**Findings**

- Medium: compat/block-editor/widget-block.js:280 The debounced preview now calls sowbGenerateWidgetPreview with loadingWidgetPreview forced to false, bypassing the in‑flight guard. Rapid form changes can trigger overlapping preview requests, and out‑of‑order responses can overwrite the latest preview. Consider passing the current loading flag or tracking a request token to ignore stale responses.

**Questions / Assumptions**

- Is concurrent preview fetching acceptable here, or should we keep a single in‑flight request to avoid stale previews?

**Testing Gaps**

- No automated coverage in the PR; at minimum, manual test rapid widget changes in both post editor and Site Editor iframe to ensure previews stay in sync.

**Change summary**

- Bumps block registrations to apiVersion: 3, refactors widget block edit UI into a functional component with useBlockProps, adds iframe safety checks, uses window.top.ajaxurl for iframe AJAX, and guards missing widget data in PHP.

AlexGStapleton:
Accounted for by https://github.com/siteorigin/so-widgets-bundle/pull/2276/commits/3c31fa1e8dc5e08f75dc0c8edfbc035abe7643f3

Misplon:
Thanks. Are you ok with the scoping as is?

**Potential follow‑up**

- The active request tracker is module‑scoped, so simultaneous edits of two widget blocks could cancel each other’s previews. If that’s a concern, the request should be tracked per block instance (e.g., a useRef tied to clientId).

Misplon:
The previews don't seem to load here with a built version of the latest commit: https://demo.siteorigin.com/corp/wp-admin/post.php?post=5699&action=edit.

AlexGStapleton:
> The active request tracker is module‑scoped, so simultaneous edits of two widget blocks could cancel each other’s previews. If that’s a concern, the request should be tracked per block instance (e.g., a useRef tied to clientId).

Done.

> The previews don't seem to load here with a built version of the latest commit

This issue was caused by the active request tracker cancelling some of the preview attempts. This issue was also resolved by https://github.com/siteorigin/so-widgets-bundle/pull/2276/commits/bee8b6de22618de87d8d2e408a0b44d00bd30993. Let me know if it's still occurring for you.

PR #2292, opened 2026-03-06
Add Filter to Show Cover Image When Self-Hosted Video Ends

Misplon:
## Summary
- Adds a `sow_video_show_cover_on_end` filter (defaults to `false`) that, when enabled, resets a self-hosted video back to its poster/cover image after playback ends instead of freezing on the last frame.
- Users can enable it with `add_filter( 'sow_video_show_cover_on_end', '__return_true' );`

## How it works
- The filter adds a `data-show-cover-on-end` attribute to the `<video>` element via the existing `$video_args` array.
- The JS listens for the `ended` event and calls `video.load()` to reset the player to its initial state, re-displaying the poster image.

## Test plan
- [x] Add a self-hosted video with a cover image, confirm default behavior is unchanged (freezes on last frame)
- [x] Enable the filter and confirm the video resets to the cover image after playback ends
- [x] Confirm it works with both MediaElement.js controls and hidden controls
- [x] Confirm loop still takes priority (looped videos don't fire `ended`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/widgets/video/tpl/default.php
+++ b/widgets/video/tpl/default.php
@@ -48,6 +48,10 @@ if ( ! $hide_controls ) {
 	$video_args['controls'] = '';
 }
 
+if ( apply_filters( 'sow_video_show_cover_on_end', false, $instance ) ) {
+	$video_args['data-show-cover-on-end'] = 'true';
+}
+
 $so_video = new SiteOrigin_Video();
 
 do_action( 'siteorigin_widgets_sow-video_before_video', $instance );
--- a/widgets/video/tpl/default.php
+++ b/widgets/video/tpl/default.php
@@ -11,6 +11,7 @@
  * @var $video_type
  * @var $fitvids
  * @var $hide_controls
+ * @var $show_cover_on_end
  */
 if ( ! empty( $instance['title'] ) ) {
 	echo $args['before_title'] . wp_kses_post( $instance['title'] ) . $args['after_title'];
@@ -48,7 +49,7 @@ if ( ! $hide_controls ) {
 	$video_args['controls'] = '';
 }
 
-if ( apply_filters( 'sow_video_show_cover_on_end', false, $instance ) ) {
+if ( $show_cover_on_end ) {
 	$video_args['data-show-cover-on-end'] = 'true';
 }
 
--- a/widgets/video/js/so-video-widget.js
+++ b/widgets/video/js/so-video-widget.js
@@ -4,9 +4,11 @@ var sowb = window.sowb || {};
 
 jQuery( function ( $ ) {
 	sowb.setupVideoPlayers = () => {
-		const $video = $( 'video.sow-video-widget' );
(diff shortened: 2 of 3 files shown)


AlexGStapleton:
I'm having trouble getting both widgets/video/js/so-video-widget.js to output ([it has a specific set of requirements](https://github.com/siteorigin/so-widgets-bundle/blob/74a5942bf00e955147cc381361cb81087e264eea/widgets/video/video.php#L150-L179)) and the `show-cover-on-end` attribute present. What is the required settings combination, and is that combination a little too strict? It might be worth altering the so-video-widget.js changes to allow for it to be output with more settings combinations (will require JS adjustments to prevent issues).

Misplon:
@AlexGStapleton

You were right. The requirements were too strict.

Before this change, the filter would only work when `so-video-widget.js` was already being loaded, which for self-hosted videos effectively meant a fairly narrow combination of settings such as hidden controls or FitVids being enabled.

I have loosened that now in `e5845b2e`. If `sow_video_show_cover_on_end` returns `true` for a self-hosted video, we also enqueue `so-video-widget.js`, so the poster reset behavior is no longer tied to that smaller set of settings.

I also adjusted the JS initialization so repeated `setup_widgets` calls only initialize new video elements and do not reinitialize existing ones.

That should make the filter work with the normal self-hosted video setup as well.

widgets/video/tpl/default.php
```diff
@@ -48,6 +48,10 @@
 	$video_args['controls'] = '';
 }
 
+if ( apply_filters( 'sow_video_show_cover_on_end', false, $instance ) ) {
```
AlexGStapleton: The value returned by apply_filters should ideally be typecast as a bool. 

No need to do further validation beyond this. This is just to ensure a valid value is returned, and the contents are irrelevant).

AlexGStapleton (changes requested):
Thanks for changing that. I don't see any direct issues with those changes and the cover now appears for me after the video finishes. Strangely, the image displays differently after playing. Are you getting this? This happens with every image I've tested so I don't think it's image specific.

Before play:

Image

After play:

Image

Misplon:
@AlexGStapleton I've pushed follow-up changes in 5d12b65a.

For the inline review point, `sow_video_show_cover_on_end` is now normalized through a dedicated boolean helper and that value is reused for both the JS enqueue decision and the template attribute output.

For the poster rendering issue, I removed the generic `ended -> load()` reset for MediaElement-backed videos and switched that path to MediaElement's built-in `showPosterWhenEnded` handling instead. The native `load()` fallback is still only used when MediaElement isn't in play.

I haven't done a browser pass yet on my side, so if you get a chance to retest the controls-enabled self-hosted case, that would be useful.

AlexGStapleton (approved):
Browser pass looks good. Cover is now the correct size after the video finishes.

PR #2303, opened 2026-04-07
Fix double border on TinyMCE editor in Panels dialog

### LINKED ISSUE #2302: Hero TinyMCE: Potential Double Border
Unsure of cause. Issue visible on packs.siteorigin.com.

WP: Version 6.9.4

Image

Misplon:
## Summary
- Fixes a double border around the TinyMCE editor textarea in the Panels widget dialog.
- The textarea was getting a border from Panels' admin CSS (`.so-panels-dialog .so-content textarea`), while `.wp-editor-container` already has its own border from WordPress core's `editor.min.css`. Both were visible as nested borders.
- Removes the textarea border when inside a `wp-editor-container`, letting the container border serve as the single border.

Fixes #2302

## Test plan
- [ ] Open a widget with a TinyMCE field (e.g., Hero Widget) in the Panels dialog
- [ ] Confirm only a single border appears around the editor in Code/Text mode
- [ ] Switch to Visual mode and confirm no double border
- [ ] Test in the block editor widget form to ensure no regression

### DIFF
--- a/base/inc/fields/css/tinymce-field.less
+++ b/base/inc/fields/css/tinymce-field.less
@@ -6,3 +6,7 @@
 #mceu_44-body {
 	z-index: 1000000 !important;
 }
+
+.siteorigin-widget-tinymce-container .wp-editor-container textarea.wp-editor-area {
+	border: none;
+}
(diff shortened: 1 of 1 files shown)


Misplon:
I'm unsure why but the issue was only visible to me on packs.siteorigin.com.

PR #2306, opened 2026-04-30
Block Editor: Locate widget form inside post-editor iframe + clone iris/color-picker deps

Misplon:
## Summary
- **Fixes the real bug:** SOWB widget forms (Features Widget repeater, settings sections, color pickers) wouldn't open in the WP 6.5+ post editor because the post editor renders blocks inside `<iframe name="editor-canvas">`, but `sowbGetBlockForm()` only knew about the Site Editor canvas selector. Form lookup returned empty, `$mainForm.sowSetupForm()` was never called, and no click handlers were ever bound.
- **Robustness extras:** Adds `#jquery-touch-punch-js`, `#iris-js`, and `#wp-color-picker-js` to the iframe-clone allowlist so the asset chain is complete. Verified via in-page diagnostic that these were already present via WP's bridge — but having them in the explicit allowlist closes a real gap and is the same shape as existing entries.

## Root cause (verified with runtime diagnostic, not guessed)
Inline debug emitted into both contexts on WP 6.9.4 confirmed:
- IFRAME (`name=editor-canvas`): `[data-block]` and `.siteorigin-widget-form-main` exist; `data(sow-form-setup)=undefined`; all jQuery / jQuery UI / wp-color-picker globals already present.
- PARENT: `[data-block]` count is 0 — block tree lives only in the iframe.
- `sowbGetBlockForm()` queried the parent doc, found nothing, and the form-init bailed out.

The fix re-resolves the editor canvas on every call (Site Editor selector → post editor iframe selector → parent-doc fallback) and uses `.contents()` to descend into whichever iframe is mounted. Caching is preserved but only written when the canvas is actually present, so we never lock in an empty result before the iframe mounts.

## Test plan
- [ ] WP 6.9.4 post editor: Features Widget repeater item header expands on click.
- [ ] WP 6.9.4 post editor: settings section header expands on click.
- [ ] WP 6.9.4 post editor: drag-to-reorder works on repeater handles.
- [ ] WP 6.9.4 post editor: color-picker fields render with alpha slider.
- [ ] WP 7.0 post editor: same four checks.
- [ ] WP 7.0 Site Editor: same four checks (Site Editor branch is unchanged in shape).
- [ ] `widgets.php` Block Widgets screen (no iframe): widget forms still open and behave (parent-doc fallback branch).
- [ ] No new console errors related to `[data-block]` or jQuery scope.


Misplon:
Superseded by #2308, which opens against a fresh branch off develop with four atomic commits in plan order. The previous branch had a wall-knock revert pair from a misdirected fix attempt; the diagnostic-confirmed work has been restated cleanly in #2308.

PR #2308, opened 2026-04-30
Block Editor: Fix iframe widget field lifecycle and repeater TinyMCE

### LINKED ISSUE #2307: Block Editor: TinyMCE editor field renders only in first repeater item inside post-editor iframe
## Context

After landing the iframe-asset-clone fixes (PR pending: post-editor iframe form lookup, `#utils-js`/`#editor-js`/`#quicktags-js`/`#iris-js`/`#wp-color-picker-js`/`#jquery-touch-punch-js`/`#editor-js-extra`/`#quicktags-js-extra` added to `sowbCanvasCloneElements`, clone-source-document fix, post-success latch), the SOWB widget admin form works inside the WP 6.5+ post-editor `<iframe name="editor-canvas">` — repeaters expand, drag-to-reorder works, color pickers render. But the TinyMCE editor field (Features Widget text field, Editor widget, Headline sub-headline) only renders fully for the **first** opened repeater item. Items 2..N show a bare `<textarea>` with no toolbar even though their TinyMCE engines report identical "ready" state.

## Reproduction

1. WP 6.9.4 or WP 7.0 post editor (the canvas iframe is a `blob:` URL document).
2. Insert a Features Widget block.
3. Expand the first feature item — TinyMCE renders correctly with toolbar, content area, Visual/Text tabs.
4. Expand a second feature item — bare textarea, no toolbar, no `.mce-tinymce` wrapper visible.

## Known From Diagnostics

Live in-page instrumentation captured the following on WP 6.9.4 post editor:

- `wp.editor.initialize( id, settings )` succeeds for **every** editor in the iframe. `tinymce.editors` accumulates entries `[id-1, id-2, id-3, …]` correctly.
- Per-editor lifecycle events fire in the expected order for every editor: `AddEditor → ScriptsLoaded → BeforeRenderUI → PreInit → PostRender → LoadContent → init COMPLETE → SkinLoaded`. No `LoadError`. No `window.error`. No `unhandledrejection`.
- Engine-side watchdog at +500ms / +1500ms / +3000ms reports identical state for every editor: `initialized=true`, `hidden=false`, `iframeElement=true`, `contentAreaContainer.iframe=true`.
- The first repeater item's field DOM contains `.mce-tinymce=1`, `.wp-editor-wrap=1`, `.mce-toolbar-grp=1`. Parent chain of `editorContainer` ends correctly inside `siteorigin-widget-tinymce-container > .siteorigin-widget-field-type-tinymce > .siteorigin-widget-field-repeater-item-form`.
- For items 2..N: `.mce-tinymce=0`, `.wp-editor-wrap=0`. `setupTinyMCEFieldInitializer()` is called on those fields via the `sowbBlockFormInit` postMessage handler, but the existing short-circuit at `base/inc/fields/js/tinymce-field.js:522` (`else if ( $field.attr( 'data-pre-init' ) ) { return; }`) refuses to rebind the `.one('sowsetupformfield', …)` listener.
- The reason that guard fires: when React re-renders the form via `dangerouslySetInnerHTML` (`compat/block-editor/widget-block.js:551`), the field's DOM node is replaced. The new node carries `data-pre-init="true"` in its serialised HTML, but it has no jQuery `.one()` listener attached. The guard prevents rebinding, so the field is silently parked waiting for an event that nothing dispatches against it.
- TinyMCE's `LoadContent` event was observed firing many times (~6 times in 13 seconds) for a single editor, suggesting `editor.load()` or `editor.setContent()` is being called in a loop, likely from the `change → setAttributes → re-render → setWidgetFormValues → trigger('change')` cycle in `WidgetBlockEdit`.
- `sowbBlockFormInit` postMessage is dispatched ~25 times per session because `WidgetBlockEdit`'s `.then()` callback fires on every render. The receiver iterates every field and calls `setupTinyMCEFieldInitializer`. Each pass either short-circuits (Defect A above) or hits the visible-branch fast path; under the storm, listener state on each field is hard to reason about.

## Attempted Fix and Why It's Deferred

A direct one-line fix to the guard (switch from `$field.attr('data-pre-init')` to `$field.data('sowb-pre-init-bound')`) was tried in the working tree. With the in-memory flag, the first attempt worked for items 2–4 (4 editors mounted), but a subsequent reload showed **zero** editors mounting. The interaction between (a) React's re-render storm, (b) postMessage-driven re-iteration of `setupTinyMCEFieldInitializer`, (c) the `.one()`-vs-`.data()`-vs-attribute lifecycle, and (d) the field's `:visible` transition timing is more subtle than a single-line guard swap can fix reliably without further investigation.

## Unknowns

- Exact mechanism by which TinyMCE mounts visibly for item 1 but not for items 2..N, given identical engine-side state.
- Whether the `sowbBlockFormInit` postMessage storm is causally relevant or a symptom.
- Whether throttling `initializeFormFieldsInIframe()` to once-per-form-mount would change behaviour.
- Whether the visible-but-broken case has a CSS / stacking-context / layout cause that engine-state probes can't see.

## Recommended Next Steps

1. Reproduce on a current `develop` after the asset-clone PR merges.
2. Reinstall the diagnostic harness (engine event hooks, per-editor watchdog, parent-chain dump, 2-second snap probe, `.one()` bind/fire trace).
3. Capture a clean trace of items 2..N opening, focused on the moment `setupTinyMCEFieldInitializer` is called for an item-N field while it's `:visible=true preInit=true init=false mce=0`. The branch decision at that moment is the next concrete diagnostic question.
4. If the `.data()` flag approach is revisited, pair it with throttling `initializeFormFieldsInIframe()` to once-per-mount, so the rebind logic is exercised under a less chaotic event flow.
5. Independently consider whether `editor.load()` looping (the repeated `LoadContent` events) is destroying the visible UI of items 2..N after they mount.

## Affected Files

- `base/inc/fields/js/tinymce-field.js` — `setupTinyMCEFieldInitializer` (lines ~510–533), `setupTinyMCEField` (~lines 57–230), the `sowbBlockFormInit` postMessage receiver (~line 326).
- `compat/block-editor/widget-block.js` — `initializeFormFieldsInIframe` (~line 322), the `WidgetBlockEdit` re-render `.then()` callback (~lines 551–565).

## Environment

- WP 6.9.4 and WP 7.0 (both reproduce identically).
- Browser: Chromium-based.
- so-widgets-bundle: develop + the asset-clone PR.

Misplon:
## Summary
Fixes legacy SiteOrigin widget forms inside WordPress iframe editors by restoring the iframe asset/bootstrap path and hardening field initialization after the form DOM and cloned scripts mount asynchronously.

## What Changed
- [x] Detect the WP 6.5+ post-editor iframe (`iframe[name="editor-canvas"]`) as well as the Site Editor canvas when resolving widget forms.
- [x] Clone the required WordPress/SOWB scripts, styles, and templates into the editor iframe from the correct parent document.
- [x] Avoid latching failed iframe asset clone attempts so setup can retry after the iframe body exists.
- [x] Keep keyed form initialization and send staged iframe `sowbBlockFormInit` messages after form mount.
- [x] Remove repeater-wide iframe init broadcasts from add/open/copy flows and use targeted `sowsetupformfield` triggers for visible repeater fields.
- [x] Persist copied TinyMCE repeater editor IDs to real `data-tinymce-id` attributes so IDs survive cloning/replacement.
- [x] Reinitialize TinyMCE fields when the field says initialized but editor chrome is missing, and use safer editor teardown fallbacks.
- [x] Observe iframe form mutations for late TinyMCE and Date Range fields, waiting for iframe dependencies such as `jQuery` and `Pikaday` when needed.
- [x] Stabilize Icon picker selection coverage in the iframe.
- [x] Add Site Editor E2E coverage for Icon selection and repeater TinyMCE chrome in Hero/Features widgets.

## Why
Iframe editors can mount widget form HTML and cloned admin scripts out of order. That left some field scripts present in the iframe but not actually initialized, and TinyMCE repeater editors after the first item could render as bare textareas because copied/replaced editors only kept transient jQuery data for their IDs.

This PR covers both layers of the fix: the original iframe asset delivery problem and the follow-up repeater field lifecycle problem from #2307.

## Validation
- [x] `node --check base/inc/fields/js/date-range-field.js`
- [x] `node --check compat/block-editor/widget-block.js`
- [x] `node --check tests/e2e/wb-form-field-site-editor.test.js`
- [x] `npx playwright test tests/e2e/wb-form-field-site-editor.test.js -g "Test the Blog widget" --project="Google Chrome" --retries=0`
- [x] `npm run test:e2e` (`9 passed`)

## Reviewer Checklist
- [ ] WP 6.9/7.0 Site Editor: Icon widget initializes and stores the selected icon value.
- [ ] WP 6.9/7.0 Site Editor: Blog widget Date Range picker opens and stores a selected date.
- [ ] WP 6.9/7.0 Site Editor: Editor widget TinyMCE media flow still inserts an image.
- [ ] WP 6.9/7.0 Site Editor: Hero and Features repeaters render full TinyMCE editor chrome for at least three items.
- [ ] WP 6.9/7.0 post editor iframe: widget forms initialize, repeaters expand/reorder, and color/icon/media fields still work.
- [ ] `widgets.php` Block Widgets screen: repeater and field initialization behavior regresses neither iframe nor non-iframe screens.
- [ ] Browser console shows no `LoadError`, `window.error`, `unhandledrejection`, or `sowbBlockFormInit` message storm during normal add/open/copy flows.

## Notes
- Fixes #2307.
- Supersedes the earlier #2306 branch history.
- This intentionally updates existing PR #2308 because this branch already had an open PR against `develop`.
- The final full suite was run against a manually started Playground server using `--mount-dir-before-install`; `npm run tests` still calls the removed camelCase Playground CLI flag `--mountDirBeforeInstall` through `siteorigin-tests-common/playground/startPlayground.js`.


AlexGStapleton:
I came across a number of issues when testing with this PR during testing and when investigating https://github.com/siteorigin/so-widgets-bundle/issues/2310. I've submitted a follow up PR https://github.com/siteorigin/so-widgets-bundle/pull/2311 that resolves a bulk of the issues I came across related to this PR.

AlexGStapleton:
Manually merged into https://github.com/siteorigin/so-widgets-bundle/pull/2313

Issue #2310, opened 2026-05-10
Blocked: direct WB iframe save bridge branch state and cleanup report


Misplon:
## Summary

This issue logs the current state of the pushed branch after the direct Widgets Bundle block iframe save-bridge work became unsafe to continue inside the same implementation slice.

Branch: [`feature/wb-direct-block-save-bridge`](https://github.com/siteorigin/so-widgets-bundle/tree/feature/wb-direct-block-save-bridge)  
Current pushed HEAD: `e89760d8` (`Cleanup: Remove unproven Features regression path`)  
Base for review: `develop`

The high-level save bridge idea may still be useful, but the branch should be treated as blocked until it is reviewed or split back to a smaller, proven core change.

## Why This Branch Exists

The original problem is that direct Widgets Bundle blocks in the iframed post editor can have current values in the iframe form while serialized block `widgetData` remains stale at save time.

The core research finding was that WordPress computes edited post content before `editor.preSavePost`, so updating block attributes alone during that hook is too late. A save-time bridge needs to snapshot the mounted iframe form, rewrite `edits.content` from a cloned block tree, and then catch the block editor store up.

## Current Pushed Diff

Against `develop`, current HEAD changes five files:

```text
base/inc/fields/js/tinymce-field.js          |  21 +-
base/js/admin.js                             | 126 +++++++++
compat/block-editor/widget-block.js          | 366 +++++++++++++++++++++---
compat/block-editor/widget-block.php         |  26 +-
tests/e2e/wb-form-field-block-editor.test.js | 405 +++++++++++++++++++++++++++
```

The current branch has 901 insertions and 43 deletions against `develop`.

## Current Commit-State Correction

An earlier local report said the later churn from `2561b10b` onward had netted out to test-only changes. That was true at one intermediate HEAD, but it is not the most useful framing now.

Current HEAD includes cleanup commits:

```text
e89760d8 Cleanup: Remove unproven Features regression path
0d397c06 Revert "Step 2: Initialize iframe TinyMCE for nested fields"
01b4a463 Revert "Step 1: Fix repeater TinyMCE initialization timing"
```

So the latest unproven Features-specific source attempts were reverted or removed. However, the branch still contains shared infrastructure changes in `base/inc/fields/js/tinymce-field.js` and `base/js/admin.js`, plus the block-editor save bridge.

### `compat/block-editor/widget-block.js`

This is the main plausible core fix area.

Known changes include:

[...]

PR #2311, opened 2026-05-11
Feature/sowb block editor iframe asset clone no polling

AlexGStapleton:
This PR amends https://github.com/siteorigin/so-widgets-bundle/pull/2308 to resolve a number of performance and stability issues I was facing during testing. It also works towards resolving a portion of https://github.com/siteorigin/so-widgets-bundle/issues/2310 - I'm working on an additional focused PR for that and progress has been going well.

## What and why
Direct SiteOrigin widget blocks in the iframed block editor (Site Editor, post editor canvas) could leave TinyMCE fields in an unreachable half-initialized state, waste CPU on staggered timeout polling, and accidentally attempt setup on repeater placeholder rows. This PR replaces all of those paths with targeted, event-driven initialization and adds several layers of defensive hardening.

## Performance

- Eliminates staggered timeout polling. The previous initializeFormFieldsInIframe fired 5 postMessage calls at [0, 250, 1000, 3000, 6000] ms regardless of whether fields were ready. This is replaced by a single immediate send at the point React has already rendered and restored form values, plus a persistent sowrepeaterfieldsadded document listener that sends a fresh message only when a repeater item is actually added.
- Eliminates MutationObserver polling over the entire iframe subtree (attributes, childList, subtree) that was used as a fallback to re-run setupSiteEditorTinyMCEFields on every DOM change. Replaced by the targeted sowbBlockFormInit postMessage pathway.
- Eliminates staggered setInterval fallback (20 × 250 ms = 5 s of polling) for browsers without MutationObserver support.
- Form fingerprint instead of full HTML blob as the useEffect cache key — avoids storing and comparing multi-kilobyte strings on every render cycle.

### Half-initialised field prevention:

- setupTinyMCEField now acquires a sowb-tinymce-initializing lock on entry and releases it in the TinyMCE init event callback, preventing a second concurrent setup attempt from running on the same field.
- A 5 s safety timeout releases the lock if the init event never fires (e.g. the field is removed from the DOM mid-setup), so the field is never permanently stranded.
- setupTinyMCEFieldInitializer calls hasHealthyTinyMCEEditor before re-entering setup and clears stale data-initialized when the editor is found to be unhealthy, directly targeting the data-initialized=true / editorInitialized: false / tinymceIframeCount: 0 stuck state.

### Pre-init listener/poll race prevention:

- A per-field preInitEventNamespace (stored as sowb-pre-init-namespace jQuery data) makes the visibility poll and the sowsetupformfield listener mutually exclusive — whichever fires first cancels the other.
- clearTinyMCEFieldPendingSetup reads the stored namespace back and calls $field.off(), so the listener is always cancelled even when setupTinyMCEField is called independently (e.g. from sortStopEvent).
- wp.oldEditor guard:

- resolveWpEditor() requires the resolved editor object to have an initialize() method before returning it, preventing a wp.oldEditor reference from being used in contexts where it is only partially populated.

### Repeater template placeholder exclusion:

- getEligibleTinyMCEFields and an early-return guard in setupTinyMCEFieldInitializer both reject fields whose textarea ID contains _id_ or whose closest ancestor is .siteorigin-widget-field-repeater-item-html, so unclonable template rows are never initialized.
- triggerVisibleRepeaterVisibilityFieldSetup in admin.js filters template rows out of the $allFields collection before dispatching sowrepeaterfieldsadded, so no downstream consumer ever receives placeholder fields in the event payload.

### postMessage handler deduplication:

The message listener reference is stored as window._sowbTinyMCEMessageHandler and removed via removeEventListener before re-registration, replacing the window.sowbTinyMCEMessageListenerBound boolean that could not actually remove a duplicate listener.

### CSS selector precision:

All sowbGetBlockForm and formSelector references tightened from .siteorigin-widget-form-main to .siteorigin-widget-form.siteorigin-widget-form-main, matching the compound class the PHP template always emits and preventing false matches on elements that only carry the -main suffix.

## Miscellaneous:

- _tinymceIdSeq split into _tinymceDomIdSeq (TinyMCE textarea IDs) and _tinymcePreInitSeq (pre-init event namespaces) — the two counters have different lifetimes and should not share state.
- selectBestFormInstance connected signal now uses this.ownerDocument.documentElement.contains(this) instead of the closure document, so the score is accurate when the function is called from the parent window against iframe-DOM elements.
- sowsetupform iframe listener bails early on absent $form rather than passing null to setupSiteEditorTinyMCEFields.
- All document-level event bindings use namespaced events (.sowTinymce) with explicit .off() before .on() to prevent listener accumulation across hot reloads or repeated script evaluation.

## Testing
Please test this PR when either all Widget Bundle Blocks are present on the page, or a _really_ large Hero block or two. The more you do, the more you remove, etc, the worse the performance would previously get. Please note that this PR intentionally doesn't update the Playwrite tests - those are very time-consuming to get right, and overall this is quite a broad issue. Once the bleeding has stopped, we'll circle back and add tests.

### DIFF
--- a/base/inc/fields/js/media-field.js
+++ b/base/inc/fields/js/media-field.js
@@ -1,12 +1,19 @@
 /* global jQuery, soWidgets */
 
 ( function( $ ) {
+	const isRepeaterTemplateField = function( $field ) {
+		return $field.closest( '.siteorigin-widget-field-repeater-item-html' ).length > 0;
+	};
 
 	const setupMediaField = function( e ) {
 		const $field = $( this );
 		const $media = $field.find( '> .media-field-wrapper' );
 		const $inputField = $field.find( '.siteorigin-widget-input' ).not( '.media-fallback-external' );
 		const $externalField = $field.find( '.media-fallback-external' );
+
+		if ( isRepeaterTemplateField( $field ) ) {
+			return;
+		}
 
 		if ( $field.attr( 'data-initialized' ) ) {
 			return;
@@ -487,7 +494,12 @@
 	window.addEventListener( 'message', function( e ) {
 		if ( e.data && e.data.action === 'sowbBlockFormInit' ) {
 			$( '.siteorigin-widget-field-type-media' ).each( function() {
-				if ( $( this ).attr( 'data-initialized' ) ) {
+				const $field = $( this );
+
+				if (
+					isRepeaterTemplateField( $field ) ||
+					$field.attr( 'data-initialized' )
+				) {
 					return;
 				}
 
--- a/base/inc/fields/js/media-field.js
+++ b/base/inc/fields/js/media-field.js
@@ -471,9 +471,10 @@
 		} );
 	};
 
-	// If the current page isn't the site editor, set up the Media field now.
+	// In the Site Editor iframe, repeater rows added after the initial block
+	// postMessage still rely on the normal field setup event.
 	if (
-		window.top === window.self &&
+		window.top !== window.self ||
 		(
 			typeof pagenow === 'string' &&
 			pagenow !== 'site-editor'
--- a/base/inc/fields/js/multiple-media-field.js
+++ b/base/inc/fields/js/multiple-media-field.js
@@ -156,16 +156,17 @@
 		$field.data( 'initialized', true );
 	};
 
-	 // If the current page isn't the site editor, set up the Multiple Media field now.
-	 if (
-		 window.top === window.self &&
-		 (
-			 typeof pagenow === 'string' &&
-			 pagenow !== 'site-editor'
-		 )
-	 ) {
-		 $( document ).on( 'sowsetupformfield', '.siteorigin-widget-field-type-multiple_media', setupMultipleMediaField );
-	 }
+	// In the Site Editor iframe, repeater rows added after the initial block
+	// postMessage still rely on the normal field setup event.
+	if (
+		window.top !== window.self ||
+		(
+			typeof pagenow === 'string' &&
+			pagenow !== 'site-editor'
+		)
+	) {
+		$( document ).on( 'sowsetupformfield', '.siteorigin-widget-field-type-multiple_media', setupMultipleMediaField );
+	}
 
 	// Add support for the Site Editor.
 	window.addEventListener( 'message', function( e ) {
--- a/base/inc/fields/js/multiple-media-field.js
+++ b/base/inc/fields/js/multiple-media-field.js
@@ -1,8 +1,16 @@
 /* global jQuery, soWidgets */
 
 ( function( $ ) {
+	const isRepeaterTemplateField = function( $field ) {
+		return $field.closest( '.siteorigin-widget-field-repeater-item-html' ).length > 0;
+	};
+
 	const setupMultipleMediaField = function() {
 		const $field = $( this );
+
+		if ( isRepeaterTemplateField( $field ) ) {
+			return;
+		}
 
 		if ( $field.data( 'initialized' ) ) {
 			return;
@@ -172,6 +180,10 @@
 	window.addEventListener( 'message', function( e ) {
 		if ( e.data && e.data.action === 'sowbBlockFormInit' ) {
 			$( '.siteorigin-widget-field-type-multiple_media' ).each( function() {
+				if ( isRepeaterTemplateField( $( this ) ) ) {
+					return;
+				}
+
 				setupMultipleMediaField.call( this );
 			} );
 		}
--- a/base/inc/fields/js/tinymce-field.js
+++ b/base/inc/fields/js/tinymce-field.js
@@ -575,16 +575,19 @@
 			.attr( 'data-tinymce-id', id )
 			.attr( 'id', id );
 
-		let _resolveInitPending;
-		_tinymceInitPending[ id ] = {
-			promise: new Promise( function( resolve ) {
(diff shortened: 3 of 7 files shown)


Misplon:
Thanks.

- Blocking: The tightened selector .siteorigin-widget-form.siteorigin-widget-form-main does not match the block markup in this file. The render path puts siteorigin-widget-form on the Placeholder wrapper and siteorigin-widget-form-main on the child container, not the same element: /tmp/sowb-pr2311.BJ3c13/compat/block-editor/widget-block.js:635 and /tmp/sowb-pr2311.BJ3c13/compat/block-editor/widget-block.js:651. As a result, sowbGetBlockForm() now returns an empty set in both iframe and non-iframe paths: /tmp/sowb-pr2311.BJ3c13/ compat/block-editor/widget-block.js:1120 and /tmp/sowb-pr2311.BJ3c13/compat/block-editor/widget-block.js:1127. The same mismatch also makes the new iframe formSelector miss the form: /tmp/sowb-pr2311.BJ3c13/compat/block-editor/widget-block.js:367. I would keep this as .siteorigin-widget-form-main unless the rendered markup is changed to put both classes on the same element.

AlexGStapleton:
> Blocking: The tightened selector 

That's partially correct. The fix here to ensure both of the tightened selectors are valid in all cases and make some changes to account the knock-on effects by doing this. Currently working on this and will submit a follow-up commit to this PR shortly. I recommend holding off on testing/auditing this PR until then.

Misplon:
Thanks, standing by.

Misplon:
Thanks Alex. I ran another review pass against the current PR head (`db405fd`) after the latest commits and found a couple of additional things worth checking.

- The selector blocker from the earlier comment still looks unresolved. The render path still puts `siteorigin-widget-form` on the `Placeholder` wrapper and `siteorigin-widget-form-main` on the child container (`compat/block-editor/widget-block.js:635` and `compat/block-editor/widget-block.js:651`), while `formSelector` / `sowbGetBlockForm()` still require both classes on the same element (`compat/block-editor/widget-block.js:367`, `compat/block-editor/widget-block.js:1120`, `compat/block-editor/widget-block.js:1127`). Unless there is another markup path I’m missing, those lookups will still miss the form.

- Additional likely blocker: `_tinymceInitPending[id]` can be stranded. `clearTinyMCEFieldPendingSetup()` clears `sowb-tinymce-init-timeout` and removes the initializing data (`base/inc/fields/js/tinymce-field.js:311-318`), but it doesn’t resolve/delete the matching `_tinymceInitPending[id]` entry created at `base/inc/fields/js/tinymce-field.js:578-590`. Since that timeout is the fallback resolver, cleanup during a pending init can leave `sowbGetTinyMCEInitPromise()` waiting indefinitely.

- Additional race to sanity-check: `setupTinyMCEFieldInitializer()` can remove `data-initialized` during an in-flight init (`base/inc/fields/js/tinymce-field.js:790-803`), then `setupTinyMCEField()` immediately returns because `sowb-tinymce-initializing` is still set (`base/inc/fields/js/tinymce-field.js:469-471`). If the original init then completes cleanly, the editor can be healthy but no longer marked initialized, so a later setup event may tear it down/reinitialize unnecessarily.

Syntax checks passed for the three changed JS files, and `git diff --check` passed. These are from review rather than a full browser reproduction pass.

Misplon:
The above comment was only meant to contain the new observations. Thanks.

AlexGStapleton:
All previously flagged issues should be resolved, and ready to be merged into the wb-direct-block-save-bridge branch.

Misplon:
Legend, thanks. I'll  get going this evening.

Misplon:
While testing the latest beta build from this branch, I hit one more blocker in the direct widget block preview flow.

Repro path:

1. Add/edit a direct `SiteOrigin Features` block.
2. Add text in a TinyMCE field using the Visual tab.
3. Save the page.
4. Click the block Preview control.

Result before the fix: the editor canvas flashed/reloaded and ended up with a nested editor/sidebar UI inside the canvas.

Root cause looks to be that `sowSetupForm()` still binds the legacy widget preview handler to the form's `.siteorigin-widget-preview > a` link. In the direct block flow we already have a React preview path, but the legacy handler can still submit the old modal/iframe preview form. In the iframed editor that can navigate the canvas instead of just flipping the block into preview mode.

I stacked a tiny draft PR on top of this branch rather than touching your branch directly:

- #2312
- Base: `feature/sowb-block-editor-iframe-asset-clone-no-polling`
- Head: `fix/direct-block-preview-legacy-handler`
- Commit: `22c7d3ed Block Editor: prevent legacy preview navigation`
- Scope: one file, `compat/block-editor/widget-block.js`

What #2312 does:

- Rebinds the form preview link after `sowSetupForm()` runs.
- Removes the legacy modal/iframe submit click handler for direct blocks.
- Makes the link only switch the React block preview state.
- Adds explicit `preventDefault()` / propagation guards to the toolbar preview/edit controls.

Validation so far:

- Reproduced the nested-editor/sidebar issue before the patch.
- Confirmed the same Features/TinyMCE/save/preview path works after the patch.
- Ran `node --check compat/block-editor/widget-block.js`.
- Ran `git diff --check`.
- Rebuilt the test package as `so-widgets-bundle.1.73.0-beta.zip` with the patch included.

The stacked PR is draft so you can decide whether to pull it in, adapt it, or fold the same idea into this branch another way.

AlexGStapleton:
Merging up.

PR #2315, opened 2026-06-08
Fix Blog widget form freeze on open in Page Builder (#84645)

Misplon:
## Summary

Fixes the ~10s freeze when opening the SiteOrigin Blog widget settings form in Page Builder (support ticket #84645). Customer has confirmed the fix resolves the lag.

## Root cause

Each font field builds a native `<select>` client-side by appending the full bundled Google Fonts list (~6,832 `<option>`s, ~395KB of HTML) — `font.class.php` renders only a single "Use theme font" option, and `base/js/admin.js` appends the rest. The Blog widget has 10 font fields (9 in the bundle + 1 from the Premium addon), one per collapsible section.

WordPress persists each section's open/closed state (`so_field_container_state`). A section saved **open** re-rendered with `siteorigin-widget-section-visible`, and its font `<select>` was built **immediately on form open**. With several sections expanded, the browser parsed tens of thousands of `<option>` nodes (~3.9MB) at once — the freeze. This matched the customer's report exactly: all sections collapsed = no lag; more sections expanded = worse lag.

## Fix

Lazy-populate each font `<select>` on first `focus` instead of eagerly on form open. A one-time `.one('focus', ...)` handler calls the existing `setupFontField()` the first time the user opens a given dropdown. Opening the form is now instant regardless of how many sections are expanded; only the dropdown the user actually opens pays the single-select build cost.

- Keeps the native scrollable dropdown (no select2 / search picker — browsing/recognition is how users pick fonts).
- JS-only change; no PHP changes, no new dependencies.
- `setupFontField()`'s `sow-font-setup` guard + `.one()` prevent double-binding/double-append, including in repeater clones.
- Removed the now-unused `setupFontFieldsOnHover` function.

## Commits

- Fix Blog widget font field initialization lag (always-truthy visibility check, selector, per-select guard)
- Lazy-populate font selects on focus to fix Blog widget form freeze (#84645)

## Testing

- `node --check base/js/admin.js` passes.
- Manual repro in Page Builder: Blog widget with sections expanded now opens instantly; font dropdowns populate on open and show the previously-saved selection; keyboard (tab) focus also populates; repeater-nested font fields populate without double-appending.
- Customer confirmed the lag is resolved.

## Known follow-up (not in this PR)

A section persisted open shows only "Use theme font" until the select is focused (the saved value is never lost — it's re-applied on first focus). A future polish would render a placeholder `<option>` for the saved value in `font.class.php` so the saved font label shows without requiring focus.

### DIFF
--- a/base/inc/fields/font.class.php
+++ b/base/inc/fields/font.class.php
@@ -18,7 +18,10 @@ class SiteOrigin_Widget_Field_Font extends SiteOrigin_Widget_Field_Base {
 				id="<?php echo esc_attr( $this->element_id ); ?>" class="siteorigin-widget-input"
 				data-selected="<?php echo esc_attr( $value ); ?>"
 			>
-				<option value="default" selected="selected"><?php esc_html_e( 'Use theme font', 'so-widgets-bundle' ); ?></option>
+				<option value="default"<?php echo ( empty( $value ) || $value === 'default' ) ? ' selected="selected"' : ''; ?>><?php esc_html_e( 'Use theme font', 'so-widgets-bundle' ); ?></option>
+				<?php if ( ! empty( $value ) && $value !== 'default' ) : ?>
+					<option value="<?php echo esc_attr( $value ); ?>" data-sow-saved-option selected="selected"><?php echo esc_html( isset( $widget_font_families[ $value ] ) ? $widget_font_families[ $value ] : $value ); ?></option>
+				<?php endif; ?>
 			</select>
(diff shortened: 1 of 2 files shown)


AlexGStapleton:
This looks good to me. There's a slight timing risk related to the PHP changes (values may be overridden on load), but that appears to be a very low risk, so not worth worrying about as compared to the issue itself. Especially in PB.

Issue #2343, opened 2026-08-08
Block Editor: Page Builder dialog controls lose WordPress admin styling in the canvas


Misplon:
Open a SiteOrigin widget block with a builder field in the block editor, click **Open Builder**, then **Layouts**. The Prebuilt Layouts dialog renders with unstyled controls: the **Enable** and **Insert** buttons are raw browser buttons rather than WordPress admin buttons.

The builder dialog itself is fine. It is specifically WordPress's own admin UI controls inside the dialog that lose their styling.

## Cause

WordPress scopes every admin button rule in `wp-admin/css/buttons.css` under `.wp-core-ui`. Read from the running canvas:

```
.wp-core-ui .button, .wp-core-ui .button-primary, .wp-core-ui .button-secondary
.wp-core-ui .button-primary
.wp-core-ui .button-primary.hover, .wp-core-ui .button-primary:hover
```

The parent admin page carries that class on `<body>`. **The block editor canvas iframe does not:**

```js
// parent
document.body.classList.contains('wp-core-ui')   // true

// canvas
canvasDoc.body.className
// "block-editor-iframe__body editor-styles-wrapper post-type-post
//  admin-color-modern wp-embed-responsive"
canvasDoc.body.classList.contains('wp-core-ui')  // false
```

So the stylesheet is present and readable in the canvas — `buttons-css`, `forms-css` and `editor-buttons-css` all clone in successfully, and iterating `document.styleSheets` finds the `.button` rules — it simply cannot match anything.

Measured on the Enable button, which carries `class="button-primary so-panels-enable-directory"`:

| Property | Computed in canvas | Expected |
|---|---|---|
| `background-color` | `rgb(239, 239, 239)` | WP admin blue |
| `border` | `2px outset rgb(0, 0, 0)` | `1px solid` accent |
| `border-radius` | `0px` | `3px` |
| `padding` | `1px 6px` | `0 10px` |

Those are the user-agent defaults for `<button>`, i.e. no WordPress styling applied at all.

Text inputs are less affected because `forms.css` is not scoped the same way — the search field computes with a real border and radius, which is why the dialog looks partly styled rather than completely raw.

## Not a regression

This is pre-existing and was simply never visible: until the fix for #2341 the Open Builder button did nothing, so nobody reached the dialog to see its buttons. Cosmetic only — the controls work, they just look wrong.

## Two ways to fix, and the tradeoff

[...]

Issue #2329, opened 2026-07-22
E2E Suite Does Not Run Routinely: CI Hangs Before Tests


Misplon:
The Playwright e2e workflow hangs at the browser-install step and times out, so tests never execute in CI; local runs require env-file/Playground setup most contributors don't do. Consequence, discovered 2026-07-22: the block-editor spec was silently dead — an ambiguous form locator (strict-mode violation, introduced alongside 5a0d419b's container classes) failed the first test and serial mode skipped the entire file. Fixed in 220cffc9, but two genuine pre-existing test failures were also exposed (filed separately). Getting this suite running in CI is the root fix.
- Also observed: the first test after Playground boot intermittently fails with 'Not logged in' and passes on retry (session warm-up race).

PR #2344, opened 2026-08-08
Block Editor: Wait for the Canvas to Load What the Clone Walk Cannot Supply

### LINKED ISSUE #2341: Layout Slider Block: Open Builder Button Does Nothing in the Block Editor
Open a widget block with a builder field in the block editor, click **Open Builder**, and nothing happens. Sometimes it works.

Affects Layout Slider in the free plugin and eight Premium addons: Tabs, Accordion, Carousel, Lightbox Builder, Contact Form Fields, Toggle Visibility, Embed Blocker, `inc/gate.php`. Only Layout Slider has been exercised.

[console.log](https://github.com/user-attachments/files/30809863/console.log)

> **Corrected 8 Aug 2026.** An earlier version of this issue named `#editor-js-after` as the cause and explicitly ruled out script concatenation. Both were wrong, and the concatenation dismissal actively pointed the reader away from the real cause. See *Cause* below; the superseded claims are listed at the end so the record stays honest.

---

## Cause — `#jquery-core-js` is never cloned, because it does not exist

The block editor renders in an iframe. Widgets Bundle copies assets into it from the parent document using a hand-written allowlist of element IDs, `sowbCanvasCloneElements` (`compat/block-editor/widget-block.js:1586`).

**Its first entry is `'#jquery-core-js'` (`:1587`), and that element is not on the page.** WordPress concatenates `jquery-core` into a single bundled request:

```
load-scripts.php?c=1&load[chunk_0]=jquery-core,jquery-migrate,react,react-jsx-runtime,wp-autop,…
```

That bundle carries one element ID, which is not `jquery-core-js`. So the selector matches nothing, `sowbCloneElementsToCanvas` (`:2214`) silently `continue`s, and **jQuery has never once been cloned into the canvas.**

jQuery **UI** is *not* concatenated — `jquery-ui-core`, `mouse`, `slider`, `sortable`, `resizable`, `draggable`, `touch-punch` each emit their own element. Those clone successfully, then evaluate inside a canvas whose jQuery is still in flight inside the bundle. Every one throws `ReferenceError: jQuery is not defined`, and so do `iris`, `color-picker`, `wplink`, the Widgets Bundle field scripts, and finally `siteorigin-panels.min.js`.

Page Builder therefore never registers `jQuery.fn.soPanelsSetupBuilderWidget`, and the button has nothing bound to it.

It is intermittent because it is a race against one large external file: it turns on cache state and network timing, not on configuration.

### Evidence

Failing state, measured inside the canvas iframe with the button confirmed broken:

```
editorJsAfter:   true      jqueryCore:      false     ← jQuery absent
jqueryUiCore:    true      backbone:        true
soPanelsAdminJs: true      panelsDialogTpl: true
totalScripts:    150       jQueryDefined:   "function"
soPanelsReady:   false     ← Page Builder never registered
```

Parent document, same page:

```
parent_jqueryCore:    false     ← no #jquery-core-js element exists
parent_jquerySrc:     null
parent_jqueryUiCore:  true      ← jQuery UI does emit separately
```

`totalScripts: 150` with every entry after `#editor-js-after` present is the key result: **the clone walk completes.** Nothing aborts it.

## Reproduction

Cache-disable alone is not reliable. What reproduced it:

1. Log out of WordPress, then log back in
2. DevTools → Network → tick **Disable cache**, keep DevTools open
3. Open a post containing a Layout Slider widget block
4. Edit widget → expand the frame → **Open Builder** → nothing happens

Console shows the `jQuery is not defined` cascade ending at `siteorigin-panels.min.js`.

**Any fix must be verified this way.** A pass on a warm cache proves nothing, which is how this went unresolved for a day.

## Also verified

**Page Builder's registration is all-or-nothing.** `siteorigin-panels/js/siteorigin-panels/main.js:15` reads `Backbone.Events` at module top level; `jQuery.fn.soPanelsSetupBuilderWidget` is not assigned until `:65`. If anything it needs is missing when that file evaluates, it throws at line 15 and the button has nothing bound — silently, with no error naming the button.

**Nothing retries.** `base/js/admin.js:585` is `if ( typeof jQuery.fn.soPanelsSetupBuilderWidget !== 'undefined' )` — a plain `if`, no `else`, no retry. If Page Builder has not loaded at that moment, builder fields are skipped and the form is marked initialised.

**The readiness gate has no clause for builder fields.** `sowbGetFormSetupState` (`widget-block.js:1309-1325`) waits for `sortable` when a repeater is present and `slider` when a slider is present. There is no equivalent for a builder field, and `soPanelsSetupBuilderWidget` appears nowhere in the Widgets Bundle.

**`wp.media` is absent inside the canvas** (measured at runtime), while Page Builder's style image controls call it directly at `siteorigin-panels/js/siteorigin-panels/view/styles.js:163` and `:185`.

**`plupload` has a hard `moxiejs` dependency** (`wp-includes/script-loader.php:1043-1044`). Cloning `plupload` without it leaves `window.plupload` undefined.

**`backbone`, `moxiejs`, `plupload`, `underscore` and `so-panels-admin` do each emit their own `<script id="…-js">` element**, and `moxiejs-js` and `so-panels-admin-js` were confirmed present inside the canvas after cloning. The allowlist works for these. `jquery-core` is the exception that breaks it.

## How to fix it

The allowlist is keyed on element IDs, and the asset it most needs has no ID to name. So the ordering work is not enough on its own — jQuery has to reach the canvas first, and be known to have arrived before its dependants are appended.

Open questions for whoever picks this up, none yet answered:

- Is the concatenated bundle already loaded by the canvas itself, or does it need cloning? If the canvas loads its own copy, the fix is to **wait** for it rather than to clone one.
- Can the bundle be matched by `src` (it contains `load-scripts.php` and `jquery-core`) rather than by ID?
- Should `jquery-core` simply be dequeued from concatenation in the block editor, so it emits with a real ID?

**The bigger alternative, if the allowlist proves unfixable:** have Page Builder register its own assets into the canvas via `enqueue_block_assets`, and delete Page Builder's entries from the Widgets Bundle allowlist. Structurally better — one plugin should not maintain an allowlist of another plugin's internal handles — but a much larger change touching WordPress internals.

## Only if you take the bigger alternative

Three blocking steps, all in `siteorigin-panels`:

1. **Make `enqueue_admin_scripts()` safe inside the iframe asset collector** (`inc/admin.php`). `wp_enqueue_media()` at `:508` short-circuits on `did_action()`, so calling it during collection consumes the parent's call and leaves the Site Editor without media. The template output at `:771-778` echoes unbuffered at a point before the doctype.

2. **Add an `enqueue_block_assets` callback** registering Page Builder's assets into the canvas. Use `SiteOrigin_Panels_Admin::is_block_editor()` (`inc/admin.php:136`, null-checks the screen), not `is_admin()` (`:123`, dereferences `$screen->base` unguarded).

3. **Get the dialog templates into the canvas.** Without them the dialog has nothing to render.

**Caveats that must be resolved, not assumed:**

- `doing_action( 'enqueue_block_assets' )` identifies an action, not the collector. Pass an explicit mode instead.
- The collector's internals were verified on WordPress 7.0.3 only. Confirm on every supported version, and confirm whether it runs more than once.
- **Do not delete the Widgets Bundle allowlist entries until all eight Premium addons are verified.** They are not on disk here. Removing the fallback while they are unverified could silently break their builder UI.

## Not investigated

When the field is inside the canvas, should the dialog render inside the iframe or cover the whole admin screen? Neither was tested. Confining it to the iframe would clip it; taking it full-screen would need the dialog to work across two documents. Which is right, and how much work either is, is unknown.

## Companion issues

**`siteorigin-panels#1350`** — `initSortable()` throws in the canvas. **Not a blocker for the button:** the dialog is already visible when it throws, because `view/dialog.js:374` shows the element before triggering `open_dialog_complete` at `:377`. But Backbone fires handlers synchronously with no `try`/`catch`, so everything registered after it dies — including row resizing. **Blocker for sign-off, not for the button.** Fix exists on `fix/1350-init-sortable-iframe`; audit the commit before merging rather than trusting the branch.

**`siteorigin-panels#1351`** — widget lists render with bullets in the canvas. Cosmetic, independent, one-line CSS fix.

## Verification

WordPress 7.0.3, Page Builder 2.36.0, Widgets Bundle 1.74.2, **`SCRIPT_DEBUG` off** — the default configuration, not a developer one.

1. Add a Layout Slider widget block to a post
2. Edit widget → expand the frame → **Open Builder**
3. The Page Builder dialog opens
4. Add Widget lists the widget picker
5. Add a widget, set a value, Done, Save
6. Confirm the value is in `post_content` and renders on the front end
7. Console clean — no `jQuery is not defined`, no `Cannot read properties of undefined`

**Steps 1-6 were observed passing** on a build of `fix/editor-canvas-clean` — but on a warm cache, which is now known not to exercise the bug.

**Step 7 has never been observed passing.**

## Prior art

`fix/editor-canvas-clean` in the Widgets Bundle has three commits adding Page Builder handles to the allowlist. It adds the missing handles including `moxiejs`, and reorders `#editor-js-after` to follow `#editor-js`. **It does not address the cause above** — it cannot, because `#jquery-core-js` matches nothing regardless of order. Assess and build on it rather than starting over.

**Do not use `fix/editor-canvas-inline-script-order`.** It carries 23 commits including a backed-out AI abilities merge (draft PR #2342, three critical defects).

## Superseded claims

Kept so nobody re-derives them:

1. ~~"On a cold cache the inline `#editor-js-after` runs while `#editor-js` is still in flight, throws, and the throw takes out everything after it."~~ It does throw, and the error is real and worth removing — but it takes out nothing. Per the HTML spec, an exception inside a dynamically inserted inline script is reported as a global error on that window, not rethrown into the `appendChild()` caller. Measured: 150 scripts present, every entry after it cloned. **Cosmetic.**

2. ~~"Some core handles are concatenated — `jquery-core` among them — but not the ones this allowlist depends on. So concatenation is not what breaks it."~~ False. The allowlist's first entry is `#jquery-core-js`. Concatenation is exactly what breaks it. The original check sampled `backbone`, `underscore` and `so-panels-admin`, found them separate, and generalised to a handle that is not.

3. ~~"An earlier draft said the copying approach was unworkable because script concatenation hid the elements. That was checked and is false — ignore it."~~ Withdrawn. That draft was closer to right than the correction that replaced it.

4. ~~"`SCRIPT_DEBUG` explains the intermittency."~~ Still false. It was off in every observation, passing and failing alike.

Misplon:
Fixes the Open Builder button doing nothing in the block editor. Closes #2341.

## What was wrong

The block editor renders in an iframe. Widgets Bundle clones the assets a widget form needs into that iframe by walking a hardcoded allowlist of element IDs, `sowbCanvasCloneElements`.

WordPress concatenates core scripts into `load-scripts.php?c=1`, and **those bundle elements carry no `id`** (`wp-includes/script-loader.php`). `jquery-core` is inside one, so `#jquery-core-js` does not exist in the admin document on a default install. The first entry in the allowlist has never matched anything and **jQuery has never once been cloned into the canvas.**

The canvas loads its own jQuery in its `<head>`, asynchronously, while the clone walk runs from a React `ref` that fires first. So every cloned script that needs jQuery evaluated against a window without it and threw — jQuery UI, iris, color-picker, wplink, the field scripts, and finally `siteorigin-panels.min.js`, which therefore never registered `jQuery.fn.soPanelsSetupBuilderWidget`. The button had nothing bound to it.

It presented as intermittent because it is a race against one large external fetch: a warm cache wins, a cold cache loses. Every new build changes `?ver=` and busts the cache, which is why it appeared to fail on every fresh install.

**With `SCRIPT_DEBUG` on there is no concatenation**, `#jquery-core-js` exists, the walk clones jQuery ahead of its dependants, and it works. That is how this shipped.

## The fix

Gate the clone walk until the canvas has evaluated the handles the walk cannot supply itself, deciding that **per handle at runtime** rather than assuming which ones WordPress bundles.

Waiting rather than retrying is deliberate: `sowbCloneElementToCanvas` dedupes by id against the whole canvas document, so a script appended too early can never be replaced. Retrying afterwards is a no-op unless failed nodes are removed first, and removing nodes is exactly the machinery this avoids.

The wait is bounded at 20 × 250 ms and then proceeds regardless, so the worst case is the behaviour that shipped before, five seconds later. State lives on the canvas window so every caller shares one budget, and the timer belongs to the canvas so teardown disposes of it.

A second commit covers a distinct race it exposed: `siteorigin-widget-admin` reaches the canvas through its `<head>` (WordPress enqueues it while collecting iframe assets) while `so-panels-admin` arrives only through the clone walk. A form can finish setting up in the window where `sowSetupForm` exists but `soPanelsSetupBuilderWidget` does not, and `base/js/admin.js` checks for it with a bare `if` and no retry — so those builder fields stayed dead with nothing in the console to explain it. Re-runs the setup once `so-panels-admin` has loaded. Fields Page Builder has already initialised are skipped, so an open dialog or unsaved edit is never touched.

## Verified

Cold cache, on the bug's own reproduction — logged out, logged back in, DevTools "Disable cache" on:

- No `ReferenceError: jQuery is not defined`
- `jQuery`, `Backbone`, `underscore` and `jQuery.ui` all live in the canvas
- `soPanelsSetupBuilderWidget` registered
- **Open Builder opens and the dialog renders with existing content intact**
- Gate engaged and resolved after one 250 ms tick (`sowbCanvasCloneWaitAttempts: 1`, `sowbCanvasCloneWaitTimer: 0`)

Canvas script count dropped 150 → 130, because the existing dedupe now runs its predicates at a point where they can answer truthfully and stops re-cloning handles the canvas already has.

Tested against Page Builder built from `fix/canvas-combined`.

## Not verified — please cover before merge

- **`SCRIPT_DEBUG` on.** The most important gap. `f40516bd` changed behaviour specifically in that configuration, and it is the configuration that currently works, so a regression would land there. Needs a `wp-config.php` edit.
- **Fail-open path** — block the canvas `load-scripts.php` request and confirm `sowbCanvasCloneWaitAttempts === 20`, `sowbCanvasCloneWaitTimer === 0`, no runaway timer.
- **Multiple widget blocks** — `sowbCanvasCloneWaitAttempts` must stay 1–3 and not scale with block count, which would mean callers are advancing the budget instead of ticks.
- **Site Editor**, and the older end of the supported WordPress range.

## Known and accepted

- **The wait budget does not reset after timeout.** A load whose canvas dependencies take longer than 5 s leaves the button dead for that canvas until the editor is reopened. Deliberate — recovery would require removing already-appended nodes. Rationale and the alternatives considered are recorded in [this comment on #2341](https://github.com/siteorigin/so-widgets-bundle/issues/2341#issuecomment-5226602860).
- **No regression tests.** The repo has no JS unit runner and the Playwright suite has not run since October 2025 (#2329). Verification here is manual and browser-based.
- **`.gitignore`** gains `docs/plans/` and `tasks/runs/` in the first commit. Unrelated to the fix — say the word and I will strip it.

## Follow-up

#2343 — the dialog now opens, which exposed that the canvas body lacks `wp-core-ui`, so WordPress's admin button styling cannot match inside it. Cosmetic, pre-existing, separate.

The first three commits are from `fix/editor-canvas-clean`, which adds Page Builder's handles to the allowlist and was never separately merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,8 @@ vendor/
 
 .cursorrules
 .claude/
+docs/plans/
+tasks/runs/
 # Playwright.
 tests/cache
 tests/results
--- a/compat/block-editor/widget-block.js
+++ b/compat/block-editor/widget-block.js
@@ -2385,7 +2385,15 @@ const sowbCanvasIsReadyForAssetClone = ( frame, sourceDoc ) => {
  * @returns {boolean} True when Page Builder was available to set fields up.
  */
 const sowbEnsureCanvasBuilderFields = ( canvasWindow ) => {
-	const canvasJQuery = canvasWindow && canvasWindow.jQuery;
+	let canvasJQuery;
+
+	try {
+		canvasJQuery = canvasWindow && canvasWindow.jQuery;
+	} catch ( e ) {
+		// Ignore cross-window errors (e.g. iframe detached or navigated
+		// between the caller's checks and this read).
+		return false;
+	}
 
 	if (
 		! canvasJQuery ||
--- a/compat/block-editor/widget-block.js
+++ b/compat/block-editor/widget-block.js
@@ -1591,16 +1591,15 @@ const sowbCanvasCloneElements = [
 	'#wp-tinymce-js',
 	'#utils-js',
 	// WordPress emits `-extra` (localization) before a script and `-after`
-	// following it. Cloning `#editor-js-after` ahead of `#editor-js` put them in
-	// the wrong relative order, so it ran against a window with no wp.editor and
-	// the throw aborted the walk before jQuery's dependants were cloned.
+	// following it, so this list keeps them in that order.
 	//
-	// This corrects the ORDER only. It does not sequence them: the loop appends
-	// synchronously, and an inline script executes on append rather than waiting
-	// for the external file it depends on. Every inline entry in this list has
-	// that same exposure, which is why sowbEnsureEditorJs() exists as a retry.
-	// A real fix would chain each inline script off its external partner's load
-	// event.
+	// Order is all this does. It does not sequence them: the loop appends
+	// synchronously and an appended inline script runs immediately rather than
+	// waiting for the external file it depends on. When one throws, the
+	// exception is reported to the canvas window and does not propagate back
+	// out of appendChild(), so the walk carries on - the script simply had no
+	// effect. sowbEnsureIframeOldEditorApi() re-injects editor-js for exactly
+	// that case.
 	'#editor-js-extra',
 	'#editor-js',
 	'#editor-js-after',
--- a/compat/block-editor/widget-block.js
+++ b/compat/block-editor/widget-block.js
@@ -2336,11 +2336,18 @@ const sowbCanvasUnclonableDependencies = ( canvasWindow, sourceDoc, canvasDoc ) 
  * evaluates against a window that has none.
  *
  * sowbCloneElementToCanvas() dedupes by id against the whole canvas document,
- * so a script appended too early can never be replaced. That is why this gates
- * the walk rather than retrying it afterwards.
+ * so a script appended too early can never be replaced. That is why the walk
+ * never runs before the canvas holds its dependencies and its body: cloning
+ * into a canvas that cannot evaluate the clones is worse than not cloning,
+ * because the failed nodes block every later attempt.
  *
- * Fails open: once the budget is spent the walk runs regardless, which is the
- * behaviour that shipped before this gate existed.
+ * There is no budget. Ready is observable, so the wait polls until the
+ * dependencies and the body exist, however long that takes. A canvas document
+ * that finishes loading without them gets one console.warn naming the missing
+ * handles, and the poll continues so a dependency arriving late by any route
+ * is still picked up. A canvas still loading after twenty ticks gets one
+ * "still waiting" warn. Wait state lives on the iframe element, per canvas
+ * document, so the about:blank to blob navigation cannot discard it.
  *
  * @param {HTMLIFrameElement} frame     The editor canvas iframe.
  * @param {Document}          sourceDoc The document the walk clones from.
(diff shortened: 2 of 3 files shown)


Misplon:
## SCRIPT_DEBUG pass — done, and it validates `f40516bd`

This was the outstanding gap. Tested with `define( 'SCRIPT_DEBUG', true )`, then `wp-config.php` restored byte-identical (SHA unchanged).

Confirmed the setting actually took effect rather than trusting the constant:

```
parentLoadScriptsBundles: 0        // concatenation genuinely off
parentHasJqueryCoreId:    true     // #jquery-core-js now exists in the parent
canvasHasJqueryCoreId:    true     // and in the canvas head
```

Gate behaviour:

```
waitAttempts: 1     waitTimer: 0
jQuery: function    Backbone: object    jQuery.ui: true
soPanelsReady: true    so-panels-admin-js: present    dialog templates: present
totalScripts: 172
```

Open Builder opens, the dialog renders with its full layout, console clean.

**Note the expectation changed during review, and this is the more important half.** The plan originally asserted the gate would *never engage* under SCRIPT_DEBUG. `f40516bd` deliberately changed that, and this run is what confirms the change was right:

- Before that commit, the gate short-circuited whenever the source document had the handle — which under SCRIPT_DEBUG it does.
- But the walk would then have **skipped cloning it anyway**, because `sowbCloneElementToCanvas` dedupes by id against the whole canvas document and the canvas head already holds `#jquery-core-js`.
- Net effect would have been the original race, in the configuration everyone assumed was healthy.

With the fix a handle counts as ours only when the source has it *and* the canvas does not, so here the gate correctly waits one 250 ms tick for the canvas's own copy and proceeds.

So SCRIPT_DEBUG now engages the gate by design, resolves in a single tick, and produces a working editor.

AlexGStapleton (approved):

AlexGStapleton:
> The wait budget does not reset after timeout. A load whose canvas dependencies take longer than 5 s leaves the button dead for that canvas until the editor is reopened. Deliberate — recovery would require removing already-appended nodes. Rationale and the alternatives considered are recorded in https://github.com/siteorigin/so-widgets-bundle/issues/2341#issuecomment-5226602860.

This will definitely need to be resolved before/after https://github.com/siteorigin/so-widgets-bundle/issues/2329

Misplon:
Pushed four commits resolving the recorded limitation — the wait budget that never reset, leaving Open Builder dead for a canvas whose dependencies took longer than 5 s.

**The budget is gone rather than reset.** The end condition it approximated is observable: the gate now polls until the canvas holds its dependencies *and* its body, however long that takes, and the walk never runs without them — cloning into a canvas that cannot evaluate the clones is what made the failure permanent, so the poisoned state is now unreachable instead of recovered from. A canvas that finishes loading without the dependencies gets one `console.warn` naming the missing handles (debounced 1 s, because a healthy canvas measurably reports `complete` a beat before its scripts finish evaluating), and polling continues so a dependency arriving late by any route still recovers. A canvas still loading at ~5 s gets one "still waiting" warn. Wait state lives on the iframe element per canvas document — state on the canvas window is destroyed by the about:blank → blob navigation, which could previously strand the wait with the form-side latch already set.

Also closed: `sowbEnsureIframeOldEditorApi` was a second door into the canvas, appending its retry script under the same permanent dedupe independently of the walk; it now defers behind the same readiness check.

**Tests** (`wb-canvas-asset-gate.test.js`, local via `npm run tests`; CI is #2329): both seed a saved widget and reload while intercepting the canvas's concatenated jquery bundle. The slow case holds it 7 s and requires clean recovery through to `soPanelsSetupBuilderWidget` registering; the emptied case delivers an empty bundle and requires exactly one diagnostic and nothing cloned — verified red against the previous gate, green here. Also verified by hand end to end: Layout Slider → add frame → Open Builder → dialog renders.

**Known limit, unchanged from this PR's base:** the pre-existing form-setup retry (40 × 250 ms, fail-closed) is untouched, so dependencies arriving after ~10 s can leave an already-open form uninitialised until the block re-renders or the editor reopens. On the base behaviour that world also had a poisoned canvas that could never recover; now it recovers on the next render. Extending that budget is separate work. #2361 tracks further e2e coverage.

PR #2345, opened 2026-08-09
Widget Blocks: Stop Stripping Saved Rich-Text Content for Logged-Out Visitors

Misplon:
> ### 🧪 Ready-to-test build — install this
> **[so-widgets-bundle.1.75.0-pr2344-2345.zip](https://siteorigin.com/wp-content/uploads/2026/08/so-widgets-bundle.1.75.0-pr2344-2345.zip)**
>
> This is **this PR combined with #2344** (the Open Builder jQuery-gate fix), so the block-editor builder works while you test. Install this zip to check that admin-authored iframes/rich text render for logged-out visitors — no local build needed.

---

Fixes a SiteOrigin widget block rendering an admin-authored `<iframe>` (or any rich HTML) to a logged-in admin but stripping it for logged-out visitors.

## What was wrong

`render_widget_block()` re-runs `$widget->update( $instance, $instance )` on every slow-path render (`compat/block-editor/widget-block.php`). Inside it, the TinyMCE field sanitizer branches on the **viewer's** capability:

```php
// base/inc/fields/tinymce.class.php
if ( current_user_can( 'unfiltered_html' ) ) { /* keep */ }
else { $sanitized_value = wp_kses_post( $value ); } // strip
```

`current_user_can()` asks about whoever is *viewing* the page. A logged-out visitor never has `unfiltered_html`, so the iframe an admin saved is stripped on every public view — while the admin still sees it. Every styled widget takes this slow path (its preview `widgetMarkup` is blanked at save because it carries inline `<style>`), so the re-sanitize runs on essentially every front-end render.

This affects **every bundled widget with a rich-text (TinyMCE) field**, not just the Editor widget: Editor, Tabs, Accordion, Features, Testimonials, Anything Carousel, Hero, Contact Form, Google Map — and TinyMCE content nested inside a Layout Slider.

## Why it is safe to stop re-sanitizing at render

Widget block attributes are already sanitized **at save, under the saving user**. Since WordPress 5.3.1, `content_save_pre` → `wp_filter_post_kses` → `pre_kses` → `wp_pre_kses_block_attributes` → `filter_block_content()` recurses the decoded block attributes and `wp_kses()`-es every string for users without `unfiltered_html`. So a low-privileged author's `<script>`/`<iframe>` is stripped when they save; an admin's is intentionally kept. Re-sanitizing at render with the *viewer's* capability is a redundant second pass keyed to the wrong user — the bug.

## The change

The TinyMCE sanitizer keeps the stored value when we are rendering a widget block on a core that has the save-time sanitizer:

```php
if (
    ( ! empty( $GLOBALS['SITEORIGIN_WIDGET_BLOCK_RENDER'] ) && function_exists( 'wp_pre_kses_block_attributes' ) )
    || current_user_can( 'unfiltered_html' )
) {
    $sanitized_value = $value;
} else {
    $sanitized_value = wp_kses_post( $value );
}
```

`render_widget_block()` already sets `SITEORIGIN_WIDGET_BLOCK_RENDER` for the render; the `function_exists()` guard means a pre-5.3.1 core (which lacks the save-time block-attribute sanitizer) keeps the old render-time behaviour. Save paths and editor previews never set the flag, so their capability semantics are unchanged. All of `update()`'s non-security work (measurement combining, colour shaping, URL escaping, unknown-key stripping) still runs at render, so there is no styling or shape change.

Supporting hardening in `render_widget_block()`, since the flag now controls sanitization:

- The render flag, wrapper filters, widget anchor and output buffer are cleaned up in a `finally`, so a widget that throws mid-render cannot leave the flag set for the rest of the request.
- The flag is restored (not blindly unset) so a nested widget-block render cannot clear it for a still-active outer render.
- The anchor filter teardown is scoped to the render that added it.

## Scope notes

- **Nested Layout Slider content is covered**: a known widget nested in a slider is re-sanitized through its own `update()`, which now honours the flag. (A nested widget whose class does not resolve still hits Page Builder's own kses fallback — pre-existing, unrelated, out of scope here.)
- Independent of the Open Builder canvas fix (#2341 / PR for `widget-block.js`); they touch disjoint files and were tested together on a combined build.
- Existing content needs no migration or re-save — it was sanitized by core at its original save and renders as stored.

## Tests

Adds `tests/e2e/wb-widget-block-render-capability.test.js`:
- An admin-authored iframe in an Editor widget block renders for a logged-out visitor (asserted on a real `<iframe>` element on the live render path, forced via POST since REST saves pre-render markup).
- Content saved by a user without `unfiltered_html` is stripped at save.
- The block-renderer preview endpoint rejects unsaved `widgetData` (schema), so request-supplied attributes cannot reach the render path.

Mutation-tested: reverting the field change makes the render test fail; restored, it passes.

## Manual verification

On WordPress 7.0.3, before/after on the same pages: an Editor widget iframe and a Layout-Slider-nested Editor iframe are stripped for logged-out visitors before, present after; an author without `unfiltered_html` still has script/iframe stripped at save; button/measurement styling is byte-identical before and after.

### DIFF
--- a/base/inc/fields/tinymce.class.php
+++ b/base/inc/fields/tinymce.class.php
@@ -561,7 +561,14 @@ class SiteOrigin_Widget_Field_TinyMCE extends SiteOrigin_Widget_Field_Text_Input
 			$value = wpautop( $value );
 		}
 
-		if ( current_user_can( 'unfiltered_html' ) ) {
+		// Widget block content is sanitized at save under the user who saved it.
+		// When rendering the block we keep the stored value as is — checking the
+		// current user here would apply the viewer's capability to it, stripping
+		// admin-authored markup (e.g. iframes) for logged out visitors.
+		if (
+			! empty( $GLOBALS['SITEORIGIN_WIDGET_BLOCK_RENDER'] ) ||
+			current_user_can( 'unfiltered_html' )
+		) {
 			$sanitized_value = $value;
 		} else {
 			$sanitized_value = wp_kses_post( $value );
--- a/base/inc/fields/tinymce.class.php
+++ b/base/inc/fields/tinymce.class.php
@@ -565,8 +565,14 @@ class SiteOrigin_Widget_Field_TinyMCE extends SiteOrigin_Widget_Field_Text_Input
 		// When rendering the block we keep the stored value as is — checking the
 		// current user here would apply the viewer's capability to it, stripping
 		// admin-authored markup (e.g. iframes) for logged out visitors.
+		// wp_pre_kses_block_attributes() is the core save-time sanitizer that
+		// guarantee rests on; where a very old core lacks it, keep the old
+		// render-time behaviour.
 		if (
-			! empty( $GLOBALS['SITEORIGIN_WIDGET_BLOCK_RENDER'] ) ||
+			(
+				! empty( $GLOBALS['SITEORIGIN_WIDGET_BLOCK_RENDER'] ) &&
+				function_exists( 'wp_pre_kses_block_attributes' )
+			) ||
 			current_user_can( 'unfiltered_html' )
 		) {
 			$sanitized_value = $value;
--- a/compat/block-editor/widget-block.php
+++ b/compat/block-editor/widget-block.php
@@ -672,73 +672,78 @@ class SiteOrigin_Widgets_Bundle_Widget_Block {
 		$instance = $block_content['widgetData'];
 		add_filter( 'siteorigin_widgets_wrapper_classes_' . $widget->id_base, $add_custom_class_name );
 
-		ob_start();
+		// The render flag skips render time sanitization of stored values, so it
+		// must never outlive this render — clean it up even if a widget throws.
+		try {
+			ob_start();
 
-		$always_render_widget_list = array(
-			'SiteOrigin_Widget_PostCarousel_Widget',
-			'SiteOrigin_Widgets_ContactForm_Widget',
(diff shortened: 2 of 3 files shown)


AlexGStapleton (approved):
This works, but it's a bit of an anti-pattern. The ideal approach here is to separate the saving and normalisation pipeline and continue to normalise the output. I understand that this and the other PRs are pragmatic patches, so I think it's good to go. With that said, this will leave the door open for stored XSS. Something to keep in mind - it should hopefully only be applicable to older data.

Long term, it would be worth considering shifting away from the current methods of loading data and rendering blocks. Relying on a global render flag ($GLOBALS['SITEORIGIN_WIDGET_BLOCK_RENDER']) and parent-to-iframe DOM cloning is inherently fragile. Because these are pragmatic patches over legacy foundations, fixing one race condition or sanitisation edge case often exposes another (which explains why so many hardening commits were required here).

In other words, it's a band-aid that will always require more band-aids due to the nature of the patch. This is why it took so many commits are required every time a new issue is identified. 

Creating a proper update and render pipeline is more robust than a global flag/DOM cloning approach. Those changes were purely a pragmatic patch intended to resolve the significant compatibility issues introduced by the Block Editor Canvas changes, and while it does do that, it's proving to be more expensive and prone to breaking than I think either of us expected.

Misplon:
Thanks @AlexGStapleton. First stages of a plan logged here in response to your feedback https://github.com/siteorigin/so-widgets-bundle/issues/2362.

PR #2358, opened 2026-08-20
Fix: Don't pass null $old_value to single-arg callable sanitizers

AlexGStapleton:
`Deprecated: intval(): Passing null to parameter #2 ($base) of type int is deprecated in wp-content/plugins/so-widgets-bundle/base/inc/fields/base.class.php on line 405`

Fix deprecated intval() call when a field uses a callable sanitizer such as 'intval'. sanitize() passed $old_value as a second argument to every callable sanitizer, so a null $old_value was forwarded to intval()'s optional $base parameter, triggering a PHP 8.1+ deprecation and, when a non-null old value was present, silently converting with the wrong base.

Only pass $old_value to callables that declare a second required parameter:
- Add private sanitize_callback_accepts_old_value() using reflection (ReflectionFunction / ReflectionMethod, incl. 'Class::method' strings), defaulting to false on reflection failure.
- One-argument callables like 'intval' are now invoked with the value only; callables that genuinely require $old_value keep receiving it.
- Uses the two-argument ReflectionMethod constructor for PHP 7.0 compatibility.

Add `tests/BaseFieldSanitizeTest.php` covering the intval sanitizer (null and non-null old value), one-param callables, and two-required-param callables (closure, array, and static method forms). The intval test reproduces the reported deprecation against the unpatched code.

### DIFF
--- a/base/inc/fields/base.class.php
+++ b/base/inc/fields/base.class.php
@@ -402,7 +402,11 @@ abstract class SiteOrigin_Widget_Field_Base {
 				default:
 					// This isn't a built in sanitization. Maybe it's handled elsewhere.
 					if ( is_callable( $this->sanitize ) ) {
-						$value = call_user_func( $this->sanitize, $value, $old_value );
+						if ( $this->sanitize_callback_accepts_old_value( $this->sanitize ) ) {
+							$value = call_user_func( $this->sanitize, $value, $old_value );
+						} else {
+							$value = call_user_func( $this->sanitize, $value );
+						}
 					} elseif ( is_string( $this->sanitize ) ) {
 						$value = apply_filters( 'siteorigin_widgets_sanitize_field_' . $this->sanitize, $value );
 					}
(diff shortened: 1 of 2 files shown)


Misplon (changes requested):
Verified: new tests fail on develop (5), branch green (25). Closure, `'Class::method'` and invokable branches checked. Premium CPT Builder's `sanitize_reserved_post_types( $post_type, $old_value )` still receives `$old_value`. Also fixes a latent fatal: 1-arg internal sanitizers like `'strlen'` throw `ArgumentCountError` on develop.

**Change requested: the check is too broad.**

PHP ignores extra arguments to user-defined functions. The deprecation, wrong-base and `ArgumentCountError` problems only occur with internal functions. But `required >= 2` also strips `$old_value` from user callables that got it on develop:

- `function ( $value, $old_value = null )` — loses `$old_value`
- `function ( $value, ...$args )` — loses it
- `'sanitize_title_with_dashes'` (Premium lightbox) — fallback behaviour changes

Fix:

```php
return ! ( $reflection->isInternal() && $reflection->getNumberOfRequiredParameters() < 2 );
```

Internal + fewer than 2 required params → value only. Everything else → two args, as on develop. `Closure::fromCallable( 'intval' )` reflects internal on 8.3, so it is covered too.

**Tests:**

1. The one-param data provider never registers the provided callables — the variadic wrapper is the sanitizer in every case. The one-param `ReflectionMethod` branch is untested.
2. After the fix above, assert single-arg dispatch for internal callables only. Assert user callables (one-param, optional-second, variadic) still receive both args.

Misplon:
Thanks, Alex. Your commit stays as the base; I have pushed three commits on top that carry the change requested in review, plus a develop merge.

- The predicate is now `! ( isInternal() && getNumberOfRequiredParameters() < 2 )`. PHP ignores surplus arguments to user-defined functions, so every user callable keeps receiving `$old_value` as it always has (optional-second, variadic, `sanitize_title_with_dashes`, Premium's `sanitize_reserved_post_types`). Only internal functions that cannot take a second argument get the value alone.
- Closures go through `ReflectionFunction`; `ReflectionMethod( $closure, '__invoke' )` reports user closures as internal.
- If a callable cannot be reflected, both arguments are passed, matching develop.
- The two tests that pinned the broader behaviour are replaced by a matrix: internal optional-second (`intval` null and non-null old value, `strlen`, `round`, `trim`, `htmlspecialchars`, `json_decode`), internal callable forms (`Closure::fromCallable`, first-class callable, `ArrayObject::append`), internal required-second (`array_slice`), every user-callable form, and an unreflectable `__call` callable. Six of the user rows fail against `c3b100b5`.

Verified: PHPUnit 183 tests / 1338 assertions green; browser tests on a local site — Premium CPT Builder keeps the old slug when a reserved one is entered, Premium Lightbox slugifies the album name, Livemesh Portfolio (eight `'intval'` fields) stores `900` from `900abc` with no deprecation, and a Widget Block render of the same widget is clean where develop emits the `intval()` deprecation.

Merging now.

PR #2359, opened 2026-08-20
Fix Contact: Prevent Undefined Array Key Type Errors

AlexGStapleton:
` An error of type E_ERROR was caused in line 1197 of the file so-widgets-bundle/widgets/contact/contact.php. Error message: Uncaught TypeError: Cannot access offset of type string on string in so-widgets-bundle/widgets/contact/contact.php:1197`

The contact form widget's get_less_variables() read design sections such as $instance['design']['fields']['font'] directly. When a widget instance held a corrupt or legacy design value saved as a string (rather than an array), this offset access threw "Cannot access offset of type string on string" on PHP 8+, fataling during template variable generation.

Add a `DESIGN_SECTIONS` const listing the design sections that must always be arrays, and a normalize_design_sections() helper that coerces each to an array. Call it at the top of get_less_variables() so all section lookups (fields, labels, container, descriptions, errors, submit, focus, success) are safe against missing or string-typed data. Only class constants and standard methods are used, preserving the plugin's PHP 7.0 minimum.

### DIFF
--- a/tests/ContactLessVariablesTest.php
+++ b/tests/ContactLessVariablesTest.php
@@ -1,0 +1,224 @@
+<?php
+
+use Brain\Monkey\Functions;
+use PHPUnit\Framework\Attributes\DataProvider;
+use SiteOrigin\Tests\SiteOriginTests;
+
+/**
+ * Unit tests for SiteOrigin_Widgets_ContactForm_Widget::get_less_variables().
+ *
+ * contact.php registers itself by calling siteorigin_widget_register() while
+ * the file is being required, and the widget constructor resolves its plugin
+ * directory, so both have to exist as real functions at that moment — the same
+ * pattern CacheCompatTest uses for add_action().
+ */
+if ( ! function_exists( 'siteorigin_widget_register' ) ) {
+	function siteorigin_widget_register() {
+		return true;
+	}
+}
+
+if ( ! function_exists( 'plugin_dir_path' ) ) {
+	function plugin_dir_path( $file ) {
+		return rtrim( dirname( $file ), '/\\' ) . '/';
+	}
+}
+
+if ( ! defined( 'SOW_BUNDLE_BASE_FILE' ) ) {
+	define( 'SOW_BUNDLE_BASE_FILE', __DIR__ . '/../so-widgets-bundle.php' );
+}
+
+/**
+ * contact.php registers a filter hook at the top level (line 2143) while the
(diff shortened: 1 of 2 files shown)


Misplon (changes requested):
Verified: new tests fail on develop (6), branch green (28), merges cleanly with #2358 (36 green combined). `DESIGN_SECTIONS` matches every section the widget reads.

**Changes requested: the same corrupt data still fatals in four places.**

1. `array( 'design' => 'corrupt-string' )` passes the `empty()` guard, then fatals inside `normalize_design_sections()` itself: writing `$design['container']` on a string throws the same `TypeError` (verified, 8.3). The reported error message fits this input as well as a string section. Guard `! is_array( $design )` first.

2. Normalisation is local to `get_less_variables()`. The corrupt instance still reaches the render path and fatals at:
   - `render_form_fields()` — `$instance['design']['labels']['position']`
   - `tpl/default.php:80` — `$instance['design']['submit']['styled']`
   - `tpl/simple.php:6/8/37` — `$instance['design']['labels']['position']`

   Normalise once in `modify_instance()` so every consumer sees repaired data, instead of per-read guards in three files.

3. Tests cover only `get_less_variables()`. Add a string-`design` case and a render-path case.

Misplon:
Thanks, Alex. Your fix stands as the base. I have pushed 13 commits on top that carry the changes requested in review, so this can merge without another round trip:

- Every raw `$instance['design'][section][key]` read in `get_less_variables()`, `render_form_fields()` and both templates now coalesces (42 reads, not one), so any section stored as a string, null or object renders without a fatal or warning.
- `modify_instance()` repairs malformed design sections before they are read, so a re-save from the builder stores arrays again.
- `style_value()` blocks arrays and objects where a value becomes CSS, without changing scalar types. Blank measurement fields are stored as `false`, and the style hash is built from those values, so a cast here would have regenerated CSS on every site. A test now pins `get_style_hash()` for two fully populated instances to the values your head (`19820b78`) produces.
- A partially populated design (a section missing outright) does get a new hash once, because `null` and `''` serialise differently. CSS output is identical; the file is regenerated on demand. Recorded as an accepted change.

Verified: PHPUnit 160 tests / 1289 assertions green; wp-cli save and render probes across 22 malformed shapes with zero throwables; browser test on a local site — corrupt sections seeded, front end renders, builder opens the form, re-save repairs the data, and a normal save of a valid instance keeps its hash.

Merging now.

PR #2366, opened 2026-09-04
Editor: Honour Page Builder's Keep-Shortcodes Opt-Out During the Post Content Render

### LINKED ISSUE #2365: Editor: Widget Ignores Page Builder's Keep-Shortcodes Opt-Out Filter During the Post Content Mirror Render
## Summary

Once #2364 ships, the Editor widget skips `do_shortcode()` whenever `SITEORIGIN_PANELS_POST_CONTENT_RENDER` is set. That guard is unconditional. It does not consult Page Builder's `siteorigin_panels_post_content_keep_shortcodes` filter, so a site that returns `false` from that filter to opt back into baked shortcode output gets it for core Text widgets but not for Editor widgets.

## Steps to reproduce

1. Page Builder with siteorigin/siteorigin-panels#1372 and Widgets Bundle with #2364. Copy Content enabled.
2. `add_filter( 'siteorigin_panels_post_content_keep_shortcodes', '__return_false' );`
3. Save a page with an Editor widget containing a shortcode.
4. Inspect `post_content`.

**Expected:** the shortcode output is baked into the mirror, matching the filter's contract and the core Text widget's behaviour.

**Actual:** the raw shortcode is kept. The same page with a core Text widget bakes correctly.

Verified locally against both fix branches: Text widget honours the filter, Editor widget does not.

## Suggested fix

In `widgets/editor/editor.php`, when `SITEORIGIN_PANELS_POST_CONTENT_RENDER` is set, run `do_shortcode()` anyway if `apply_filters( 'siteorigin_panels_post_content_keep_shortcodes', true )` returns `false`. Keep the legacy `SITEORIGIN_PANELS_CACHE_RENDER` guard as is.

Low impact: the filter is new and opt-in, and the default behaviour is correct in both plugins.

Misplon:
Fixes #2365.

The Editor widget now consults Page Builder's `siteorigin_panels_post_content_keep_shortcodes` filter during the post content render. When a site returns false to opt back into baked output, the widget runs `do_shortcode()` on the raw text, matching what it produced before #2364 and matching the core Text widget under Page Builder's own opt-out. The legacy cache render keeps shortcodes unconditionally, as before.

**Held as a draft.** The filter is introduced by siteorigin/siteorigin-panels#1373, open for review (the original #1372 was reverted on develop). Without it the tag is unfiltered and this branch changes nothing. Merge only if #1373 lands.

Tests: two cases in `EditorWidgetShortcodeTest`, red against develop, full suite green. Verified live through a classic editor save with the opt-out on (baked) and off (raw), and a logged-out front end render.

### DIFF
--- a/tests/EditorWidgetShortcodeTest.php
+++ b/tests/EditorWidgetShortcodeTest.php
@@ -106,4 +106,45 @@ class EditorWidgetShortcodeTest extends SiteOriginTests {
 
 		$this->assertSame( 'Intro text [ninja_forms id=1]', $variables['text'] );
 	}
+
+	// Page Builder's opt-out filter returns false for a site that wants baked
+	// shortcode output in its post content copy; every other tag passes through.
+	private function opt_out_of_keeping_shortcodes(): void {
+		Functions\when( 'apply_filters' )->alias(
+			fn( $tag, $value ) => $tag === 'siteorigin_panels_post_content_keep_shortcodes' ? false : $value
+		);
+	}
+
+	public function test_keep_shortcodes_opt_out_executes_shortcodes_in_post_content_render() {
+		$GLOBALS['SITEORIGIN_PANELS_POST_CONTENT_RENDER'] = true;
+		$this->opt_out_of_keeping_shortcodes();
+
+		Functions\when( 'shortcode_unautop' )->alias(
+			fn( $text ) => 'unautop(' . $text . ')'
+		);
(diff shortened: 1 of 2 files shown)


Misplon:
Closing: siteorigin/siteorigin-panels#1373 introduced the filter this consults and was closed without merging (the reported case is fixed by #2364 alone). Re-open together with #1373 if the global change is ever re-landed.

